PR #3, opened 2022-01-25
Update index.adoc

stopWarByWar:
the interface get_status is worry.
From canisters -> canister

### DIFF
--- a/modules/interface-spec/pages/index.adoc
+++ b/modules/interface-spec/pages/index.adoc
@@ -545,8 +545,8 @@ All requested paths must have one of the following paths as prefix:
  * `/time`. Can be requested by anyone.
  * `/subnet`. Can be requested by anyone.
  * `/request_status/<request_id>`. Can only be read if `/request_id/<request_id>` is not present in the state tree, or if this `read_state` request was signed by same sender as the original request referenced by `<request_id>`, and the effective canister id of the original request matches the `<effective_canister_id>` (see <<http-effective-canister-id>>) in this HTTP request’s path.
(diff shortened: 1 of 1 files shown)


modules/interface-spec/pages/index.adoc
```diff
@@ -545,8 +545,8 @@ All requested paths must have one of the following paths as prefix:
...
- * `/canisters/<canister_id>/module_hash`. Can be requested by anyone, if `<canister_id>` matches `<effective_canister_id>`.
- * `/canisters/<canister_id>/controllers`. Can be requested by anyone, if `<canister_id>` matches `<effective_canister_id>`. The order may vary depending on the implementation.
+ * `/canister/<canister_id>/module_hash`. Can be requested by anyone, if `<canister_id>` matches `<effective_canister_id>`.
+ * `/canister/<canister_id>/controllers`. Can be requested by anyone, if `<canister_id>` matches `<effective_canister_id>`. The order may vary depending on the implementation.
 
 Note that the paths `/canisters/<canister_id>/certified_data` are not accessible with this method; these paths are only exposed to the canister themselves via the System API (see <<system-api-certified-data>>).
```
nomeata, line 551: Looks like there are more

```suggestion
Note that the paths `/canister/<canister_id>/certified_data` are not accessible with this method; these paths are only exposed to the canister themselves via the System API (see <<system-api-certified-data>>).
```

Dfinity-Bjoern:
Thank you for suggesting the changes! I unfortunately have to close the PR for now as we cannot accept external contributions yet. We hope to sort that out as soon as possible, and we will announce on the forum when that is the case. Apologies for the inconvenience, and I hope we can just re-open this PR then.

PR #4, opened 2022-01-25
EXC-783: Remove limit on Cycles in a canister

adambratschikaye:
We no longer enforce a limit on the amount of cycles a canister can
hold, so this limit should be removed from the spec.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -98,9 +98,9 @@ Amount of cycles that a canister has to have before a message is attempted to be
 +
 Amount of cycles that the IC sets aside when a canister performs a call. This is used to pay for processing the response message, and unused cycles after the execution of the response are refunded. See <<rule-message-execution>>.
 
-* `MAX_CANISTER_BALANCE`
+* `DEFAULT_PROVISIONAL_CYCLES_BALANCE`
 +
-Maximum canister cycle balance. Any excess is discarded. Less than 2^128^.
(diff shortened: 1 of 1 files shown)


Dfinity-Bjoern (approved):
Looks good, thanks @adambratschikaye!
Does the change in the replica break anything in the `ic-ref-test` suite, meaning: do we have to adapt that test suite?

adambratschikaye:
Here's an MR for updates to `ic-ref-test`: https://github.com/dfinity/ic-hs/pull/70

nomeata:
What happens if the balance exceeds 2^128? Or is that still an “implicit” technical limit? (Not a practical concern, but imagine we get to formally verify the implementation, then we should have clarity on that).

Maybe better to keep the limit and set it to 2^128-1?

adambratschikaye:
Currently, in the actual implementation, we're using saturating operations. So if you try to send cycles to a canister that already has 2^128-1 then they'll just disappear.

I prefer that over making the limit be 2^128-1 because I don't think it's worth the added code complexity to handle a situation that isn't actually going to happen. And from the user perspective it also shouldn't make a difference since they won't hit 2^128.

nomeata:
I see. Would it be reasonable to specifying this saturation semantics in the reference as well? Or at least mention it in tbe prose?

adambratschikaye:
Yeah, makes sense - I can add it.

adambratschikaye:
I added a note about it in the section on `Canister cycles`.

nomeata (approved):
Thanks!

adambratschikaye:
Thanks @nomeata

PR #5, opened 2022-01-25
Adding the Bitcoin candid file and specification

THLO:
This PR adds the candid file for the Bitcoin integration.
It also provides the semantics of all new APIs.


spec/bitcoin.did
```diff
@@ -0,0 +1,61 @@
...
+  MalformedAddress;
+  // More error types to be added here.
+};
+
+type SendTransactionError = variant {
+  MalformedTransaction;
+  // More error types to be added here.
+};
+
+service bitcoin: (InitPayload) -> {
```
ielashi, line 42:
```suggestion
service bitcoin : {
```

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+== The IC bitcoin canisters
+
+Similar to the <<ic-management-canister,management canister>>, the Bitcoin functionality is also exposed through a facade, which we refer to as the "Bitcoin canisters".
+There are two Bitcoin canisters differing only in that one operates on the real Bitcoin blockchain whereas the other only works with the Bitcoin testnet.
+The Bitcoin canisters can be accessed using their addresses:
+
+- The canister address of the Bitcoin canister that uses the real Bitcoin blockchain is `TBD`.
+- The canister address of the Bitcoin canister that uses the Bitcoin testnet is `TBD`.
+
+It is *not* possible to interact with either Bitcoin canister via external requests (i.e., ingress messages).
```
ielashi, line 1688: Also worth mentioning that it is only possible to interact with this canister on specific subnets (i.e. bitcoin-enabled subnets).

This might potentially be a can of worms, as now we'll need to define what a bitcoin-enabled subnet is, and also we need to introduce the concept of subnets, which this specification has been trying to avoid for the longest time.

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+=== IC method `get_utxos`
+
+Given a https://en.bitcoin.it/wiki/Base58Check_encoding[base58-encoded] address as part of the
+`GetUtxosRequest` , the function returns all unspent transaction outputs (UTXOs) associated with the
+provided address based on the current view of the Bitcoin blockchain available to the Bitcoin canister.
+
+If the call fails, e.g., because the address is malformed, a `GetUtxosError` variant is returned,
+indicating the reason for the failed call.
+
+The optional `min_confirmations` parameter can be used to limit the returned UTXOs to those with at
```
ielashi, line 1712: Is there a maximum `min_confirmations` that can be used? I assume the upper bound is our `delta` parameter correct?

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+- All signatures are correct.
+- There is a positive transaction fee.
+- The transaction does not create dust, i.e., an output that holds a smaller Bitcoin amount
+than it costs to spend the Bitcoin in the output, making the output worthless.
+
+If at least one of these checks fails, a `SendTransactionError` is returned,
+indicating the reason for the failed call.
+
+If the transaction passes these tests, the transaction is forwarded to the
+Bitcoin adapter, which offers it to all connected Bitcoin nodes.
```
ielashi, line 1759: nit: IMHO I think the bitcoin adapter should be completely internal and not part of the interface specification. I'd keep it more high-level (and ditto for the paragraph below).

```suggestion
If the transaction passes these tests, the transaction is forwarded to the
Bitcoin network.
```

spec/index.adoc
THLO: Hhmm... so what should we do here? It's impossible to talk about this without talking about subnets.

spec/index.adoc
THLO: Technically, `min_confirmations` can be as large as the length of the whole chain since we know the confirmation count for all outputs in the UTXO set.
If `min_confirmations` is larger than the length of the whole chain, an empty set is returned.

spec/index.adoc
ielashi: I don't think there's a way around not talking about subnets either. I'd be curious what Bjoern thinks here.

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
 
 This method is only available in local development instances.
 
+
+[#ic-bitcoin-canisters]
+== The IC bitcoin canisters
+
+Similar to the <<ic-management-canister,management canister>>, the Bitcoin functionality is also exposed through a facade, which we refer to as the "Bitcoin canisters".
```
Dfinity-Bjoern, line 1681: Why did we opt for making the API available trough a canister facade instead of extending the System API? Are (some of) the functions inherently asynchronous, or was it "just" making "two copies" of the API easily available for Bitcoin testnet and mainnet, respectively?

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+There are two Bitcoin canisters differing only in that one operates on the real Bitcoin blockchain whereas the other only works with the Bitcoin testnet.
+The Bitcoin canisters can be accessed using their addresses:
+
+- The canister address of the Bitcoin canister that uses the real Bitcoin blockchain is `TBD`.
```
Dfinity-Bjoern, line 1685: I'd like @bogwar and @akhi3030 to chime in here as well, but IMO we could choose a sub-range from the `01` type of principals. Currently we use everything with `nat64 | 0101` as a typical canister id, and I could imagine using `nat64(i) | 0201` for the Bitcoin canisters, with `i in {0,1}`.

spec/index.adoc
Dfinity-Bjoern: I think that is breaking consistency of the interface. If the interface should not be available everywhere, in particular via ingress, then we should maybe do this as an extension to the System API instead of as a canister facade?

An alternative would be to specify a `NotAllowed` error type for the return value. The canister facade is free to return that value, but we should not have this fail as "not addressable" from some sources.

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+----
+include::{example}bitcoin.did[]
+----
+
+The binary encoding of arguments and results are as per Candid specification.
+
+[#ic-bitcoin_get_utxos]
+=== IC method `get_utxos`
+
+Given a https://en.bitcoin.it/wiki/Base58Check_encoding[base58-encoded] address as part of the
```
Dfinity-Bjoern, line 1705: Why do we use an encoded request and a `text` instead of a `blob`?

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+If the call fails, e.g., because the address is malformed, a `GetUtxosError` variant is returned,
+indicating the reason for the failed call.
+
+The optional `min_confirmations` parameter can be used to limit the returned UTXOs to those with at
+least the provided number of confirmations, i.e., transactions with fewer than this number of
+confirmations are not considered.
+If this parameter is not used, the default value is 0.
+
+The optional `offset` parameter can be used to specify a starting offset in the list of UTXOs.
+Since a response will contain at most `TBD` `Utxo` records, this parameter is useful for addresses
```
Dfinity-Bjoern, line 1718: I don't think we can define the API like that:

- If we want pagination, then we should also let the canister specify the `limit`, and say the `limit` can be at most `TBD`, and return an error if `limit > TBD`. Otherwise we can never change `TBD` even if we figure out that we really need to, without breaking existing canisters in an extremely bad way. (It's better if they consistently fail than if they return wrong information.)
- As you point out below: The interaction with the Bitcoin canister is asynchronous, so the state of the Bitcoin component may change between the first and second calls. The requestor does not even have a way to understand whether the responses are inconsistent. Pagination with respect to inconsistent state seems dangerous especially for applications that deal with funds.

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+Given a https://en.bitcoin.it/wiki/Base58Check_encoding[base58-encoded] address as part of the
+`GetUtxosRequest` , the function returns all unspent transaction outputs (UTXOs) associated with the
+provided address based on the current view of the Bitcoin blockchain available to the Bitcoin canister.
+
+If the call fails, e.g., because the address is malformed, a `GetUtxosError` variant is returned,
+indicating the reason for the failed call.
+
+The optional `min_confirmations` parameter can be used to limit the returned UTXOs to those with at
+least the provided number of confirmations, i.e., transactions with fewer than this number of
+confirmations are not considered.
```
Dfinity-Bjoern, line 1714: How does the function treat UTXOs that have more than `min_confirmations` but which have been spent with less than `min_confirmations`? How does this function generally treat UTXOs that have been spent with few confirmations? Or, more conceptually, how do you define the U in UTXO?

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+Given an address and the optional `min_confirmations` parameter, `get_balance` iterates over
+all UTXOs returned by `get_utxos` for the same two parameters and returns the sum of
+UTXO values.
+Since the function is executed on the replica, there is no pagination limit, i.e., `get_balance`
+will iterate over all relevant UTXOs.
+
+[#ic-send_transaction]
+=== IC method `send_transaction`
+
+Given a `blob` of a Bitcoin transaction, several checks are performed:
```
Dfinity-Bjoern, line 1746: The `blob` is inherently different for testnet and mainnet, right? As the allowed addresses in the `blob` should be different.

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+
+- The transaction is well formed.
+- The transaction only consumes unspent outputs.
```
Dfinity-Bjoern, line 1749: With respect to which state do you check that (a) the UTXOs were created and (b) they are unspent?

spec/index.adoc
```diff
@@ -1674,6 +1674,96 @@ Any user can top-up any canister this way.
...
+- All signatures are correct.
+- There is a positive transaction fee.
+- The transaction does not create dust, i.e., an output that holds a smaller Bitcoin amount
```
Dfinity-Bjoern, line 1752: How is this checked? There is no minimal transaction fee, or is there?

spec/bitcoin.did
```diff
@@ -0,0 +1,61 @@
...
+type GetUtxosRequest = record {
+  address : text;
+  min_confirmations: opt nat32;
+  offset: opt nat32;
+};
+
+
+type GetUtxosError = variant {
+  MalformedAddress;
+  // More error types to be added here.
```
Dfinity-Bjoern, line 24: We cannot evolve this in a backward compatible way by _extending_ the variant we return, to the best of my knowledge. @crusso and @chenyan-dfinity should be the experts here. I think we have to provide the exhaustive list from the beginning.

Dfinity-Bjoern (changes requested):
Thanks @THLO and @ielashi! I have a few concrete concerns as stated in the comments.

More broadly: have we checked that one can actually build useful services based on that API? Or do we expect that to happen while the developer preview is out in the community?

spec/bitcoin.did
crusso: Yes, I agree.

It's possible that having wrapped each Err in opt will allow you to add a variant  in those options in a future version of Candid, but I believe that in the current version, all old clients wll received `null` if a new implementation extends the error variant type, because the deserializer will detect the content types as being incompatible and default to `null`, rather than looking at the value in the option and filtering out "new" ones.

So un-optioned variant returns cannot be extended as you desire, and optioned ones probably don't have the behavior you expect until Candid is fixed. 

My (perhaps naive) suggestion would be to enumerate all the known errors and have catch-all variant for future errors that carries a code and a text value.

I'm no Candid expert, so will defer to @chenyan-dfinity  who should definitely chime in here.

spec/index.adoc
crusso: The System API is super low level and wouldn't have nearly as nice an interface - even passing blob or text values, let alone structured values, is tedious and error prone and requires Motoko compiler support or library support in Rust.
.Receiving multiple values is even worse.

So using a canister is a good choice, IMO.

spec/index.adoc
crusso: Could you not use the IC Management Canister to obtain the optional Bitcoin canister, with the behaviour depending on the subnet?

spec/index.adoc
crusso: Can the canister methods inspect the caller to see if the message is ingress or not? That's a question, not a suggestion.

spec/bitcoin.did
chenyan-dfinity: This is true at the moment, but we are updating the spec to allow adding/removing tags in `opt variant` to be backward compatible. The implementation should rollout in a week or two. So as long as the return type is `opt variant`, instead of `variant`, it should be backward compatible very soon.

spec/index.adoc
chenyan-dfinity: I agree address is better typed as `blob` or even better as `principal`. `blob` is not very easy to copy/paste in the CLI. Principal has the same inner representation as blob, but is easier to copy/paste because of the textual representation and checksum. I wish we use principal for our ledger address as well.

spec/index.adoc
chenyan-dfinity: I think ingress message has to go through boundary node, so we can simply disable routing in the boundary nodes. Similar to the management canister, without an effective canister id, the ingress message cannot call the management canister either.

spec/bitcoin.did
chenyan-dfinity: I agree with @crusso, it's best to add a catch-all variant, and once the candid implementation is fixed, we can add more concrete error variants.

THLO:
> 

Regarding usability of the API, there are two parts to the answer:
1. Yes, the goal is to get feedback from the community after the launch of the developer preview. If necessary, we will then update the API.
2. By design, it is a **low-level API** that is meant to be generic and powerful, and user-friendliness is not the primary concern. We expect the community to build abstractions and (user-friendly) libraries on top.

spec/index.adoc
crusso: Alternatively, could the CLI be extended to allow the same, less error-prone, textual representation with checksum for Blobs? It seems weird to abuse Principal for something that doesn't identify an entity on the IC.

chenyan-dfinity: `blob` has other use cases, e.g. the content of an asset/file. I think `Principal` isn't tied to the IC, it's a type for identity.

spec/bitcoin.did
ielashi: The current candid has each of these variants as `opt` (see endpoints below). Is that not sufficient for backward compatibility? I agree though, a catch-all variant would be more easily understood than an optional variant.

spec/index.adoc
ielashi: With respect to the state of the canister you're talking to (i.e. the mainnet state if talking to the btc mainnet canister, and the bitcoin testnet if talking to the btc testnet canister).

spec/index.adoc
ielashi: Correct, but all the checks below work consistently across both addresses.

spec/index.adoc
ielashi: The dust isn't about the transaction fee. It's about the value of the outputs.

Bitcoin core prevents transactions to be sent that have outputs below a certain value.

```
Bitcoin Core considers a transaction output to be dust, when its value is lower than the cost of spending it at the dustRelayFee rate. The default value for dustRelayFee is 3,000 sat/kvB
```

See https://bitcoin.stackexchange.com/questions/10986/what-is-meant-by-bitcoin-dust

spec/index.adoc
ielashi: Good question. I don't have a good answer personally as to why we use `text`. It was mostly chosen because that's the "standard" of how people think about addresses and provided the most ergonomic interface without any notable downsides compared to a `blob`.

I am not sure a `principal` would be better than `text`. It would have a different textual representation than bitcoin addesses, which would be very confusing for users.

spec/index.adoc
ielashi: > I think that is breaking consistency of the interface. If the interface should not be available everywhere, in particular via ingress, then we should maybe do this as an extension to the System API instead of as a canister facade?

We already do this for a number of methods on the management canisters, like `raw_rand`., but I'd add that not supporting ingress was more out of precaution - there's no fundamental reason why we can't/shouldn't expose it to preserve the symmetry as far as I know.

@THLO aren't we going to be charging cycles for these calls?

THLO: Yes, the plan is to charge cycles for these calls (which obviously won't work for ingress messages).

Dfinity-Bjoern: But if we anyway plan to charge cycles, then don't we anyway need an error code `InsufficientFunds` (or similar) and we can just reject ingress messages consistently?

spec/index.adoc
Dfinity-Bjoern: I don't understand your proposal. What do you want to "obtain"? What exactly would the IC Management Canister return?

spec/index.adoc
Dfinity-Bjoern: I don't think we can/should use the type principal here, since principals are senders/receivers of messages on the IC layer, and Bitcoin addresses are certainly not that. Also the support of better I/O on the CLI seems a pretty good idea (it's indeed pretty painful to use `blob` from dfx) but orthogonal to the spec.

@ielashi I think it comes down to what a Bitcoin address _is_. A principal, for instance, is an octet string of up to 29 bytes, and it has a canonical textual representation. So the question is what _is_ a Bitcoin address, and it seems you're right, is per [this](https://en.bitcoin.it/wiki/Invoice_address):
> A Bitcoin invoice address, or simply invoice, is an identifier of 26-35 alphanumeric characters, beginning with the number 1, 3 or bc1 that represents a possible destination for a bitcoin payment.
That means a Bitcoin address _is_ a `text`. I will close this thread.

spec/index.adoc
Dfinity-Bjoern: Wonderful, that seems consistent. @THLO can we maybe make that a bit more unambiguous in the description, maybe by giving an example or writing in parenthesis explicitly that this holds for both receiving and spending?

spec/index.adoc
Dfinity-Bjoern: I assume that means we use the longest branch known to the Bitcoin canister if there are multiple ones, and the behavior is undefined in case there are two equally long branches (where one may create/spend the UTXO and the other one does not)?

spec/index.adoc
Dfinity-Bjoern: Right but doesn't
> ... than it costs to spend the Bitcoin in the output ...

relate to the transaction fee? What does it cost to spend the output if not the transaction fee? Can we put the definition of _dust_ we intend to use down in the specification here?

spec/index.adoc
THLO: I wrote "**transactions** with fewer than..." to stress that neither inputs nor outputs are considered in this computation but I agree with you, it could be clearer. I'll add a bit more detail.

Dfinity-Bjoern:
After discussion with @bogwar @michelabdalla-dfinity @derlerd-dfinity :

There are three possible ways in which the calls to the Bitcoin canister could fail in different ways:
- `DESTINATION_INVALID` – the target canister does not exist or the target canister does not have such a method
- `CANISTER_REJECT` – rejected by the canister
- application-specific error code returned by the canister.

I think you intended one of the first two, although I am not even entirely sure which one. We may want to have a broader discussion about which is the right behavior, since this pattern may appear more widely in the future.

spec/index.adoc
crusso: > I don't understand your proposal. What do you want to "obtain"? What exactly would the IC Management Canister return?

Sorry, didn't see your question. 
I'm suggesting the ManagementCanister exposes an endpoint for getting the bitcoin service, perhaps as an option if there isn't one. Something like (in pseudo-candid).

```
service IC {
    get_bitCoin : () -> BitCoin opt 
}

service BitCoin {
     ....
}
```
So then you don't hard-code the canister ID but need to obtain it from the system.

spec/index.adoc
ielashi: I agree. There is one caveat here that's worth mentioning: if we introduce an application-level error to indicate that cycles are insufficient, then we have two choices when it comes to ingress messages:

- Accept ingress messages in `canister_inspect_message`, and then reject them with the proper application-level error. This preserves the consistency between ingress and canister calls at the expense of not being able to filter out the ingress message early.
- Reject ingress messages in `canister_inspect_message`, which would be better performance-wise, but then the behavior of ingress messages (receiving a reject) and canisters (receiving an application-level error) would be inconsistent.

I wish candid interfaces had a way of specifying rejects, along with their structure. That way, we could reject with a candid error and clients would know from the interface to expect that, but that's beyond the scope of this PR.

spec/bitcoin.did
chenyan-dfinity: Correct.

spec/index.adoc
ielashi: I agree. The following should address both points. WDYT?

1. The limit can now be specified by the caller.
2. The response returns the tip that was used when computing the response, and that tip can be later used in subsequent requests to ensure the same state is being queried. The UTXOs for any given tip will be available for at least `STABILITY_THRESHOLD` confirmations (which will like be ~144 or greater initially).


```
  type GetUtxosRequest = record {
    address : text;
    min_confirmations: opt nat32;
    offset: opt nat32;
    limit: opt nat32;
    tip: opt BlockHash;
  };

  get_utxos: (GetUtxosRequest) -> (variant {
    Ok : record {
      utxos: vec Utxo;
      total_count: nat32;
      tip: BlockHash;
    };
    Err : opt GetUtxosError;
  });
```

spec/index.adoc
THLO: Personally, I would go for consistency - at the risk of paying a (hopefully small) performance penalty.

spec/index.adoc
THLO: The point is that an output is considered dust regardless of the fee that is paid for the transaction. In other words, even if there is a large fee for a transaction, an output can still be considered "dust" (namely when the Bitcoin amount in the output is less than it costs to spend it at the `dustRelayFee` rate).

However, as @ielashi once pointed out, it may be fine for us to ignore the dust problem entirely and let the users deal with the consequences of sending Bitcoin transactions that will likely not be added to blocks (we cannot make any promise that any transaction is added to a block anyway).

spec/index.adoc
nomeata: I don't think there is a choice:

Rejecting intrercanister messages without cycles in the message handler is only reasonable because the sending canister paid an upfront cost for the message, so this is not an efficient cycle drain attack.

(I expect that eventually we want to refine this and make the first _n_ instructions of an inter-canister message handler free for the canister and implicitly covered by the cost of sending the message, enough for a typical access control check, so that an attackers cannot even run an inefficient cycle drain attack, but that is just an aside.)

But for ingress messages, no cycles were paid, and the receiving canister would have to pay for it, and that's _precisely_ the reason we have introduced the `inspect_message` hook - else the cost of draining cycles is just the cost of HTTP (and maybe the rate limits).

It may be slightly inconsistent (nobody says that `inspect_message` is great), but it _is_ the best practice for canisters, and I don't think we should do it differently for “system-provided” canisters, as that would be another inconsistency.

spec/index.adoc
```diff
@@ -1674,6 +1674,101 @@ Any user can top-up any canister this way.
...
+
+The optional `filter` parameter can be used to restrict the set of returned UTXOs, either providing a
+minimum number of confirmations or a pagination record.
+In the first case, only UTXOs with at least the provided number of confirmations
+are returned, i.e., transactions with fewer than this number of
+confirmations are not considered. In other words, if the number of confirmations is `c`, an output
+is returned if it occurred in a transaction with at least `c` confirmations
+and there is no transaction that spends the same output with at least `c` confirmations.
+In the second case, a pagination record must be provided, specifying a block hash that constitutes
+the tip of the considered chain for this request and, optionally, an offset and a limit.
```
ielashi, line 1720: What about the case where this is no `filter` present? Would it make sense to explicitly address that?

spec/bitcoin.did
```diff
@@ -0,0 +1,69 @@
...
+    Ok : record {
+      utxos: vec Utxo;
+      total_count: nat32;
+      tip_hash: BlockHash;
+      tip_height: nat32;
+    };
+    Err : opt GetUtxosError;
+  });
+
+  send_transaction: (raw_transaction: blob) -> (variant {
```
ielashi, line 65: Ah, we've update this in the implementation to the following, as it makes it easier to add fields later:

```
  send_transaction: (SendTransactionRequest) -> (variant {
    Ok;
    Err : opt SendTransactionError;
  });
```

where `SendTransactionRequest` is:

```
type SendTransactionRequest = record {
  transaction: blob;
};
```

spec/bitcoin.did
```diff
@@ -0,0 +1,69 @@
...
+};
+
+type GetBalanceError = variant {
+  MalformedAddress;
+  // More error types to be added here.
+};
+
+type SendTransactionError = variant {
+  MalformedTransaction;
+  // More error types to be added here.
```
ielashi, line 45: Should we remove these comments from the candid? We've already established that these can be extended in a backward-compatible way.

spec/index.adoc
THLO: See line 1727ff. Is this information sufficient?
Would you prefer to have that paragraph first and then talk about the workflow?

ielashi: Ah, I didn't notice it further down below. I think it'd be a bit more clear to first talk about all the possible cases, and then talk about the recommended workflow, but that's not a strong preference.

THLO: I agree. I changed the order.

spec/index.adoc
THLO: Note: We decided to ignore the dust problem and remove this check.

ielashi:
@Dfinity-Bjoern @manudrijvers: We resolved all the outstanding issues in the PR. Do you mind taking another look?

spec/index.adoc
```diff
@@ -1674,6 +1674,100 @@ Any user can top-up any canister this way.
...
+
+[#ic-bitcoin-canisters]
+== The IC bitcoin canisters
+
+Similar to the <<ic-management-canister,management canister>>, the Bitcoin functionality is also exposed through a facade, which we refer to as the "Bitcoin canisters".
+There are two Bitcoin canisters differing only in that one operates on the real Bitcoin blockchain whereas the other only works with the Bitcoin testnet.
+The Bitcoin canisters can be accessed using their addresses:
+
+- The canister address of the Bitcoin canister that uses the real Bitcoin blockchain is `u25tu-yqaaa-aaaaa-aaaaa-eai`.
+- The canister address of the Bitcoin canister that uses the Bitcoin testnet is `jf4vc-xabaa-aaaaa-aaaaa-eai`.
```
ielashi, line 1686: I am starting to reconsider these IDs. These IDs were generated based on Björn's suggestion:

> Currently we use everything with nat64 | 0101 as a typical canister id, and I could imagine using nat64(i) | 0201 for the Bitcoin canisters, with i in {0,1}.

And then I built on that suggestion to generate them as follows:
```
> Principal.fromUint8Array(Uint8Array.from([0,0,0,0,0,0,0,0,2,1])).toText()
'u25tu-yqaaa-aaaaa-aaaaa-eai'
> Principal.fromUint8Array(Uint8Array.from([1,0,0,0,0,0,0,0,2,1])).toText()
'jf4vc-xabaa-aaaaa-aaaaa-eai'
```

The problem I'm starting to see is that these canister IDs don't play nice with our implementation of `RoutingTable`, which assumes that all canisters are nat64 | 0101 (which these are not). This means that we can never use these canister IDs as part of the `RoutingTable` if we wanted to at some point.

I don't have a concrete suggestion yet, but given that some components in our implementation already assume that canisters are always nat64 | 0101, I do think it'd be beneficial for these canisters to also be nat64 | 0101 to avoid surprises.

nomeata: Where does the routing table (at least the publicly visible part of it, e.g. canister_range) assume a certain length? I thought we carefully avoided such implementation details leaking into the interfaces?

Or are you indeed talking about the internal stuff? Then surely the routing table can be changed?

Dfinity-Bjoern: Well, yes and no. The Bitcoin canister (facade) _is not a canister_ in the same sense that the IC management canister (facade) is not a canister. If it were a single canister, then it would need to be deployed on a single subnet. But that's not how we intend to implement it, correct?

nomeata: Why is it not part of `aaaaa-aa` again? It doesn't seem to have much different characteristics than other methods there - like `raw_rand` it's not just canister management, and like most methods it requires routing to another subnet.

nomeata: Ah, because you want to be able to switch networks based on canister IDs? Wouldn't that work better if it's an enum argument? Also easier if more Bitcoin clones need to be supported in the future.

Dfinity-Bjoern: In discussion with @akhi3030: Why don't we just add the methods to IC00?

ielashi: Yes, let's just add them to the management canister. The consequences of different canister IDs on the implementation are rather complicated. I'll update this PR.

ielashi: I had a chat with @THLO about this. We introduced separate canister IDs initially because we thought it's more elegant. Callers won't have to send an enum with which network they want to talk to for every request, and we wouldn't have to prefix all these methods with something like `bitcoin_`.

I have an idea of a simple implementation that will allow us to keep those separate canister IDs that are currently specified in the spec. I'll prototype it to see if it works and will get back to you all.

Dfinity-Bjoern:
As we agreed last week, we discussed the interface issue again more broadly (including @jwiegley @granstrom @bogwar). The outcome is that a general preference for moving the functions to `aaaaa-aa` instead of new "virtual" canisters. The main reason is that virtual canisters are inherently, well, not actual canisters: they do not belong to a specific subnet, so they are not represented in a state tree, they don't have controller and wasm module, and so forth. We already have (some of) these inconsistencies with `aaaaa-aa`, of course, but at least they are in only one place.

If we want to keep the door open for a more canister-centric implementation of the Bitcoin functionality (which, I think, has quite some support), then we may choose to mark the API on `aaaaa-aa` as "provisional" or "experimental", so developers are aware they may need to access the same functionality through different means later on.

spec/ic.did
```diff
@@ -35,6 +35,68 @@ type http_request_error = variant {
...
+
+type utxo = record {
+  outpoint: outpoint;
+  value: satoshi;
+  height: nat32;
+};
+
+type pagination = record {
+  tip_block_hash : block_hash;
+  offset: opt nat32;
```
ielashi, line 62: Note that, for performance reasons, the `offset` will likely be a blob rather than a `nat32`. I'll update the PR here once I have sufficient clarity.

spec/ic.did
```diff
@@ -35,6 +35,68 @@ type http_request_error = variant {
...
+  address : bitcoin_address;
+  network: bitcoin_network;
+  filter: opt variant {
+    min_confirmations: nat32;
+    pagination: pagination;
+  };
+};
+
+type get_utxos_error = variant {
+  MalformedAddress;
```
ielashi, line 76: @THLO I think we need to add this new error type and document it in the spec: https://github.com/dfinity/ic/blob/master/rs/bitcoin/types/public/src/lib.rs#L53

spec/index.adoc
```diff
@@ -1743,6 +1743,12 @@ are returned, i.e., transactions with fewer than this number of
 confirmations are not considered. In other words, if the number of confirmations is `c`, an output
 is returned if it occurred in a transaction with at least `c` confirmations
 and there is no transaction that spends the same output with at least `c` confirmations.
+There is an upper bound of 144 on the minimum number of confirmations because only the 144 most recent
+blocks are cached and it is not possible to return the correct result based on the UTXO set.
```
ielashi, line 1747:
```suggestion
There is an upper bound of 144 on the minimum number of confirmations.
```

1) I don't think we need/should provide technical justifications or implementation details as part of the spec, so I'd lean towards removing that part.

2) The upper bound we choose doesn't have to be related to the stability threshold we have internally. Can we not start with a more conservative number (e.g. 10 or 20)? That can help us do some optimizations, and we can always expand that limit later.

3) If I understand correctly, I don't think we can always serve requests with 144 confirmations. In cases where we have forks, for example, we may not have enough information and return a `MinConfirmationsTooLarge` error. If that's correct, then maybe worth mentioning that this number is also not a strict lower bound?

spec/ic.did
Dfinity-Bjoern: Can you elaborate? Where does the impact come from and how big is it?

Dfinity-Bjoern (approved):
Looks good to me now. Thank you @ielashi and @THLO for incorporating all the changes!

spec/ic.did
jwiegley: @ielashi Can you say more?

nomeata: BTW, the official Candid guidance by @rossberg is to not use `nat32` in Candid unless you really want a bit vector (e.g. for crypto). `nat` is usually _smaller_ on the wire, too.

Maybe the performance of `blob` benefit @ielashi alludes to is that we can avoid that discussion :-D

spec/ic.did
```diff
@@ -72,6 +136,29 @@ service ic : {
...
+    Ok : satoshi;
+    Err: opt get_balance_error;
+  });
+
+  bitcoin_get_utxos: (get_utxos_request) -> (variant {
+    Ok : record {
+      utxos: vec utxo;
+      tip_block_hash: block_hash;
+      tip_height: nat32;
+	    next_page: opt blob;
```
ielashi, line 150: nit: indentation

ali-dfinity:
droid update cla

ielashi:
@Dfinity-Bjoern @jwiegley @manudrijvers  Did you get a chance to review the change we did to the pagination API? See [my earlier comment](https://github.com/dfinity/interface-spec/pull/5#discussion_r845082858) on the reason behind this change.

spec/ic.did
ielashi: I don't think so. Even if such a list existed, we cannot efficiently find and remove elements. It's also more complicated due to the following point that I briefly brushed over:

> There's [...] an additional challenge related to storing a list of unbounded size in stable memory, but it's not too relevant for this discussion so I'll leave out the details.

Even if we had a `SerializedList` implementation, it would make the solution more complex. This map is stored in stable memory where we manage the memory ourselves. With our approach of a `BTreeSet` every entry is bounded in size, and we need is a simple constant-size allocator for managing memory. On the other hand, a `SerializedList` would be unbounded in size, and would require implementing a dynamic memory allocator, and that introduces a lot of additional complexity.

Dfinity-Bjoern (approved):
The pagination change looks fine to me. Some thoughts:

* We do not allow the user to specify the `limit`. Which means that the user must be comfortable with any number of UTXOs the system returns. That is probably fine from a practical perspective, since we're not going to use outrageously large numbers there anyway, but given that we may be dealing with smart contracts that can never be upgraded, I am thinking whether we should be providing a *guaranteed* upper bound on number of UTXOs that will be on one page. What do you think?
* We do not return the total number of UTXOs for a given address anymore. That may be fine, since the canister can just iterate over all of them, and having the total number in the beginning may not be important.
* We still return the hash and the height. Is the thinking here that a canister may want to use that to detect when the previous information got outdated due to switching to a different fork?

THLO:
> * We do not allow the user to specify the `limit`. Which means that the user must be comfortable with any number of UTXOs the system returns. That is probably fine from a practical perspective, since we're not going to use outrageously large numbers there anyway, but given that we may be dealing with smart contracts that can never be upgraded, I am thinking whether we should be providing a _guaranteed_ upper bound on number of UTXOs that will be on one page. What do you think?

We can definitely introduce such an upper bound.
It will require some experimenting to come up with a reasonable value, though. I can update the spec once we have a fixed value. @ielashi, do you agree?

> * We do not return the total number of UTXOs for a given address anymore. That may be fine, since the canister can just iterate over all of them, and having the total number in the beginning may not be important.

Correct, we can't return the total number of UTXOs efficiently, so we don't do it. I also think that this is not a big limitation.

> * We still return the hash and the height. Is the thinking here that a canister may want to use that to detect when the previous information got outdated due to switching to a different fork?

The thinking is more that the information could be useful later when the HTTP call functionality is in place, which could then be used to download more information or entire blocks of transactions based on the provided hash or height.
In theory, the information could reveal the presence of forks: If successive `get_utxos` call return different hashes for the same block height, this would indicate that there is a fork. However, forks may go unnoticed if the block heights of the returned tips are not identical.
As customary in Bitcoin, a large number of confirmations should be used to get confidence in the returned data.

THLO:
> * We do not allow the user to specify the `limit`. Which means that the user must be comfortable with any number of UTXOs the system returns. That is probably fine from a practical perspective, since we're not going to use outrageously large numbers there anyway, but given that we may be dealing with smart contracts that can never be upgraded, I am thinking whether we should be providing a _guaranteed_ upper bound on number of UTXOs that will be on one page. What do you think?

We can definitely introduce such an upper bound.
It will require some experimenting to come up with a reasonable value, though. I can update the spec once we have a fixed value. @ielashi, do you agree?

> * We do not return the total number of UTXOs for a given address anymore. That may be fine, since the canister can just iterate over all of them, and having the total number in the beginning may not be important.

Correct, we can't return the total number of UTXOs efficiently, so we don't do it. I also think that this is not a big limitation.

> * We still return the hash and the height. Is the thinking here that a canister may want to use that to detect when the previous information got outdated due to switching to a different fork?

The thinking is more that the information could be useful later when the HTTP call functionality is in place, which could then be used to download more information or entire blocks of transactions based on the provided hash or height.
In theory, the information could reveal the presence of forks: If successive `get_utxos` call return different hashes for the same block height, this would indicate that there is a fork. However, forks may go unnoticed if the block heights of the returned tips are not identical.
As customary in Bitcoin, a large number of confirmations should be used to get confidence in the returned data.

ielashi:
> We can definitely introduce such an upper bound.
> It will require some experimenting to come up with a reasonable value, though. I can update the spec once we have a fixed value. @ielashi, do you agree?

I updated the spec to include a bound of 100,000 UTXOs per request.

The rationale here is that a UTXO is 48 bytes, and we currently have a cap of 2MiB per request, so we can return ~43k UTXOs. I double that number and rounded it up to 100k to give us more breathing room should we ever increase the response size in the future want want to return more UTXOs.

spec/index.adoc
```diff
@@ -1738,6 +1738,113 @@ Any user can top-up any canister this way.
...
+- The transaction only consumes unspent outputs with respect to the current (longest)
+blockchain, i.e., there is no block on the (longest) chain that consumes any of these outputs.
+- There is a positive transaction fee.
+
+If at least one of these checks fails, the call is rejected.
+
+If the transaction passes these tests, the transaction is forwarded to the
+specified Bitcoin network.
+
+The Bitcoin component of the replica periodically forwards the transaction
```
stefan-at-dfinity, line 1831: The spec so far does not mention any implementation specific components (including the `replica`). Sometimes the spec makes reference to the "most likely behavior", like in the case of heartbeats. 

I would strongly recommend that the behavior is specified based on the abstract state of the Internet Computer only. Does that make sense?

ielashi: I agree that we should not be mentioning the replica here, as its only one implementation of the spec. WDYT @THLO?

THLO: I agree.
Does one of you have a suggestion on how to rephrase it?

spec/index.adoc
```diff
@@ -1687,6 +1687,113 @@ Any user can top-up any canister this way.
...
+- There is a positive transaction fee.
+
+If at least one of these checks fails, the call is rejected.
+
+If the transaction passes these tests, the transaction is forwarded to the
+specified Bitcoin network.
+
+The Bitcoin component periodically forwards the transaction
+until the transaction appears in a block appended to the blockchain or the transaction
+times out after 24 hours. As soon as it appears in a block or expires,
```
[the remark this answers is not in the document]
Dfinity-Bjoern: Right, but let's deal with that later. I actually think most of the Bitcoin functionality would be described at the lower level.

PR #6, opened 2022-01-25
System API for threshold ECDSA

ninegua:
We propose two new system APIs for the upcoming threshold ECDSA feature in the management canister.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -48,7 +48,7 @@ service ic : {
   get_ecdsa_public_key : (record {
     canister_id : opt canister_id;
     derivation_path : vec blob;
-  }) -> (blob);
+  }) -> (record { public_key: blob; chain_code: blob; });
   sign_with_ecdsa : (record {
     message_hash : blob;
     derivation_path : vec blob;
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -44,6 +44,16 @@ service ic : {
   deposit_cycles : (record {canister_id : canister_id}) -> ();
   raw_rand : () -> (blob);
 
+  // Threshold ECDSA signature
+  get_ecdsa_public_key : (record {
+    canister_id : opt canister_id;
+    derivation_path : vec blob;
+  }) -> (blob);
+  sign_with_ecdsa : (record {
+    message_hash : blob;
+    derivation_path : vec blob;
+  }) -> (blob);
+
   // provisional interfaces for the pre-ledger world
   provisional_create_canister_with_cycles : (record {
     amount: opt nat;
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -48,10 +48,12 @@ service ic : {
   get_ecdsa_public_key : (record {
     canister_id : opt canister_id;
     derivation_path : vec blob;
+    key_id : text;
   }) -> (record { public_key: blob; chain_code: blob; });
   sign_with_ecdsa : (record {
     message_hash : blob;
     derivation_path : vec blob;
+    key_id : text;
   }) -> (blob);
 
   // provisional interfaces for the pre-ledger world
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -73,7 +73,7 @@ service ic : {
   }) -> (variant { Ok : http_response; Err: opt http_request_error });
 
   // Threshold ECDSA signature
-  get_ecdsa_public_key : (record {
+  ecdsa_public_key : (record {
     canister_id : opt canister_id;
     derivation_path : vec blob;
     key_id : text;
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -49,7 +49,7 @@ service ic : {
     canister_id : opt canister_id;
     derivation_path : vec blob;
     key_id : text;
-  }) -> (record { public_key: blob; chain_code: blob; });
+  }) -> (record { public_key : blob; chain_code : blob; });
   sign_with_ecdsa : (record {
     message_hash : blob;
     derivation_path : vec blob;
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -35,6 +35,8 @@ type http_request_error = variant {
   conn_timeout;
 };
 
+type ecdsa_curve = variant { secp256k1; };
+
 service ic : {
   create_canister : (record {
     settings : opt canister_settings
@@ -76,12 +78,12 @@ service ic : {
   ecdsa_public_key : (record {
     canister_id : opt canister_id;
     derivation_path : vec blob;
-    key_id : text;
+    key_id : record { curve: ecdsa_curve; name: text };
   }) -> (record { public_key : blob; chain_code : blob; });
   sign_with_ecdsa : (record {
     message_hash : blob;
     derivation_path : vec blob;
-    key_id : text;
+    key_id : record { curve: ecdsa_curve; name: text };
   }) -> (record { signature : blob });
 
   // provisional interfaces for the pre-ledger world
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -54,7 +54,7 @@ service ic : {
     message_hash : blob;
     derivation_path : vec blob;
     key_id : text;
-  }) -> (blob);
+  }) -> (record { signature : blob });
 
   // provisional interfaces for the pre-ledger world
   provisional_create_canister_with_cycles : (record {
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1657,7 +1657,7 @@ This method takes no input and returns 32 pseudo-random bytes to the caller. The
 [#ic-get_ecdsa_public_key]
 === IC method `get_ecdsa_public_key`
 
-This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
+This method returns the ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
 The `derivation_path` is a vector of variable length byte strings.
 For https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032] compatibility, each byte string (`blob`) must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^ (non-hardened).
 The only supported `key_id` value at the moment is `"secp256k1"`.
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1654,6 +1654,22 @@ There is no restriction on who can invoke this method.
 
 This method takes no input and returns 32 pseudo-random bytes to the caller. The return value is unknown to any part of the IC at time of the submission of this call. A new return value is generated for each call to this method.
 
+[#ic-get_ecdsa_public_key]
+=== IC method `get_ecdsa_public_key`
+
+This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller. The derivation path is a vector of bytes.
+
+This call requires that the ecdsa feature is enabled and working properly, and the `canister_id` meets the requirement of a canister id.
+Otherwise it will be rejected.
+
+[#ic-sign_with_ecdsa]
+=== IC method `sign_with_ecdsa`
+
+This method returns a new threshold ECDSA signature of the given `message_hash` that can be separately verified against the caller's ECDSA public key derived from the given derivation path.
+
+This call requires that the ecdsa feature is enabled and working properly, the caller is a canister, and `message_hash` is of 32-bytes long.
+Otherwise it will be rejected.
+
 [#ic-provisional_create_canister_with_cycles]
 === IC method `provisional_create_canister_with_cycles`
 
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1657,8 +1657,10 @@ This method takes no input and returns 32 pseudo-random bytes to the caller. The
 [#ic-get_ecdsa_public_key]
 === IC method `get_ecdsa_public_key`
 
-This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller. The derivation path is a vector of variable length byte strings.
+This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
+The `derivation_path` is a vector of variable length byte strings.
 For https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032] compatibility, each byte string (`blob`) must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^ (non-hardened).
+The only supported `key_id` value at the moment is `"seckp256k1"`.
 
 This call requires that the ECDSA feature is enabled and working properly, and the `canister_id` meets the requirement of a canister id.
 Otherwise it will be rejected.
@@ -1666,7 +1668,8 @@ Otherwise it will be rejected.
 [#ic-sign_with_ecdsa]
 === IC method `sign_with_ecdsa`
 
-This method returns a new threshold ECDSA signature of the given `message_hash` that can be separately verified against the caller's ECDSA public key derived from the caller's canister id and the given derivation path.
+This method returns a new threshold ECDSA signature of the given `message_hash` that can be separately verified against a derived ECDSA public key.
+This public key can be obtained by calling `get_ecdsa_public_key` with the caller's `canister_id`, and the same `derivation_path` and `key_id` used here.
 
 This call requires that the ECDSA feature is enabled and working properly, the caller is a canister, and `message_hash` is of 32-bytes long.
 Otherwise it will be rejected.
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1662,9 +1662,10 @@ This method takes no input and returns 32 pseudo-random bytes to the caller. The
 
 This method returns a https://www.secg.org/sec1-v2.pdf[SEC1] encoded ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
 The `derivation_path` is a vector of variable length byte strings.
-The only supported `key_id` value at the moment is `"secp256k1"`.
+The `key_id` is a struct specifying both a curve and a name.
+The availability of a particular `key_id` depends on implementation.
 
-For `key_id="secp256k1"`, the public key is derived using a generalization of BIP32 (see https://ia.cr/2021/1330[ia.cr/2021/1330, Appendix D]). To derive (non-hardened) https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032]-compatible public keys, each byte string (`blob`) in the `derivation_path` must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^.
+For curve `secp256k1`, the public key is derived using a generalization of BIP32 (see https://ia.cr/2021/1330[ia.cr/2021/1330, Appendix D]). To derive (non-hardened) https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032]-compatible public keys, each byte string (`blob`) in the `derivation_path` must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^.
 
 The return result is an extended public key consisting of an ECDSA `public_key`, encoded in https://www.secg.org/sec2-v2.pdf[SEC1] compressed form, and a `chain_code`, which can be used to deterministically derive child keys of the `public_key`.
 
@@ -1677,7 +1678,7 @@ Otherwise it will be rejected.
 This method returns a new https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf[ECDSA] signature of the given `message_hash` that can be separately verified against a derived ECDSA public key.
 This public key can be obtained by calling `ecdsa_public_key` with the caller's `canister_id`, and the same `derivation_path` and `key_id` used here.
 
-The signatures are encoded as the concatenation of the https://www.secg.org/sec2-v2.pdf[SEC1] encodings of the two values r and s. For key_id="secp256k1", this corresponds to 32-byte big-endian encoding.
+The signatures are encoded as the concatenation of the https://www.secg.org/sec2-v2.pdf[SEC1] encodings of the two values r and s. For curve `secp256k1`, this corresponds to 32-byte big-endian encoding.
 
 This call requires that the ECDSA feature is enabled, the caller is a canister, and `message_hash` is 32 bytes long.
 Otherwise it will be rejected.
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1657,8 +1657,8 @@ There is no restriction on who can invoke this method.
 
 This method takes no input and returns 32 pseudo-random bytes to the caller. The return value is unknown to any part of the IC at time of the submission of this call. A new return value is generated for each call to this method.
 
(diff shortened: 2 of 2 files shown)


ninegua:
Points for discussion:

1. Support additional parameters like key type / curve, key identifier (which subnet key), etc.

These can be added later by adding optional fields to the input. It will be backward compatible as long as the new fields will have sensible defaults.

2. Support more elaborate error message.

At the moment all errors are delivered through reject messages instead of returning a variant like `Result` type.

3. Whether / how the `derivation_path` can be made compliant to BIP32.

spec/ic.did
```diff
@@ -44,6 +44,16 @@ service ic : {
...
 
+  // Threshold ECDSA signature
+  get_ecdsa_public_key : (record {
+    canister_id : opt canister_id;
+    derivation_path : vec blob;
+  }) -> (blob);
+  sign_with_ecdsa : (record {
+    message_hash : blob;
+    derivation_path : vec blob;
+  }) -> (blob);
```
manudrijvers, line 55:
```suggestion
  sign_with_ecdsa : (record {
    message_hash : blob;
    derivation_path : vec blob;
    key_id : text;
  }) -> (blob);
```

manudrijvers (commented):
comments from the ecdsa sync meeting

manudrijvers (approved):

spec/ic.did
```diff
@@ -44,6 +44,18 @@ service ic : {
   deposit_cycles : (record {canister_id : canister_id}) -> ();
   raw_rand : () -> (blob);
 
+  // Threshold ECDSA signature
+  get_ecdsa_public_key : (record {
+    canister_id : opt canister_id;
+    derivation_path : vec blob;
+    key_id : text;
+  }) -> (record { public_key : blob; chain_code : blob; });
```
rsubramaniyam, line 52: Can these APIs return error?

spec/index.adoc
```diff
@@ -1654,6 +1654,28 @@ There is no restriction on who can invoke this method.
...
+=== IC method `get_ecdsa_public_key`
+
+This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
+The `derivation_path` is a vector of variable length byte strings.
+For https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032] compatibility, each byte string (`blob`) must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^ (non-hardened).
+The only supported `key_id` value at the moment is `"secp256k1"`.
+
+The return result also contains a `chain_code` which can be used for key derivation purpose if needed.
+
+This call requires that the ECDSA feature is enabled and working properly, and the `canister_id` meets the requirement of a canister id.
```
rsubramaniyam, line 1667: "ECDSA feature is enabled and working properly" -> "ECDSA feature is enabled" (everywhere)

spec/index.adoc
```diff
@@ -1654,6 +1654,28 @@ There is no restriction on who can invoke this method.
...
 
+[#ic-get_ecdsa_public_key]
+=== IC method `get_ecdsa_public_key`
+
+This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
+The `derivation_path` is a vector of variable length byte strings.
+For https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032] compatibility, each byte string (`blob`) must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^ (non-hardened).
+The only supported `key_id` value at the moment is `"secp256k1"`.
+
+The return result also contains a `chain_code` which can be used for key derivation purpose if needed.
```
fspreiss, line 1665: nit: should this be "purposes" (rather than "purpose")?

spec/index.adoc
```diff
@@ -1654,6 +1654,28 @@ There is no restriction on who can invoke this method.
...
+This call requires that the ECDSA feature is enabled and working properly, and the `canister_id` meets the requirement of a canister id.
+Otherwise it will be rejected.
+
+[#ic-sign_with_ecdsa]
+=== IC method `sign_with_ecdsa`
+
+This method returns a new threshold ECDSA signature of the given `message_hash` that can be separately verified against a derived ECDSA public key.
+This public key can be obtained by calling `get_ecdsa_public_key` with the caller's `canister_id`, and the same `derivation_path` and `key_id` used here.
+
+This call requires that the ECDSA feature is enabled and working properly, the caller is a canister, and `message_hash` is of 32-bytes long.
```
fspreiss, line 1676: nit: I guess this should either be "is 32 bytes long" (i.e., without "of" and "-") or "is a 32-byte `blob`"?

spec/index.adoc
andreacerulli, line 1665: We should specify the encoding of the public key. I suggested it here, but we can also put it at the beginning.

```suggestion
The return result is an extended public key consisting of an ECDSA `public_key`, encoded in https://www.secg.org/sec2-v2.pdf[SEC1] compressed form, and a `chain_code`, which can be used to deterministically derive child keys of the `public_key`.
```

spec/index.adoc
```diff
@@ -1654,6 +1654,28 @@ There is no restriction on who can invoke this method.
 
 This method takes no input and returns 32 pseudo-random bytes to the caller. The return value is unknown to any part of the IC at time of the submission of this call. A new return value is generated for each call to this method.
 
+[#ic-get_ecdsa_public_key]
+=== IC method `get_ecdsa_public_key`
+
+This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
```
andreacerulli, line 1660: It should be transparent to the canister whether this is a threshold key.

```suggestion
This method returns the ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
```

spec/index.adoc
```diff
@@ -1654,6 +1654,28 @@ There is no restriction on who can invoke this method.
...
+The `derivation_path` is a vector of variable length byte strings.
+For https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032] compatibility, each byte string (`blob`) must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^ (non-hardened).
+The only supported `key_id` value at the moment is `"secp256k1"`.
+
+The return result also contains a `chain_code` which can be used for key derivation purpose if needed.
+
+This call requires that the ECDSA feature is enabled and working properly, and the `canister_id` meets the requirement of a canister id.
+Otherwise it will be rejected.
```
andreacerulli, line 1668: I wonder if we should really make this clarification, or just defer the publication of the spec to when the feature is enabled.

spec/index.adoc
andreacerulli, line 1676: This restriction applies to "secp256k1" (and possibly other future curves), so maybe it make sense to add this clarification

spec/index.adoc
```diff
@@ -1654,6 +1654,28 @@ There is no restriction on who can invoke this method.
...
+
+[#ic-sign_with_ecdsa]
+=== IC method `sign_with_ecdsa`
+
+This method returns a new threshold ECDSA signature of the given `message_hash` that can be separately verified against a derived ECDSA public key.
+This public key can be obtained by calling `get_ecdsa_public_key` with the caller's `canister_id`, and the same `derivation_path` and `key_id` used here.
+
+This call requires that the ECDSA feature is enabled and working properly, the caller is a canister, and `message_hash` is of 32-bytes long.
+Otherwise it will be rejected.
+
```
andreacerulli, line 1678: We should also specify the encoding of the signatures. E.g. `The signatures are encoded as the concatenation of the 32-byte big endian encodings of the two values r and s`. To be precise the 32 bytes are again specific to `secp256k1`. So maybe we can make the clarification that we only support this id for now, so we don't have to repeat the clarification for the hash and the output format. 

Otherwise we can rephrase the above with something like. `The signatures are encoded as the concatenation of the https://www.secg.org/sec2-v2.pdf[SEC1] encodings of the two values r and s. For key_id="secp256k1", this corresponds to 32-byte big-endian encoding`.

andreacerulli (commented):
Thanks a lot @ninegua! I left some suggestions and nits, but otherwise it looks good.

spec/ic.did
ninegua: I think all other APIs are using reject message to deliver errors. So I followed this convention here.

spec/index.adoc
```diff
@@ -1654,6 +1654,28 @@ There is no restriction on who can invoke this method.
 
 This method takes no input and returns 32 pseudo-random bytes to the caller. The return value is unknown to any part of the IC at time of the submission of this call. A new return value is generated for each call to this method.
 
+[#ic-get_ecdsa_public_key]
+=== IC method `get_ecdsa_public_key`
+
+This method returns the threshold ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
+The `derivation_path` is a vector of variable length byte strings.
+For https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032] compatibility, each byte string (`blob`) must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^ (non-hardened).
+The only supported `key_id` value at the moment is `"secp256k1"`.
```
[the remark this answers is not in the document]
ninegua: Thanks! Suggestion has been applied

spec/ic.did
Dfinity-Bjoern: As the `chain_code` allows to do arbitrary further derivations, shall we maybe make that an `opt blob` and return `null` unless the request comes from `canister_id = caller` (or `null`).
In particular, if we *always* return the chain code, there would be no need to support a `derivation_path` at all, and all derivation could be done in canister land.

spec/ic.did
```diff
@@ -44,6 +44,18 @@ service ic : {
   deposit_cycles : (record {canister_id : canister_id}) -> ();
   raw_rand : () -> (blob);
 
+  // Threshold ECDSA signature
+  get_ecdsa_public_key : (record {
```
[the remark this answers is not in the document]
Dfinity-Bjoern: I agree, we should be consistent with the naming we've used so far.

spec/index.adoc
Dfinity-Bjoern: If this is supposed to be _temporal_, then indeed we should take care using the versioning of the interface spec. (Each replica used on a given subnet implements some specified version of the interface spec.) If it's about being enabled on a subnet (so we'd switch the feature on and off per subnet), then I think the comment is appropriate. Do we plan the latter (except for using different versions during the roll-out)?

spec/ic.did
```diff
@@ -44,6 +44,18 @@ service ic : {
...
+    canister_id : opt canister_id;
+    derivation_path : vec blob;
+    key_id : text;
```
marcin-dziadus, line 51: Does that make sense to make `key_id` a variant?

nomeata: And maybe rename? I'd expect `key_id` to name a key (which is confusing as the derivation path does that already), not a scheme?

spec/ic.did
ninegua: The `derivation_path` is here for convenience to avoid having the caller implement derivation by themselves (which would require one set of library for each language we support). The `chain_code` returned is also here for convenience so that we don't force people to do many online calls to calculate a set of derivations.

At least that is my understand, please correct me if I'm wrong @andreacerulli

spec/ic.did
ninegua: Making it a variant means we'll have to modify the spec every time adding a new key_id.

The purpose is to identify which master key to use, as we might want to support more than one, or more than one type. Such details are not decided at the moment, so leaving it vague seems to the best we can do...

nomeata: So it does not identify the scheme? (“The only supported `key_id` value at the moment is `"secp256k1"`.” certainly sounds like it does.”)

If details are not known, then just leave the field out, and add an optional field later when there is clarity what can be specified and how?

nomeata: So it does not identify the scheme? (“The only supported `key_id` value at the moment is `"secp256k1"`.” certainly sounds like it does.”)

If details are not known, then just leave the field out, and add an optional field later when there is clarity what can be specified and how?

ninegua: Initially it is just "seckp256k1". New key_ids of the same scheme could be named as "secp256k1_1“. If using a different scheme, it could be "seckp256r1". Details are not decided yet, but we'd like to avoid having to add new APIs (or change type definitions) to use them.

Likely registry canister will be a place to look up supported key ids.

ninegua: Another way to put is, it is better to treat key_id is a runtime value instead of a statically defined enum.

nomeata: Maybe use a more helpful name initially that does not like it's defining the scheme? Start with `secp256k1_1` for the first existing master key? Or `20220226_secp256k`?

adambratschikaye: I also thought that the key_id might be referring to the scheme because the only supported one is `secp256k1`. Agree that it should have an name which makes it clear that there may be other `secp256k1` keys in the future.

spec/index.adoc
```diff
@@ -1654,6 +1654,31 @@ There is no restriction on who can invoke this method.
...
+This call requires that the ECDSA feature is enabled, and the `canister_id` meets the requirement of a canister id.
+Otherwise it will be rejected.
+
+[#ic-sign_with_ecdsa]
+=== IC method `sign_with_ecdsa`
+
+This method returns a new https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf[ECDSA] signature of the given `message_hash` that can be separately verified against a derived ECDSA public key.
+This public key can be obtained by calling `get_ecdsa_public_key` with the caller's `canister_id`, and the same `derivation_path` and `key_id` used here.
+
+The signatures are encoded as the concatenation of the https://www.secg.org/sec2-v2.pdf[SEC1] encodings of the two values r and s. For key_id="secp256k1", this corresponds to 32-byte big-endian encoding.
```
randombit, line 1677: Just the raise the issue:

The encoding described above is what we are currently using, and it is fairly common.

However many applications including both Bitcoin and X.509 use another representation which uses DER.

It is not terribly difficult to convert from one format to another, but it might be surprising for some users that this is required.

Should we switch to the DER encoding or stay with the current approach?

spec/ic.did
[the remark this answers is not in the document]
ninegua: Thanks! Points taken, and made the change.

spec/index.adoc
ninegua: As discussed on slack, we'll stick to the current encoding instead of DER.

spec/ic.did
Dfinity-Bjoern: But the `chain_code` would not be public usually. Given the proposed API, any canister can derive keys and get the `chain_code` for any other canister, which is something that, e.g., you couldn't know do for my Bitcoin keys. (Even if I had Bitcoin keys.)

spec/ic.did
Dfinity-Bjoern: It seems to me we are mixing two things here. The first one is the type of cryptographic scheme, which `secp256k1` seems to indicate, and for which I agree with @marcin-dziadus that a variant is a more natural representation. And I do think that changing the spec for additional versions we may be supporting (like P-256, P-384, ...) is perfectly fine, as long as the change is backward compatible.

The second aspect is supporting multiple keys of the same type. Here it's first unclear to me why the support via the `derivation_path` wouldn't be sufficient to solve that as well – is it because we are only supporting non-hardened derivation? And *if* there is a reason to have multiple "named" keys of the same type, then I don't understand why they should be managed globally instead of per canister.

(Finally: I don't think the registry would be the right place to store such data, but that's not really a discussion for this PR anyway.)

manudrijvers: You can imagine that different subnets hold different secp256k1 master keys right? We then need a way to express that. We could split scheme and "name" into two fields, which together are the id of the key.

spec/ic.did
victorshoup-dfinity: This is an interesting question. In the bitcoin world, chain codes can be used to create public keys that are unlinkable. Unlinkable public keys was not something we planned on supporting. One reason is is that to obtain unlinkability, the root chain code has to be secret, and , well the Internet Computer is not very good at keeping secrets, so I'm not this is something we can achieve. Certainly not today. Right now, the root chain code is a hardcoded constant. Anybody with access to our source code can reverse engineer to get the root PK.

Now....we could have a canister supply its own root chain code. Maybe we can't guarantee secrecy now, but maybe someday we can. So we may want to design the API conservatively to allow for this in the future. But then we have to rethink lots of things. As suggested, we could not allow any principal besides the canister than controls the key to obtain the chain code. But to get true unlinkability, I think we would have to go a step further and not allow any principal besides the canister that controls the key to obtain the public key.

We could do that, but I'm not sure if that creates inconveniences and even security problems. To get another canister's PK, we have to ask that canister for it. Not only that, we have to trust the canister to not lie and maybe give us somebody else's public key. We've discussed this before, and while there are no obvious problems with this, I think it could create security footguns...I think people expect to have a reliable way to obtain a public key...but maybe I'm wrong about that. In a typical PKI without PoPs, I can always claim your key is my key, so that is kind of the same.

So I think the larger question is whether only the controlling canister can ask for the public key. If the answer to that is NO, then I think the chain code question is moot. If the answer is YES, then I think the canister can create unlinkable public keys by injecting a secret into the BIPS32 derivation path, and not telling anyone the full derivation path -- I think that is equivalent to secret chain codes.

Let me know if we need a zoom call to discuss.

spec/ic.did
victorshoup-dfinity: We leave open the possibility that we might have more than one master key per curve. Maybe there is even more than a constant number of them, in which case an enum would not suffice.  But the feeling was is that the key_id should just be a arbitrary name that gets mapped somehow by the system so that for a given key_id, we know: which subnets actually host that key, what is the curve type and parameters, etc.

nomeata: Arbitrary name is good, but `secp265k` doesn't look arbitrary :-)

Here is an arbitrary name, if someone needs one: `2022a`

spec/ic.did
victorshoup-dfinity: Actually, thinking about this some more, maybe the API as is provides unlinkability. I mean, if a canister wants to have an unlinkable public key, he can generate a subkey using a path that has some secret random element in it, derive the PK, and publish that PK but not the path itself...or at least not the secret random element. That would work. So I think the API is good as is. Also, if we change the API significantly, this would have to be done in concert with Dieter and the bitcoin integration team, as this could impact their designs.

spec/ic.did
manudrijvers: We could make `key_id` a record that has
- enum curve_id
- text key_name

this shows we can have multiple keys on the same curve, but does not encode some expectations of the curve id into the name. Wdyt?

nomeata: Definitely better! Why would the caller want or need to specify curve_id, though? (I.e. why not ust `key_name`?)

manudrijvers: I imagined that only key_id is unique, not the key_name, so you could have key_id1 = (secp256k1, "1"), and key_id2 = (Curve25519, "1"), and they are not equal. 

Why include the curve? I guess we need to have the relation somewhere, we could introduce a separate map somewhere, but that seems more complex to me.

Dfinity-Bjoern: > You can imagine that different subnets hold different secp256k1 master keys right? We then need a way to express that.

IMO, the right way of solving this would be to make the functions available through a canister (not a virtual one, an actual canister) running on the ECDSA subnet. If there is a second subnet with ECDSA, then it would just run another canister and the addressing of "which key do you want to use" is via which canister you call.

> We could split scheme and "name" into two fields, which together are the id of the key, would you prefer that?

I think that makes more sense, at least it's closer to the intent that you described.

nomeata: I imagined that the key name _is_ unique across the keys of the IC, and not just across the keys of the IC of a particular curve. What's gained but confusion to re-use names?

But all in all it seems easier and cleaner to just omit this field now, and adding an optional field later if and when we have multiple keys; presumably then we have more clarity on how to identify keys.

(BTW, wouldn't the public key be the natural way to identify a key pair? Or an abridged version of that, i.e. a fingerprint? GPG and SSH seem to do that?)

bogwar: +1 to what Joachim proposes above: why not keep simpler for now?

dietersommer: There's the following reason why we would like to have the key_name field already now:
We think that it might be a good idea to first launch threshold ECDSA to the community only with a test key and let people play around with it. This can, e.g., be used with a first Bitcoin feature release that only connects to the Bitcoin test network. This would help us in running the threshold ECDSA feature in a kind of a "beta" release before we make the release with a production key with which people associate actual monetary value. Having a string-based key id would allow us to name it accordingly, like "temporary_test_key" to reduce the chance as much as possible that people use the test key for storing real value, e.g., Bitcoin.
Also, we think that having defaults is not optimal and creates risk. It's better if the application needs to clearly specify which key it uses. Defaults also become increasingly harder in the longer run.

nomeata: Well, a name `temporary_test_key` wouldn't have caused this discussion; `secp265k` (which doesn't look like a key name to me, let alone a temporary one) has, though :-)

adambratschikaye: > IMO, the right way of solving this would be to make the functions available through a canister (not a virtual one, an actual canister) running on the ECDSA subnet. If there is a second subnet with ECDSA, then it would just run another canister and the addressing of "which key do you want to use" is via which canister you call.

I don't really understand how this would work with sharing keys between subnets. If two subnets both have keys with the same name, how am I supposed to know if it's the same key shared between the subnets, or if they are different?

spec/ic.did
ninegua: Just to summarize the latest offline discussion from the ecdsa weekly meeting, the master ECDSA public keys will be made available in the registry registry, which is public. So anyone would be free to use that information to derive canister public keys given sufficient time of studying our implementation. We tend to agree that trying to hide this from the public interface is only a obscurity which gives a false sense of privacy.

So the conclusion is to keep this interface as is. Any strong objections? @victorshoup-dfinity @Dfinity-Bjoern

andreacerulli: > Right now, the root chain code is a hardcoded constant. Anybody with access to our source code can reverse engineer to get the root PK.

Since individual chain codes are publicly computable from the master key of the subnet (which can be computed from the registry), I don't see a strong reason to hide them from the API. Maybe we even give a false sense of privacy to the caller if we only return their own chain codes. @Dfinity-Bjoern are you still of the opinion that we should not disclose them?

andreacerulli (approved):

spec/ic.did
nomeata: Yes, it does! (And while I understand why a developer might want to choose a specific curve, I wonder why a developer needs to be bothered with choosing among one of many key names … but maybe I’ll shut up and _not_ suggest to make that field optional to make life simpler for the majority of users who just want to sign stuff :))

spec/ic.did
Dfinity-Bjoern: Otherwise, seems fine, so I'll close the thread.

Dfinity-Bjoern (approved):
Thanks for the careful consideration of all discussions and feedback!

jwiegley (approved):

nomeata:
Did we intentionally switch to non-squash merges now, or is that simply not discussed yet?

(I'd be in favor of re-enabling mergify; automatic squash merge with PR-description-as-commit-message seems to be the simplest nicest option.)

PR #7, opened 2022-01-27
IC-530:Canister HTTP requests

ielashi:
This PR provides the specifications for the feature IC-530: Canister HTTP requests.

The spec explains what are the constraints without explaining intricacies related to the underlying protocol. 

For this first iteration:
* the maximal size of the response is defined to `2MiB` (maximal size of a message in the consensus queue). 
* no mentions of cycles-related charges; Specifying additional cost that this call will incur should be added in the developers' guide, [Computation costs](https://smartcontracts.org/docs/developers-guide/computation-and-storage-costs.html ), once the cost is established.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -16,6 +16,20 @@ type definite_canister_settings = record {
   freezing_threshold : nat;
 };
 
+type http_header = record { 0: text; 1: text };
+
+type http_response = record {
+  status: nat;
+  headers: vec http_header;
+  body: blob;
+};
+
+type http_request_error = variant {
+  no_consensus;
+  timeout;
+  bad_tls;
+};
+
 service ic : {
   create_canister : (record {
     settings : opt canister_settings
@@ -43,6 +57,15 @@ service ic : {
   delete_canister : (record {canister_id : canister_id}) -> ();
   deposit_cycles : (record {canister_id : canister_id}) -> ();
   raw_rand : () -> (blob);
+  http_request : (record {
+    url : text;
+    method : variant { get };
+    headers: vec http_header;
+    body : opt blob;
+    transform : opt variant {
+      function: func (http_response) -> (http_response) query
+    };
+  }) -> (variant { Ok : http_response; Err: opt http_request_error });
 
   // provisional interfaces for the pre-ledger world
   provisional_create_canister_with_cycles : (record {
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -16,7 +16,7 @@ type definite_canister_settings = record {
   freezing_threshold : nat;
 };
 
-type http_header = record { 0: text; 1: text };
+type http_header = record { name: text; value: text };
 
 type http_response = record {
   status: nat;
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -28,6 +28,7 @@ type http_request_error = variant {
   no_consensus;
   timeout;
   bad_tls;
+  invalid_url;
 };
 
 service ic : {
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -29,6 +29,10 @@ type http_request_error = variant {
   timeout;
   bad_tls;
   invalid_url;
+  transform_error;
+  dns_error;
+  unreachable;
+  conn_timeout;
 };
 
 service ic : {
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1659,7 +1659,7 @@ This method takes no input and returns 32 pseudo-random bytes to the caller. The
 
 This method makes an HTTP request to a given URL and returns the HTTP response, possibly after a transformation.
 
-For security, the request is executed by multiple IC replicas (usually the entire subnet) and therefore must be _idempotent_. This means that the request must not change state at the remote server, or that the remote server has means to identify duplicated requests and return the same response to all of them.
+For security, the request is executed by multiple IC replicas (usually the entire subnet) and therefore must be _idempotent_. This means that the request must not change state at the remote server, or that the remote server has means to identify duplicated requests.
 
 The responses for all the requests must be, eventually, identical. However, it is possible that a web service would return slightly different responses for the same idempotent request. For example, it may include some unique identification or a timestamp that would vary across responses.
 
@@ -1676,7 +1676,7 @@ The maximal size of a request URL is 2048 bytes.
 The maximal size of a response is `TBD`. If a response is larger than this size, only the first `TBD` bytes will be returned. This size limit also applies to the value returned by the `transform` function.
 // NOTE: the size above should be up to 2MB which is the maximal size of a message in the consensus queue
 
-The follwing paramters should be supplied for the call:
+The following parameters should be supplied for the call:
 
 - `url` - the requested URL
 - `method` - currently only GET is supported
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1530,7 +1530,6 @@ The binary encoding of arguments and results are as per Candid specification.
 Before deploying a canister, the administrator of the canister first has to register it with the IC, to get a canister id (with an empty canister behind it), and then separately install the code.
 
 The optional `settings` parameter can be used to set the following settings:
-
 * `controllers` (`vec principal`)
 +
 A list of principals. Must be between 0 and 10 in size. This value is assigned to the _controllers_ attribute of the canister.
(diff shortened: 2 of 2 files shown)


spec/ic.did
```diff
@@ -16,6 +16,20 @@ type definite_canister_settings = record {
...
+type http_response = record {
+  status: nat;
+  headers: vec http_header;
+  body: blob;
+};
+
+type http_request_error = variant {
+  no_consensus;
+  timeout;
+  bad_tls;
```
ielashi, line 30: Reading the spec, I think we need a couple of more errors:

```
malformed_url; // in case the url isn't valid
not_https; // url is valid, but isn't https
```

Both of these cases can arguably be merged into a single error as well. No strong opinions on my side.

spec/ic.did
```diff
@@ -43,6 +57,15 @@ service ic : {
   delete_canister : (record {canister_id : canister_id}) -> ();
   deposit_cycles : (record {canister_id : canister_id}) -> ();
   raw_rand : () -> (blob);
+  http_request : (record {
+    url : text;
+    method : variant { get };
+    headers: vec http_header;
+    body : opt blob;
+    transform : opt variant {
```
marcin-dziadus, line 65: Why is `transform` not declared as just `opt func`?

ielashi: In case we want to offer different ways of supplying a `transform` in the future.

spec/index.adoc
```diff
@@ -1654,6 +1654,46 @@ There is no restriction on who can invoke this method.
 
 This method takes no input and returns 32 pseudo-random bytes to the caller. The return value is unknown to any part of the IC at time of the submission of this call. A new return value is generated for each call to this method.
 
+[#ic-http_request]
+=== IC method `http_request`
+
+This method makes an HTTP request to a given URL and returns the HTTP response, possibly after a transformation.
+
+For security, the request is executed by multiple IC replicas (usually the entire subnet) and therefore must be _idempotent_. This means that the request must not change state at the remote server, or that the remote server has means to identify duplicated requests and return the same response to all of them.
```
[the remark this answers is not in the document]
yotamhc: I think that if we don't tell the users that the request is executed multiple times they have no way to understand that from the rest of the text. I agree that the `same response` part is incorrect here.

spec/ic.did
```diff
@@ -29,6 +29,10 @@ type http_request_error = variant {
   timeout;
   bad_tls;
   invalid_url;
+  transform_error;
+  dns_error;
+  unreachable;
```
ielashi, line 34: Question: what does `unreachable` mean here?

spec/index.adoc
```diff
@@ -1654,6 +1654,51 @@ There is no restriction on who can invoke this method.
...
+
+The responses for all the requests must be, eventually, identical. However, it is possible that a web service would return slightly different responses for the same idempotent request. For example, it may include some unique identification or a timestamp that would vary across responses.
+
+For this reason, the calling canister can supply a transformation function, which is used by the IC to let the canister sanitize the responses from such unique values, and therefore eventually agree on an identical response. The transformation function is executed separately by each replica on the corresponding response it has received for the request. The final response will only be available to the calling canister if enough replicas have agreed on an identical response.
+
+Currently, only GET method is supported for HTTP requests.
+
+For security reasons, only HTTPS connections are allowed (URLs must start with `https://`). The IC uses industry-standard root CA lists to validate certificates of remote web servers.
```
Dfinity-Bjoern, line 1670: The description of the CA list is a bit vague. If we don't want to enumerate the list, we should at least link where we get it from.

spec/index.adoc
```diff
@@ -1654,6 +1654,51 @@ There is no restriction on who can invoke this method.
...
+
+The cost of each call to the `http_request` method is `TBD` (the `transform` should also be charged).
```
Dfinity-Bjoern, line 1672: We expect that to be sent in cycles with the call to the IC management canister? (Would make sense to write that explicitly.)

spec/ic.did
```diff
@@ -43,6 +62,15 @@ service ic : {
   delete_canister : (record {canister_id : canister_id}) -> ();
   deposit_cycles : (record {canister_id : canister_id}) -> ();
   raw_rand : () -> (blob);
+  http_request : (record {
+    url : text;
+    method : variant { get };
+    headers: vec http_header;
+    body : opt blob;
+    transform : opt variant {
+      function: func (http_response) -> (http_response) query
```
Dfinity-Bjoern, line 71: I am not sure how exactly that is implemented. Is this safe for canister upgrades?

spec/ic.did
yotamhc: That the remote server was unreachable

spec/index.adoc
yotamhc: As far as I understand from Helge, we'll use Linux's standard certificate validation. But I am not sure how to write this here.

spec/ic.did
ielashi: @Dfinity-Bjoern What do you think would be unsafe here in case of an upgrade?

To my knowledge a `func` in candid is represented by a principal + a method name, and only the canister's own principal would be accepted here. If a canister is upgraded and the method being referenced here is, for example, removed, then the call to the transform will fail.

spec/ic.did
```diff
@@ -43,6 +62,15 @@ service ic : {
...
+    };
+  }) -> (variant { Ok : http_response; Err: opt http_request_error });
```
nomeata, line 73: If there is a transformation function anyways, there is little reason to return an `http_response`. Assume you use this to scrape some data off a webpage. Surely, the `transform` function will parse the result, find the relevant data (say, the temperature in Palo Alto) and will want to return _just_ that. This will also make the message much smaller.

With the current interface it would have to artificially wrap it in a `http_response`.

So an alternative is to leave the resulting type abstract. Unfortunately, [Candid does not (yet) have generics](https://github.com/dfinity/candid/issues/245), so `blob` might be a better choice:
```suggestion
      function: func (http_response) -> (blob) query
    };
  }) -> (variant { Ok : blob; Err: opt http_request_error });
```

This would make the transform _required_, but in practice, I woudln’t be surprised if any real use will need a `transform` anyways. Most APIs, for example, return headers indicating throttling information. So while toy do get more complicated.

Of course, `http_response` contains a `blob` in the body, and this could be (ab?)used instead.

spec/index.adoc
```diff
@@ -1654,6 +1654,51 @@ There is no restriction on who can invoke this method.
 
 This method takes no input and returns 32 pseudo-random bytes to the caller. The return value is unknown to any part of the IC at time of the submission of this call. A new return value is generated for each call to this method.
 
+[#ic-http_request]
+=== IC method `http_request`
+
+This method makes an HTTP request to a given URL and returns the HTTP response, possibly after a transformation.
+
+For security, the request is executed by multiple IC replicas (usually the entire subnet) and therefore must be _idempotent_. This means that the request must not change state at the remote server, or that the remote server has means to identify duplicated requests.
```
nomeata, line 1662: Any chance to describe the behavior of this feature without saying “replica” and “subnet”? Or are we giving up on having that level of abstraction in our programming model?

spec/index.adoc
```diff
@@ -1654,6 +1654,51 @@ There is no restriction on who can invoke this method.
...
+
+Currently, only GET method is supported for HTTP requests.
+
+For security reasons, only HTTPS connections are allowed (URLs must start with `https://`). The IC uses industry-standard root CA lists to validate certificates of remote web servers.
+
+The cost of each call to the `http_request` method is `TBD` (the `transform` should also be charged).
+
+The maximal size of a request URL is 2048 bytes.
+
+The maximal size of a response is `TBD`. If a response is larger than this size, only the first `TBD` bytes will be returned. This size limit also applies to the value returned by the `transform` function.
```
nomeata, line 1676: “also”? Or “only”? I'd expect that larger responses can be processed and trimmed down by the transform function, as we really only have a problem with large messages when they enter consensus, and that’s post-`transform`.

spec/index.adoc
AlexandraZapuc: Removed this cost related line(please see the description).

spec/index.adoc
AlexandraZapuc: @yotamhc replaced `TBD` with `2MiB`, we can modify this value later if needed, but for the first iteration, I think it's fine. Please see also Joachim's suggestion here. Is this something that we are planning to do now or maybe later, in the following iterations?

spec/ic.did
AlexandraZapuc: Given that in the specifications is mentioned that the canister can modify headers too,  I will leave `http_response` as a return type here.

spec/index.adoc
```diff
@@ -1654,6 +1654,40 @@ There is no restriction on who can invoke this method.
 
 This method takes no input and returns 32 pseudo-random bytes to the caller. The return value is unknown to any part of the IC at time of the submission of this call. A new return value is generated for each call to this method.
 
+[#ic-http_request]
+=== IC method `http_request`
+
+This method makes an HTTP request to a given URL and returns the HTTP response, possibly after a transformation.
+
+The request must be _idempotent_, meaning that it must not change the state at the remote server, or the remote server has the means to identify duplicated requests.
```
JensGroth, line 1662: The canister has no means to ensure the request is idempotent, should we soften it to: 'the canister should aim to issue idempotent requests since otherwise the risk of failure increases'.

spec/index.adoc
yotamhc: @nomeata We also don't want to download too big responses (before the `transform`). We could have another limit for that, but opted to keep it simple for now. Do you think it's crucial to allow larger responses?

AlexandraZapuc:
@JensGroth or @Dfinity-Bjoern, could you please let me know if there are any changes required before merging this PR? cc: @yotamhc

yotamhc:
@AlexandraZapuc should we merge this PR before the feature is available?

nomeata:
Yes! At least originally the plan was:

 1. Feature is spec’d in a PR, and agreed upon
 2. Feature is merged into `master`
 3. New verison of the spec is published 
 4. Implementation work starts
 5. New replica version supporting new version of the spec is released.

I understand that for pragmatic reasons 4. happens first usually :-). Nevertheless, because this repo and the doc is versioned, it should be fine to merge and release stuff here that isn't live yet.

JensGroth (approved):
LGTM

PR #14, opened 2022-03-16
Canister http request - Add a note about the expected caller id

AlexandraZapuc:
Follow up of Canister HTTP requests: https://github.com/dfinity/interface-spec/pull/7. 

Motivation: The implementation introduces an "internal query" that does not provide a caller id because the transformation function of the HTTP response is triggered by an internal component from the networking layer.

Given that the transform function is exported as `canister_query`, for consistency, we will return the canister management id if `msg_caller_copy` is called inside the transformation function.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1690,6 +1690,7 @@ The returned response (and the response provided to the `transform` function, if
 - `body` - the response's body
 
 The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
+When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
 
 [#ic-provisional_create_canister_with_cycles]
 === IC method `provisional_create_canister_with_cycles`


spec/index.adoc
```diff
@@ -1690,6 +1690,7 @@ The returned response (and the response provided to the `transform` function, if
 - `body` - the response's body
 
 The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
+When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
```
AlexandraZapuc, line 1693: @Dfinity-Bjoern Although not very likely, a developer could still ask for the caller id inside the transformation function, which is exposed to make the HTTP response deterministic. So, I talked with @crusso about this too, and we could return one of the options below: 
* the management canister id -> the canister sends an http request to the management canister, and then before returning the http response, as mentioned in the documentation, the exported function will be invoked for cleaning the response. So I think it would make sense to the developer to see that the caller id is the management canister. 
 * self canister id -> The canister itself is the one who initiated the http_request so this alternative would probably make sense for the developer too.  
 
 But I would prefer the first one, I think a developer would not be surprized if they see the management canister id when the function was invoked as a result of a canister http request.

nomeata (approved):

PR #15, opened 2022-03-17
Canister status - Add flag low on cycles

AlexandraZapuc:
The developers recently requested a mechanism that would quickly show if the canister is frozen without needing an update call to the canister to find out it was rejected due to low balance.   

Since it is not so simple to say exactly when the canister is going to start behaving like a frozen one unless the balance is below the freezing threshold. But if the balance is slightly above, it may not have enough to perform an operation and still have the leftover balance above the threshold.

So this PR proposes to add a flag that marks the canister as `low on cycles` when the balance is smaller than the sum of the freezing threshold in cycles plus a **flat amount of additional cycles**:

canister.balance < canister.freezing_threshold_cycles + N *  MAX_CYCLES_PER_MESSAGE

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2848,6 +2848,10 @@ The controllers of a canister can obtain information about the canister.
 
 The `Memory_size` is the (in this specification underspecified) total size of storage in bytes.
 
+A canister is considered low on cycles when the balance approaches the equivalent value in cycles of the freezing threshold.
+This means that soon the canister will start acting like a frozen one.
+For example, it may not be able to perform calls due to the cost of the call and transferred cycles which push the balance into frozen territory.
+
 Conditions::
 ....
     S.messages = Older_messages · CallMessage M · Younger_messages
@@ -2872,6 +2876,10 @@ S with
           controllers = S.controllers[A.canister_id];
           memory_size = Memory_size;
           cycles = S.balance[A.canister_id];
+          low_on_cycles =
+            if S.balance[A.canister_id] < (freezing_limit(S, A.canister_id) + 100 * MAX_CYCLES_PER_MESSAGE)
+            then true
+            else false
         })
         refunded_cycles = M.transferred_cycles
       }
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -58,6 +58,7 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
+      low_on_cycles: bool;
   });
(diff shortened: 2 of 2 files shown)


spec/index.adoc
```diff
@@ -2872,6 +2876,10 @@ S with
           controllers = S.controllers[A.canister_id];
           memory_size = Memory_size;
           cycles = S.balance[A.canister_id];
+          low_on_cycles =
+            if S.balance[A.canister_id] < (freezing_limit(S, A.canister_id) + 100 * MAX_CYCLES_PER_MESSAGE)
```
AlexandraZapuc, line 2880: > canister.balance < canister.freezing_threshold_cycles + N * MAX_CYCLES_PER_MESSAGE 

I chose N to be 100, but it might be too low. Happy to hear other proposals.

MAX_CYCLES_PER_MESSAGE does not have a concrete value defined in the spec, but it is used to show that the balance is approaching the freezing threshold.

ulan:
@AlexandraZapuc: what if we expose the current value of `freezing_threshold_in_cycles` so that the users can query that and the balance and decided themselves whether they are "low on cycles" or not? That way we don't have to define and explain `N`.

AlexandraZapuc:
Yes, that was my initial intention to only expose the amount of cycles equivalent to the freezing threshold. However, the developer would prefer not to do any back-of-the-envelope calculations. So exposing a true or false flag would take complexity away from the users. They see that the canister is approaching the limit and they have the chance to top up.

ielashi:
> So exposing a true or false flag would take complexity away from the users. They see that the canister is approaching the limit and they have the chance to top up.

@AlexandraZapuc I think there's a distinction to be made here between the interface that we expose in the protocol, and the interface being exposed to users. They are not necessarily the same thing.

The interface of the protocol should _generally_ be designed to be as generic as possible. Whatever mechanism we decide for what it means to be low on cycles may suit some developers and not suit others. Convenience, on the other hand, has a lower priority imho, because we can make the API more convenient on the SDK layer or in user libraries.

Per Ulan's suggestion, if we expose the `freezing_threshold_in_cycles`, then that's the most generic and it's up to the developers to infer from that what "low on cycles" means to them. I agree it's less convenient, but the SDK can easily use the information to determine and promptly to users if they are lower on cycles, or developers can build libraries that exposes information on whether a canister is low on cycles, and each has the freedom to use whatever definition of "low on cycles" that suits them.

PR #18, opened 2022-03-22
Canister status - Add Freezing threshold in cycles

AlexandraZapuc:
This MR enhanced the status of the canister with the freezing threshold in cycles. 
 
This data will be used by the developers and can be used to check when the canister is approaching the freeze threshold in cycles which would mean that the canister may start behaving like a frozen one. In this case, certain operations may fail because there is not enough balance to perform the operation and still have the leftover cycles above the threshold.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -59,7 +59,7 @@ service ic : {
       memory_size: nat;
       cycles: nat;
       freezing_threshold: nat;
-      freezing_threshold_cycles: nat;
+      freezing_threshold_in_cycles: nat;
   });
   delete_canister : (record {canister_id : canister_id}) -> ();
   deposit_cycles : (record {canister_id : canister_id}) -> ();
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -58,6 +58,8 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
+      freezing_threshold: nat;
+      freezing_threshold_cycles: nat;
   });
   delete_canister : (record {canister_id : canister_id}) -> ();
   deposit_cycles : (record {canister_id : canister_id}) -> ();
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2849,8 +2849,7 @@ The controllers of a canister can obtain information about the canister.
 
 The `Memory_size` is the (in this specification underspecified) total size of storage in bytes.
 
-The `freezing_threshold_cycles` is the freezing threshold in cycles of the canister, which is computed based on its current memory footprint, storage cost, memory and compute allocation, and current `freezing_threshold` setting.
-
+The `freezing_threshold_in_cycles` is the current estimate of the smallest amount of cycles the canister can have without entering the frozen state. The system uses the `freezing_threshold` setting to compute this field.
 
 Conditions::
 ....
@@ -2877,7 +2876,7 @@ S with
           memory_size = Memory_size;
           cycles = S.balance[A.canister_id];
           freezing_threshold = S.freezing_threshold[A.canister_id];
-          freezing_threshold_cycles = freezing_limit(S, A.canister_id);
+          freezing_threshold_in_cycles = freezing_limit(S, A.canister_id);
         })
         refunded_cycles = M.transferred_cycles
       }
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2849,7 +2849,7 @@ The controllers of a canister can obtain information about the canister.
 
 The `Memory_size` is the (in this specification underspecified) total size of storage in bytes.
 
-The `freezing_threshold_in_cycles` is the current estimate of the smallest amount of cycles the canister can have without entering the frozen state. The system uses the `freezing_threshold` setting to compute this field.
+The `freezing_threshold_in_cycles` is the current estimate of the smallest number of cycles the canister can have without entering the frozen state. The system uses the `freezing_threshold` setting to compute this field.
 
 Conditions::
 ....
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2849,6 +2849,9 @@ The controllers of a canister can obtain information about the canister.
 
 The `Memory_size` is the (in this specification underspecified) total size of storage in bytes.
 
+The `freezing_threshold_cycles` is the freezing threshold in cycles of the canister, which is computed based on its current memory footprint, storage cost, memory and compute allocation, and current `freezing_threshold` setting.
+
+
 Conditions::
 ....
     S.messages = Older_messages · CallMessage M · Younger_messages
@@ -2873,6 +2876,8 @@ S with
           controllers = S.controllers[A.canister_id];
           memory_size = Memory_size;
           cycles = S.balance[A.canister_id];
+          freezing_threshold = S.freezing_threshold[A.canister_id];
+          freezing_threshold_cycles = freezing_limit(S, A.canister_id);
         })
         refunded_cycles = M.transferred_cycles
       }


spec/ic.did
```diff
@@ -58,6 +58,8 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
+      freezing_threshold: nat;
+      freezing_threshold_cycles: nat;
```
ulan, line 62: Wdyt about adding `in` to denotes the units: `freezing_threshold_in_cycles`

ulan (approved):
Lgtm, thanks!

spec/ic.did
AlexandraZapuc: Yes, that is fine with me. I will change it to your proposed naming.

spec/index.adoc
```diff
@@ -2849,6 +2849,9 @@ The controllers of a canister can obtain information about the canister.
 
 The `Memory_size` is the (in this specification underspecified) total size of storage in bytes.
 
+The `freezing_threshold_cycles` is the freezing threshold in cycles of the canister, which is computed based on its current memory footprint, storage cost, memory and compute allocation, and current `freezing_threshold` setting.
```
roman-kashitsyn, line 2852: I think we should explain the meaning of this value, it is not at all clear from the description. For example,

> The `freezing_threshold_in_cycles` is the current estimate of the smallest amount of cycles the canister can have without entering the frozen state. The system uses the `freezing_threshold` setting to compute this field.

spec/index.adoc
```diff
@@ -2849,6 +2849,8 @@ The controllers of a canister can obtain information about the canister.
 
 The `Memory_size` is the (in this specification underspecified) total size of storage in bytes.
 
+The `freezing_threshold_in_cycles` is the current estimate of the smallest amount of cycles the canister can have without entering the frozen state. The system uses the `freezing_threshold` setting to compute this field.
```
roman-kashitsyn, line 2852:
```suggestion
The `freezing_threshold_in_cycles` is the current estimate of the smallest number of cycles the canister can have without entering the frozen state. The system uses the `freezing_threshold` setting to compute this field.
```

Sorry, I should have suggested "number" originally: cycles are countable.

roman-kashitsyn (approved):

ielashi (approved):

JensGroth:
@AlexandraZapuc I'm lacking context for this update. 
What's the use case this change is addressing? 
Is the function freezing_limit stable enough to make the response useful?

AlexandraZapuc:
> @AlexandraZapuc I'm lacking context for this update. What's the use case this change is addressing? Is the function freezing_limit stable enough to make the response useful?

@JensGroth I have updated the description. More information was requested by the devs that would help them understand when a canister is frozen without them needing to do an update operation which is rejected due to insufficient cycles in the balance. 
 
 Yes, the function freezing_limit is stable, it is documented in the specification already. In the code, this will return the exact value that we use when rejecting due to insufficient balance.

Dfinity-Bjoern (commented):
Making the `freezing_threshold` accessible via the canister status definitely seems natural, since otherwise there is no way of retrieving it.

I am a bit skeptical about `freezing_threshold_in_cycles`, since that is a derived value and it may also be quite unreliable: If the canister storage increases one block later, then the `freezing_threshold_in_cycles` may change drastically.

@jwiegley what are your thoughts?

peterpeterparker:
Don't want to hijack this PR but yes as a developer of an application that runs on the IC, I can confirm I need to know what's the remaining threshold value that should be retained by a canister before being able to delete it - i.e I want my users to be able to delete there data - their canister - and this with as few cycles as needed.

Here for example the kind of surprised I had because I was not aware the threshold was dynamic discussed on the forum: https://forum.dfinity.org/t/cannot-delete-canisters-anymore/11594

Ping me for further context, question and demo if needed.

jwiegley:
@Dfinity-Bjoern I agree that I would rather see the "fundamental data" (threashhold and consumption) stored with the canister, rather than one fundamental datum, and another which is calculated from this and the other fundamental datum.

ulan:
> If the canister storage increases one block later, then the freezing_threshold_in_cycles may change drastically.
> @Dfinity-Bjoern I agree that I would rather see the "fundamental data" (threashhold and consumption)

We haven't thought about exposing consumption. Sounds like a good idea to me. Brainstorming the name: `idle_burn_rate_in_cycles_per_second` or `idle_cycle_consumption_per_second`. Note that consumption is derived from memory usage and compute allocation, so it changes as drastically as `freezing_threshold_in_cycles`.
One advantage of exposing consumption is that it is well defined for the system subnet: 0.

AlexandraZapuc:
I am not sure I understand what you mean by consumption in this case. 

> One advantage of exposing consumption is that it is well defined for the system subnet: 0. 
The threshold value in cycles is also 0 for the system subnet.

nomeata:
> Note that consumption is derived from memory usage and compute allocation

Are these two basic values already easily available? And the factors that go into it? (Reading the IC code doesn’t count :-))

ulan:
> The threshold value in cycles is also 0 for the system subnet.

You're right! Sorry I was confused about how we compute `freezing_threshold_in_cycles` from `freezing_threshold`.

> Are these two basic values already easily available? And the factors that go into it?

No :) Do you think we should expose the factors and let the users compute the consumption? I am worried that then it would become difficult to make adjustments in the consumption function and the factors.

AlexandraZapuc:
> No :) Do you think we should expose the factors and let the users compute the consumption? I am worried that then it would become difficult to make adjustments in the consumption function and the factors. 

Agree that if we do that it will become difficult to make any adjustments to the factors we use to compute the value of the freezing threshold. 

 Also, I would like to emphasize that the request for this change was initiated by the developers who many times faced `canister out of cycles` because their balance was low enough that they could not pay for an operation and that would have pushed the canister directly into the frozen territory.  
 My understanding is that they would want to have an **easy way** to understand when the canister is getting too close to the freezing limit (they don't have visibility on this yet). 
So given this, I am not sure why we are not trying to take this complexity away from the users.

AlexandraZapuc:
Please see a couple of requests/issues opened by the developers regarding this problem: 

https://forum.dfinity.org/t/how-to-know-if-a-canister-is-frozen/10629
https://forum.dfinity.org/t/cannot-delete-canisters-anymore/11594

AlexandraZapuc:
After a couple of discussions, we agreed to introduce a field that would specify what is the idle consumption of resources (currently memory and computation allocation) in cycles per second. The freezing threshold in cycles can be obtained using this formula: **freezing threshold in cycles * idle consumption in cycles per second**.
  
I would suggest to name this field: `idle_resource_consumption_in_cycles_per_second`.  I think that it is self-explanatory compared to the other ones.
Other alternatives:
* idle_cycles_burned_per_second 
* cycles_burned_per_second_idle
* idle_resource_consumption_in_cycles_per_second
* resource_cycles_burned_per_second
* idle_resource_cycles_burned_per_second
* cycles_burned_per_second_due_to_resources
* idle_cycle_burn_rate_in_cycles_per_second
* cycle_burn_rate_due_to_resources_in_cycles_per_second 

cc: @nomeata

nomeata:
That is very reasonable, and the new value will be of great interest on its own.

(In the formula you mean “freezing threshold in seconds“, right?)

As for the color of the bike shed, I like `idle_cycles_burned_per_second`

AlexandraZapuc:
> (In the formula you mean “freezing threshold in seconds“, right?)  

Ah yes, indeed, I corrected it. It was a typo.  

> As for the color of the bike shed, I like idle_cycles_burned_per_second

 Sure, I am ok with this one too. I will change the spec then. Thanks, Joachim @nomeata!

jwiegley (approved):

spec/ic.did
```diff
@@ -58,6 +58,8 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
+      freezing_threshold: nat;
+      freezing_threshold_in_cycles: nat;
```
stefan-at-dfinity, line 62: Maybe I misunderstand, but related to the above, does this belong into the settings record?

spec/ic.did
```diff
@@ -58,6 +58,8 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
+      freezing_threshold: nat;
```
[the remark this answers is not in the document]
AlexandraZapuc: Actually, you are right. When I was adding the other field, I had the impression we forgot to add the `freezing_threshold`, but I missed that it was added in the `definite_canister_settings`. Will remove it.

spec/ic.did
AlexandraZapuc: No, this is not something that can be set but it's computed by the IC based on the `freezing_threshold`. Also, there was a follow-up PR that was merged so it's not called `freezing_threshold_in_cycles` anymore.

PR #19, opened 2022-04-01
System API: Introduce a canister generation counter

nomeata:
## TL;DR

A counter bumped on each code installation is introduced. This is one ingredient to make safe instantaneous upgrades as described in <https://www.joachim-breitner.de/blog/789-Zero-downtime_upgrades_of_Internet_Computer_canisters> possible.

## What?

This change introduces a _canister generation counter_, a simple natural number incremented upon each canister code installation. It is available to the canister via a system api call.

Additionally, when handling a response, the canister is told at which generation it has issued the call. This allows the canister to recognize when it is handling a call across a reinstallation or upgrade.

## Why?

Reinstalling or upgrading a canister with outstanding callbacks is currently dangerous, as the late response could confuse or corrupt the canister. Canisters that choose to be instantaneously upgradable can now detect such an old response and handle it safely (possibly by simply trapping).

The concept introduced will likely be useful beyond this. For example, to observe autonomous canisters it may not be enough to know their _current_ code module hash, but it may also be important to be able to tell that the canister has not had a _different_ code re-installed in between. Observing that the generation counter has not changed is one way to check that no (possibly malferious) code has been installed.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1166,7 +1166,7 @@ During the canister upgrade process, `canister_pre_upgrade` sees the old counter
 
 * `ic0.callback_generation : () -> i64`
 +
-returns the generation of canister at the time of invoking the `ic0.call_perform` that performed the call which response is handled right now. It allows canisters to recognize that there is been a canster upgrade or installation between issuing a call and handling the response. Will trap unless called from a reply or reject handler.
+returns the generation counter value of the canister at the time of invoking the `ic0.call_perform` for the call that led to the response that is handled right now. It allows canisters to recognize that there is been a canster upgrade or installation between issuing a call and handling the response. Will trap unless called from a reply or reject handler.
 
 [#system-api-call]
 === Inter-canister method calls
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -190,6 +190,7 @@ Dapps on the Internet Computer are called _canisters_. Conceptually, they consis
  * A canister id (a <<principal,principal>>)
  * Their _controllers_ (a possibly empty list of <<principal,principal>>)
  * A cycle balance
+ * A <<system-api-generation,generation counter>>
  * The _canister status_, which is one of `running`, `stopping` or `stopped`.
  * Resource reservations
 
@@ -216,7 +217,7 @@ When the cycle balance of a canister falls to zero, the canister is _deallocated
 
 Afterwards the canister is empty. It can be reinstalled after topping up its balance.
 
-NOTE: Once the IC frees the resources of a canister, its id, _cycles_ balance, and _controllers_ are preserved on the IC for a minimum of 10 years. What happens to the canister after this period is currently unspecified.
+NOTE: Once the IC frees the resources of a canister, its id, _cycles_ balance, _controllers_ are preserved on the IC for a minimum of 10 years. What happens to the canister after this period is currently unspecified.
 
 [#canister-status]
 ==== Canister status
@@ -950,6 +951,7 @@ ic0.msg_arg_data_size : () -> i32;                                          // I
 ic0.msg_arg_data_copy : (dst : i32, offset : i32, size : i32) -> ();        // I U Q Ry F
 ic0.msg_caller_size : () -> i32;                                            // I G U Q F
 ic0.msg_caller_copy : (dst : i32, offset: i32, size : i32) -> ();           // I G U Q F
+ic0.callback_generation : i64;                                              // Ry Rt
 ic0.msg_reject_code : () -> i32;                                            // Ry Rt
 ic0.msg_reject_msg_size : () -> i32;                                        // Rt
 ic0.msg_reject_msg_copy : (dst : i32, offset : i32, size : i32) -> ();      // Rt
@@ -962,7 +964,7 @@ ic0.msg_cycles_available : () -> i64;                                       // U
 ic0.msg_cycles_available128 : (dst : i32) -> ();                            // U Rt Ry
 ic0.msg_cycles_refunded : () -> i64;                                        // Rt Ry
 ic0.msg_cycles_refunded128 : (dst : i32) -> ();                             // Rt Ry
-ic0.msg_cycles_accept : (max_amount : i64) -> ( amount : i64 );            // U Rt Ry
+ic0.msg_cycles_accept : (max_amount : i64) -> ( amount : i64 );             // U Rt Ry
 ic0.msg_cycles_accept128 : (max_amount_high : i64, max_amount_low: i64, dst : i32)
                        -> ();                                               // U Rt Ry
 
@@ -971,6 +973,7 @@ ic0.canister_self_copy : (dst : i32, offset : i32, size : i32) -> ();       // *
 ic0.canister_cycle_balance : () -> i64;                                     // *
 ic0.canister_cycle_balance128 : (dst : i32) -> ();                          // *
 ic0.canister_status : () -> i32;                                            // *
+ic0.canister_generation : () -> i64;                                        // *
 
 ic0.msg_method_name_size : () -> i32                                        // F
 ic0.msg_method_name_copy : (dst : i32, offset : i32, size : i32) -> ();     // F
@@ -988,8 +991,8 @@ ic0.call_new :                                                              // U
   ) -> ();
 ic0.call_on_cleanup : (fun : i32, env : i32) -> ();                         // U Ry Rt H
 ic0.call_data_append : (src : i32, size : i32) -> ();                       // U Ry Rt H
-ic0.call_cycles_add : (amount : i64) -> ();                               // U Ry Rt H
-ic0.call_cycles_add128 : (amount_high : i64, amount_low: i64) -> ();      // U Ry Rt H
+ic0.call_cycles_add : (amount : i64) -> ();                                 // U Ry Rt H
+ic0.call_cycles_add128 : (amount_high : i64, amount_low: i64) -> ();        // U Ry Rt H
 ic0.call_perform : () -> ( err_code : i32 );                                // U Ry Rt H
 
 ic0.stable_size : () -> (page_count : i32);                                 // *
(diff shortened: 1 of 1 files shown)


nomeata:
Opening this as a draft until the CLC can be ticked off.

crusso:
Nice idea!

jwiegley:
This looks like a good idea to me.

spec/index.adoc
```diff
@@ -1150,6 +1153,21 @@ Status `1` indicates running, `2` indicates stopping, and `3` indicates stopped.
 +
 Status `3` (stopped) can be observed, for example, in `canister_pre_upgrade` and can be used to prevent accidentally upgrading a canister that is not fully stopped.
 
+[#system-api-generation]
+=== Canister generation counter
+
+For each canister, the system maintains a _generation counter_. Upon canister generation, it is set to 0, and it is incremented upon every successful code installation (see <<ic-install_code,code installation>>).
```
Dfinity-Bjoern, line 1159:
```suggestion
For each canister, the system maintains a _generation counter_. Upon canister creation, it is set to 0, and it is incremented upon every successful code installation (see <<ic-install_code,code installation>>).
```

Dfinity-Bjoern (commented):
Thanks @nomeata – this is a very useful suggested feature! We may want to also expose the generation in the system state tree, so that external users can indeed track whether a canister has been updated.

@bogwar @ulan Any comments?

nomeata:
Yes, exposing in the state tree is reasonable. I was about to add it myself, but then hesitated, to keep the PR more focused, but also because it's possible for canisters to expose that data selectively in a query method. Together with the certified module hash, that's even verifiable, albeit it's simpler directly in the tree of course.

ulan:
I like the idea. The PR lgtm.

ali-dfinity:
droid update cla

nomeata:
BTW, while this is conceptually self-contained, to achieve the goal of zero-downtime-upgradeable-canisters, I also need to write up the “named callback entry points” feature. I wouldn’t mind keeping this open until both are done and decided upon, but if you think this is useful anyways, we can of course merge this.

origyndev:
Another vote for this. Commenting to follow.

levifeldman:
+1

spec/index.adoc
```diff
@@ -4045,6 +4089,9 @@ ic0.canister_self_size<es>() : i32 =
 ic0.canister_self_copy<es>(dst:i32, offset:i32, size:i32) =
   copy_to_canister<es>(dst, offset, size, es.wasm_state.self_id)
 
+ic0.canister_generatoin<es>() : i64 =
```
jzxchiang1, line 4092: Typo

spec/index.adoc
```diff
@@ -1150,6 +1153,21 @@ Status `1` indicates running, `2` indicates stopping, and `3` indicates stopped.
...
+
+* `ic0.canister_generation : () -> i64`
++
+returns the current value of the generation counter.
++
+During the canister upgrade process, `canister_pre_upgrade` sees the old counter value, and `canister_post_upgrade` sees the new counter value.
+
+* `ic0.callback_generation : () -> i64`
++
+returns the generation of canister at the time of invoking the `ic0.call_perform` that performed the call which response is handled right now. It allows canisters to recognize that there is been a canster upgrade or installation between issuing a call and handling the response. Will trap unless called from a reply or reject handler.
```
jzxchiang1, line 1169: Can we be more specific here, i.e. "returns the generation counter of the canister"?

I'm also not sure why we need both a `ic0.canister_generation` and a `ic0.callback_generation` function. Is it not possible to combine those into a single one?

nomeata: We certainly need both: the former is the current generation, the latter the generation of the canister that _sent_ the call that led to the currently handled callback.

```suggestion
returns the generation counter value of the canister at the time of invoking the `ic0.call_perform` for the call that led to the response that is handled right now. It allows canisters to recognize that there is been a canster upgrade or installation between issuing a call and handling the response. Will trap unless called from a reply or reject handler.
```

spec/index.adoc
nomeata:
```suggestion
ic0.canister_generation<es>() : i64 =
```

Dfinity-Bjoern:
> Yes, exposing in the state tree is reasonable. I was about to add it myself, but then hesitated, to keep the PR more focused, but also because it's possible for canisters to expose that data selectively in a query method. Together with the certified module hash, that's even verifiable, albeit it's simpler directly in the tree of course.

That's a good point. But it does require reproducible builds to work as well, right? Of course having the generation without reproducible build is far less compelling anyway, but it also means the user would actually:

- check (meaning: read!) the source code
- check the reproducible build
- query the canister

So I still do see value in exposing the generation in the state tree, but I also agree it's not literally required.

skilesare:
> 



> * check (meaning: read!) the source code
> * check the reproducible build
> * query the canister

We have developing workflows that would use this.  The OGY network will certify wasm hashes of open-source canisters and allow canisters running those hashes to join the "network".  Theoretically, those are trustworthy marking off 1 and 2, as for three we'd love for these canisters to have plug and play certification that is easy for a dev to expose without having to figure it all out themselves.  I've been trying and it isn't trivial.  I could be misunderstanding this, but I think this gets me something I've wanted since day 1.

nomeata:
Thanks for your response!

> The ic0.canister_generation API is not strictly required since the build pipeline could include build information in a custom section e.g. ic:public build_serial 7 or ic:public buid_nanosec 1650466652000000 which would be available via read_state.

Hmm, that might identify _build_, but not _instances_, which is what I want to identify with the generation counter. Note that due to reinstalls, upgrades to the same code or not deploying some builds, the relation between the two is not strong (same generation number implies same build number, but that’s it).

> as a pointer to a WASM function (a function index) which doesn't work well over upgrades

Right, that’s what I’d fix separtely with the “names callback entry” proposal, which exists as a PR against the internal repo, and which I keep procrastinating to recreate against the pubic repo.

>  If they were an nonce unique to the call then a table in the callbacks could detect if the call was over an upgrade.

Interesting observation: If they were _system-provided_ unique numbers, this would also allow work (together with the change to make entry points named or otherwise stable).

> The existing 32 bit reply_env and reject_env are just a bit too small
> …
>  Better would be to extend the reply_env/reject_env to 64 bits

Ha, absolutely right! I was hoping to use that other PR that I keep mentioning but didn’t write yet to sneak in a change to make the env a `blob`, allowing the canister to store more data there. This would allow other interesting canister design where all relevant information about that call is stored there, and allows canisters to be otherwise stateless, which for some canisters may be enough to make them instantaneously upgradeable.

>  the canister can snap ic0.time() in canister_post_install to get equivalent functionality as well.

I thought about that, but it’s not perfect (what if a canister gets installed, called, and reinstalled in the same block?). I’d avoid abusing one concept (time) for something else (unique nonces). And in-canister-incrementing of time doesn’t work across reinstallations.

>  Alternatively, enabling a canister to set a trap reject message/data would enable passing back the nonce. 

Hmm, that seems like a no-go – if it requires the “other” canister to play by some rules, and it can mess with the nonce, then that creates a bunch of attack vectors I fear. This really needs system help.

> Having the system generate the nonce as unique per call is a bit problematic since query calls which make other query calls should be able to receive replies

I don’t see a problem there – the state(s) of a canister during queries handling is isolated from other canister states, essentially forking the timeline of the canister.


So if I may paraphrase some of your points:

 * Canisters can do more clever things if they get more than 32 bits of trustworthy storage in the call context.
 
   I agree, and would be happy to extend that to more. I’d just go for “arbitary” (and charge it to the canister of course), or maybe “message sized”. 

 * A system provided nonce generator would be useful.
 
   I agree; it could be used to recognize generations (obsoleting this PR), it could be used to label callbacks etc. `ic0.time` is close and looks tempting, but when we had that  discussion two years ago we decided against such “fake increments” in the system time, because it wouldn’t then be time anymore.

   But nothing stops us from introducing a `ic0.nonce` API that returns an strictly increasing 64 bit number that’s growing even across upgrades / reinstallions / code uninstall.

And I agree with your analysis: If we have the former, we don’t need `ic0.callback_generation`, and the latter subsumes `ic0.canister_generation`, while providing strictly more value.

(There is _one_ caveat: If we leave it to the canister to put the generation or nonce in the `env` data somehow, then it’s just a convention. So after reinstallation with a different canister, there is no guarantee that they use the same convention, and – theoretically – we can have a callback confuse the new canister. That is maybe the main benefit of a system-provided `ic0.callback_generation` – it is trustworthy and does depend on what the previous  canister module did.)

jplevyak:
OK, I agree with everything you said.

While I think that a new call API would be cleaner and a larger blob of reply_env/reject_env would be better, the canister generation is not a bad feature to have in any case although I do agree that it would be good to expose it in the state tree and it does minimally allow solving the issue of handling upgrades.

The only question is whether it would be better to add ic0.call_env_append() and ic0.msg_env_data_size/copy() instead of adding ic0.callback_generation() as the former could be used to store the canister generation.  Then if/when we change ic0.call_new to say ic0.send(callee_src, callee_size, name_src, name_size, callback_src, callback_size) we will not need to have env parameters.

Dfinity-Bjoern:
I am fine with doing the state tree part in a follow-up. When it comes to `ic0.call_env_append`, that in principle seems fine to me, but we should make sure that we're actually charging for the space of call contexts. Do you know whether we're doing that already @bogwar @dsarlis?

flyq:
ic.rocks used to provide similar functionality, but doesn't work now:
image

jplevyak:
We have more work to do on the cost model, but in the short term we could limit the amount of this data e.g. 1KB then relax that once we charge.

spec/index.adoc
```diff
@@ -950,6 +951,7 @@ ic0.msg_arg_data_size : () -> i32;                                          // I
 ic0.msg_arg_data_copy : (dst : i32, offset : i32, size : i32) -> ();        // I U Q Ry F
 ic0.msg_caller_size : () -> i32;                                            // I G U Q F
 ic0.msg_caller_copy : (dst : i32, offset: i32, size : i32) -> ();           // I G U Q F
+ic0.callback_generation : i64;                                              // Ry Rt
```
Dfinity-Bjoern, line 954:
```suggestion
ic0.callback_generation () -> : i64;                                         // Ry Rt
```
```suggestion
ic0.callback_generation : i64;                                              // Ry Rt
```

spec/index.adoc
```diff
@@ -2145,7 +2168,8 @@ CallOrigin
     }
   | FromCanister {
       calling_context : CallId;
-      callback: Callback
+      callback : Callback;
+      generation : Generation;
     }
   | FromHeartbeat
```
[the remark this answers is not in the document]
nomeata: Nice ticket number, very leet :-)

What is that about?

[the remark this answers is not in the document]
mraszyk: It is about querying more information about canisters, e.g., their history (changes to their module and controller set).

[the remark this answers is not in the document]
mraszyk: It is about querying more information about canisters, e.g., their history (changes to their module and controller set).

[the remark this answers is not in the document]
mraszyk: Actually, it suffices that the generation counter is part of the MethodCall.

mraszyk:
The canister version (counter) is merged in. Should we then close this PR? @nomeata

nomeata:
Sure, thanks

Issue #9, opened 2022-01-28
Push to netlify?


nomeata:
It would be very useful to be able to convenient browse the document at various revisions, including PRs. I believe that this is relatively simple to set up using netlify, including per-PR previews.

Should we do that?

Maybe we can use the self-contained single page style sheet that we used back then, which I find more usable with it's deep ToC on the left.

nomeata:
BTW, I just did
```
~/dfinity/interface-spec $ nix-build -A interface-spec
```
and it still works and can be used to create this rendering locally, which may be useful to some.

PR #20, opened 2022-04-01
Specify the HTTP Gateway protocol

nomeata:
We have had the HTTP gateway for a while now, but 

* we didn’t have a public record of the expected Canister interface, with the authorative Candid fragment
* we already have multiple implementations (`icx-proxy` and the service worker), and people we looking at the implementations to understand how thing work.
* we see new features being implemented in one, but not the other implementation, and they start to diverge
* some features are only implementation-defined (e.g. the certification fallback to `index.html`)
* the interaction of new features (e.g. certificate, upgrade to update and streaming) is ill or at least vaguely defined.

So I think we urgently need a proper specification for the HTTP Gateway protocol, and a place to host this. The IC Interface Spec is a natural place, so here is a possible start.

This PR is descriptive, not prescriptive, with regard to the status quo: I (try to) specify the existing behavior of our Gateway implementations, not define a new and different protocol.

[Preview available](https://ic-spec-pr-preview.netlify.app/#http-gateway), pushed manually until #9 or similar has been fixed.

### DIFF
--- a/spec/http-gateway.did
+++ b/spec/http-gateway.did
@@ -15,15 +15,22 @@ type HttpResponse = record {
   streaming_strategy: opt StreamingStrategy;
 };
 
+// Each canister that uses the streaming feature gets to choose their concrete
+// type; the HTTP Gateway will treat it as an opaque value that is only fed to
+// the callback method
+
+type StreamingToken = /* application-specific type */
+
+
 type StreamingCallbackHttpResponse = record {
   body: blob;
-  token: opt T;
+  token: opt StreamingToken;
 };
 
 type StreamingStrategy = variant {
   Callback: record {
-    callback: func (T) -> (opt StreamingCallbackHttpResponse) query;
-    token: T;
+    callback: func (StreamingToken) -> (opt StreamingCallbackHttpResponse) query;
+    token: StreamingToken;
   };
 };
 
--- a/spec/http-gateway.did
+++ b/spec/http-gateway.did
@@ -1,0 +1,34 @@
+type HeaderField = record { text; text; };
+
+type HttpRequest = record {
+  method: text;
+  url: text;
+  headers: vec HeaderField;
+  body: blob;
+};
+
+type HttpResponse = record {
+  status_code: nat16;
+  headers: vec HeaderField;
+  body: blob;
+  upgrade : opt bool;
+  streaming_strategy: opt StreamingStrategy;
+};
+
+type StreamingCallbackHttpResponse = record {
+  body: blob;
+  token: opt T;
+};
+
+type StreamingStrategy = variant {
+  Callback: record {
+    callback: func (T) -> (opt StreamingCallbackHttpResponse) query;
+    token: T;
+  };
+};
+
+service : {
+  http_request: (request: HttpRequest) -> (HttpResponse) query;
+  http_request_update: (request: HttpRequest) -> (HttpResponse);
+}
+
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
 lookup_path(["e"],      pruned_tree) = Absent
 ....
 
+[#http-gateway]
+== The HTTP Gateway protocol
+
+This section specifies the _HTTP Gateway protocol_, which allows canisters to handle conventional HTTP requests.
+
+This feature involves the help of a _HTTP Gateway_ that translates between HTTP requests and the IC protocol. Such a gateway could be a stand-alone proxy, it could be implemented in a web browsers (natively, via plugin or via a service worker) or in other ways. This document describes the interface and semantics of this protocol independent of a concrete Gateway, so that all Gateway implementations can be compatible.
+
+This protocol sits conceptually on top of the interface specified in the remainder of this document, as the involved parties (HTTP Gateway and canisters) are not part of the “system” described here. It is still included in this file due to its role in the Internet Computer ecosystem and due to the importance of keeping multiple Gateway implementations in sync.
+
+=== Overview
+
+A HTTP request by an HTTP client is handled by these steps:
+
+1. The Gateway resolves the Host of the request to a canister id.
+2. The Gateway Candid-encoded the HTTP request data.
+3. The Gateway invokes the canister via a query call to `http_request`.
+4. The canister handles the request and returns a HTTP response, encoded in Candid, together with additional metadata.
+5. If requested by the canister, the Gateway sends the request again via an update call to `http_request_update`.
+6. If applicable, the Gateway fetches futher body data via streaming query calls.
+7. If applicable, the Gateway validates the certificate of the response.
+8. The Gateway sends the response to the HTTP client.
+
+[#http-gateway-interface]
+=== Candid interface
+
+The following interface description, in https://github.com/dfinity/candid/blob/master/spec/Candid.md[Candid syntax], describes the expecte Canister interface. You can also link:{attachmentsdir}/http-gateway.did[download the file].
+----
+include::{example}http-gateway.did[]
+----
+
+This type signature is a type signature _scheme_, and the canister can choose any type for the token type `T` used for streaming. The Gateway handles that type generically.
+
+Only canisters that use the “Upgrade to update calls” feature need to provide the `http_request_update` method.
+
+NOTE: Canisters not using these upgrade these features can completeley leave out the `streaming_strategy` and/or `upgade` fields in the `HttpResponse` they return, due to how Candid subtyping works. This might simplifiy their code.
+
+[#http-gateway-name-resolution]
+=== Canister resolution
+
+The Gateway needs to know the canister id of the canister to talk to, and obtains that information from the hostname as follows:
+
+1. Check that the hostname, taken from the `Host` field of the HTTP request, is of the form `<name>.raw.ic0.app` or `<name>.ic0.app`, or fail.
+
+2. If the `<name>` is in the following table, use the given canister ids:
++
+.Canister hostname resolution
+|===
+|Hostname |Canister id
+|identity | rdmx6-jaaaa-aaaaa-aaadq-cai
+|nns | qoctq-giaaa-aaaaa-aaaea-cai
+|dscvr | h5aet-waaaa-aaaab-qaamq-cai
+|personhood | g3wsl-eqaaa-aaaan-aaaaa-cai
+
+3. Else, if `<name>` a valid textual encoding of a principal, use that principal as the canister ids.
+
+4. Else fail.
+
+If the hostname was of the form `<name>.ic0.app`, it is a _safe_ hostname; if it was of the form `<name>.raw.ic0.app` it is a _raw_ hostname.
+
+=== Request encoding
+
+The HTTP request is encoded into the `HttpRequest` candid structure.
+
+The `method` field contains the HTTP methd (e.g. `HTTP`), in upper case.
+
+The `url` field contains the URL from the HTTP request line, i.e. without protocol or hostname, and including query parameters.
+
+The `headers` field contains the headers of the HTTP request.
+
+The `body` field contains the body of the HTTP request (without any content encodings processed by the Gateway).
+
+=== Upgrade to update calls
+
+If the canister sets `update = opt true` in the `HttpResponse` reply from `http_request`, then the Gateway ignores all other fields of the reply. The Gateway performs an _update_ call, with the same argument, to `http_request_update`, and uses that response instead.
+
+The value of the `update` field returned from `http_request_update` is ignored.
+
+=== Response decoding and streaming
+
+The Gateway assembles the HTTP response from the given `HttpResponse` record.
+
+The HTTP response status code is taken from the `status_code` field.
+
+The HTTP response headers are taken from the `headers` field.
+
+The HTTP response body is initialized with the value of the `body` field.
+
+If the `streaming_strategy` field is set, the HTTP Gateway then uses further query calls to obtain further chunks to append to the body:
+
+1. If the function reference in the `callback` field of the `streaming_strategy` is not a method of the given canister, the Gateway fails the request.
+
+2. Else, it makes a query call to the give method, passing the `token` value given in the `streaming_strategy` as the argument.
+
+3. That query method returns a `StreamingCallbackHttpResponse`. The `body` therein is appended to the body of the HTTP response. This is repeated as long as the method returns some token in the `token` field, until that field is `null`.
+
+=== Response certification
+
+If the hostname was safe, the HTTP Gateway performs _certificate validation_:
+
+1. It searches for a response header called `Ic-Certificate` (case-insensitive).
+
+2. The value of the header must be a structured header according to RFC8941 with fields `certificate` and `tree`, both being byte sequences.
+
+3. The `certificate` must be a valid certificate as per <<certification>>, signed by the root key. If the certificate contains a subnet delegation, the delegation must be valid for the given canister. The timestamp in `/time` must be recent. The subnet state tree in the certificate must reveal the canister’s  <<state-tree-certified-data,certified data>>.
+
+4. The `tree` must be a hash trees as per <<certification-encoding>>.
+
+5. The root hash of that `tree` must match the canister’s certified data.
+
+6. The path `["http_assets",<url>]`, where `url` is the utf8-encoded `url` from the `HttpRequest` must exist and be a leaf.  Else, if it does not exist, `["http_assets","/index.html"]` must exist and be a leaf.
+
+7. That leaf must contain the SHA256 of the _decoded_ body.
++
+The decoded body is the body of the HTTP response (in particular, after assembling streaming chunks), decoded according to the `Content-Encoding` header, if present. Supported encodings for `Content-Encoding` are `gzip` and `deflate.`
+
+
 
 [#abstract-behavior]
 == Abstract behavior
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1956,7 +1956,7 @@ This section specifies the _HTTP Gateway protocol_, which allows canisters to ha
 
 This feature involves the help of a _HTTP Gateway_ that translates between HTTP requests and the IC protocol. Such a gateway could be a stand-alone proxy, it could be implemented in a web browsers (natively, via plugin or via a service worker) or in other ways. This document describes the interface and semantics of this protocol independent of a concrete Gateway, so that all Gateway implementations can be compatible.
 
-This protocol sits conceptually on top of the interface specified in the remainder of this document, as the involved parties (HTTP Gateway and canisters) are not part of the “system” described here. It is still included in this file due to its role in the Internet Computer ecosystem and due to the importance of keeping multiple Gateway implementations in sync.
+Conceptually, this protocol builds on top of the interface specifie d in the remainder of this document, and therefore is an “application-level” interface, not a feature of the core Internet Computer system described in the other sections, and could be a separate document. We nevertheless include this protocol in the Internet Computer Interface Specification because of its important role in the ecosystem and due to the importance of keeping multiple Gateway implementations in sync.
 
 === Overview
 
@@ -1967,7 +1967,7 @@ A HTTP request by an HTTP client is handled by these steps:
 3. The Gateway invokes the canister via a query call to `http_request`.
 4. The canister handles the request and returns a HTTP response, encoded in Candid, together with additional metadata.
 5. If requested by the canister, the Gateway sends the request again via an update call to `http_request_update`.
-6. If applicable, the Gateway fetches futher body data via streaming query calls.
+6. If applicable, the Gateway fetches further body data via streaming query calls.
 7. If applicable, the Gateway validates the certificate of the response.
 8. The Gateway sends the response to the HTTP client.
 
@@ -1983,7 +1983,7 @@ This type signature is a type signature _scheme_, and the canister can choose an
 
 Only canisters that use the “Upgrade to update calls” feature need to provide the `http_request_update` method.
 
-NOTE: Canisters not using these upgrade these features can completeley leave out the `streaming_strategy` and/or `upgade` fields in the `HttpResponse` they return, due to how Candid subtyping works. This might simplifiy their code.
+NOTE: Canisters not using these features can completely leave out the `streaming_strategy` and/or `upgade` fields in the `HttpResponse` they return, due to how Candid subtyping works. This might simplify their code.
 
 [#http-gateway-name-resolution]
 === Canister resolution
@@ -2022,21 +2022,27 @@ The `body` field contains the body of the HTTP request (without any content enco
 
 === Upgrade to update calls
 
-If the canister sets `update = opt true` in the `HttpResponse` reply from `http_request`, then the Gateway ignores all other fields of the reply. The Gateway performs an _update_ call, with the same argument, to `http_request_update`, and uses that response instead.
+If the canister sets `update = opt true` in the `HttpResponse` reply from `http_request`, then the Gateway ignores all other fields of the reply. The Gateway performs an _update_ call to `http_request_update`, passing the same `HttpRequest` record as the argument, and uses that response instead.
 
 The value of the `update` field returned from `http_request_update` is ignored.
 
+[#http-gateway-streaming]
 === Response decoding and streaming
 
-The Gateway assembles the HTTP response from the given `HttpResponse` record.
+The Gateway assembles the HTTP response from the given `HttpResponse` record:
 
-The HTTP response status code is taken from the `status_code` field.
+* The HTTP response status code is taken from the `status_code` field.
 
-The HTTP response headers are taken from the `headers` field.
+* The HTTP response headers are taken from the `headers` field.
 
-The HTTP response body is initialized with the value of the `body` field.
+* The HTTP response body is initialized with the value of the `body` field, and further assembled as per the <<http-gateway-streaming,streaming protocol>>.
 
-If the `streaming_strategy` field is set, the HTTP Gateway then uses further query calls to obtain further chunks to append to the body:
+[#http-gateway-streaming]
+=== Response decoding and streaming
+
+The HTTP Gateway protocol has provisions transfer further chunks of the body data from the canister to the HTTP Gateway, to overcome the message limit of the Internet Computer. This streaming protocol is idependent of any possible streaming of data between the HTTP Gateway and the HTTP client. The gateway may assemble the response in whole before passing it on, or pass the chunks on directly, on the TCP or HTTP level, as it sees fit. When the Gateway is <<http-gateway-certification,certifying the response>>, it must not pass on uncertified chunks.
+
+If the `streaming_strategy` field of the `HttpResponse` is set, the HTTP Gateway then uses further query calls to obtain further chunks to append to the body:
 
 1. If the function reference in the `callback` field of the `streaming_strategy` is not a method of the given canister, the Gateway fails the request.
 
@@ -2044,6 +2050,7 @@ If the `streaming_strategy` field is set, the HTTP Gateway then uses further que
 
 3. That query method returns a `StreamingCallbackHttpResponse`. The `body` therein is appended to the body of the HTTP response. This is repeated as long as the method returns some token in the `token` field, until that field is `null`.
 
+[#http-gateway-certification]
 === Response certification
 
 If the hostname was safe, the HTTP Gateway performs _certificate validation_:
@@ -2064,6 +2071,7 @@ If the hostname was safe, the HTTP Gateway performs _certificate validation_:
 +
 The decoded body is the body of the HTTP response (in particular, after assembling streaming chunks), decoded according to the `Content-Encoding` header, if present. Supported encodings for `Content-Encoding` are `gzip` and `deflate.`
 
+WARNING: The certification protocol only covers the mapping from request URL to response body. It completely ignores the request method and headers, and does not cover the response headers and status code.
 
 
 [#abstract-behavior]
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2026,8 +2026,7 @@ If the canister sets `update = opt true` in the `HttpResponse` reply from `http_
 
 The value of the `update` field returned from `http_request_update` is ignored.
 
-[#http-gateway-streaming]
-=== Response decoding and streaming
+=== Response decoding
 
 The Gateway assembles the HTTP response from the given `HttpResponse` record:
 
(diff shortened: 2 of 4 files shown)


nomeata:
Draft until external contributions are possible.

@paulyoung, can you also review this? It would be helpful if you could double check this. Who else should be pinged?

nomeata:
@jplevyak, right now the spec says that all headers are passed through by the boundary node, but that’s not really true, is it? What header filtering/mangling/addition should we specify/document here?

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+=== Overview
+
+A HTTP request by an HTTP client is handled by these steps:
+
+1. The Gateway resolves the Host of the request to a canister id.
+2. The Gateway Candid-encoded the HTTP request data.
+3. The Gateway invokes the canister via a query call to `http_request`.
+4. The canister handles the request and returns a HTTP response, encoded in Candid, together with additional metadata.
+5. If requested by the canister, the Gateway sends the request again via an update call to `http_request_update`.
+6. If applicable, the Gateway fetches futher body data via streaming query calls.
```
jzxchiang1, line 1970: Typo: `futher` -> `further`

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+The following interface description, in https://github.com/dfinity/candid/blob/master/spec/Candid.md[Candid syntax], describes the expecte Canister interface. You can also link:{attachmentsdir}/http-gateway.did[download the file].
+----
+include::{example}http-gateway.did[]
+----
+
+This type signature is a type signature _scheme_, and the canister can choose any type for the token type `T` used for streaming. The Gateway handles that type generically.
+
+Only canisters that use the “Upgrade to update calls” feature need to provide the `http_request_update` method.
+
+NOTE: Canisters not using these upgrade these features can completeley leave out the `streaming_strategy` and/or `upgade` fields in the `HttpResponse` they return, due to how Candid subtyping works. This might simplifiy their code.
```
jzxchiang1, line 1986: Typo: `Canisters not using these features can completely`

`simplifiy` -> `simplify`

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+2. If the `<name>` is in the following table, use the given canister ids:
++
+.Canister hostname resolution
+|===
+|Hostname |Canister id
+|identity | rdmx6-jaaaa-aaaaa-aaadq-cai
+|nns | qoctq-giaaa-aaaaa-aaaea-cai
+|dscvr | h5aet-waaaa-aaaab-qaamq-cai
+|personhood | g3wsl-eqaaa-aaaan-aaaaa-cai
```
jzxchiang1, line 2003: Where is `personhood` specified?

I don't see it [here](https://github.com/dfinity/agent-js/blob/6e8c64cf07c7722aafbf52351eb0f19fcb954ff0/apps/sw-cert/src/sw/http_request.ts#L15).

Not really convinced this canister resolution belongs in the HTTP Gateway protocol, although it is good to know in general I suppose.

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+If the hostname was of the form `<name>.ic0.app`, it is a _safe_ hostname; if it was of the form `<name>.raw.ic0.app` it is a _raw_ hostname.
+
+=== Request encoding
+
+The HTTP request is encoded into the `HttpRequest` candid structure.
+
+The `method` field contains the HTTP methd (e.g. `HTTP`), in upper case.
+
+The `url` field contains the URL from the HTTP request line, i.e. without protocol or hostname, and including query parameters.
```
jzxchiang1, line 2017: Very important! The query parameters inclusion tripped me up.

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+The `headers` field contains the headers of the HTTP request.
+
+The `body` field contains the body of the HTTP request (without any content encodings processed by the Gateway).
+
+=== Upgrade to update calls
+
+If the canister sets `update = opt true` in the `HttpResponse` reply from `http_request`, then the Gateway ignores all other fields of the reply. The Gateway performs an _update_ call, with the same argument, to `http_request_update`, and uses that response instead.
```
jzxchiang1, line 2025: Probably clearer to say "with the same `HttpRequest`" instead of "with the same argument" IMO

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
 ....
 
+[#http-gateway]
+== The HTTP Gateway protocol
+
+This section specifies the _HTTP Gateway protocol_, which allows canisters to handle conventional HTTP requests.
+
+This feature involves the help of a _HTTP Gateway_ that translates between HTTP requests and the IC protocol. Such a gateway could be a stand-alone proxy, it could be implemented in a web browsers (natively, via plugin or via a service worker) or in other ways. This document describes the interface and semantics of this protocol independent of a concrete Gateway, so that all Gateway implementations can be compatible.
+
+This protocol sits conceptually on top of the interface specified in the remainder of this document, as the involved parties (HTTP Gateway and canisters) are not part of the “system” described here. It is still included in this file due to its role in the Internet Computer ecosystem and due to the importance of keeping multiple Gateway implementations in sync.
```
jzxchiang1, line 1959: This paragraph confused me. I'm not exactly sure what the "system" is. Also, the "it" in "It is still included in this file" didn't make sense to me.

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+The HTTP response body is initialized with the value of the `body` field.
+
+If the `streaming_strategy` field is set, the HTTP Gateway then uses further query calls to obtain further chunks to append to the body:
+
+1. If the function reference in the `callback` field of the `streaming_strategy` is not a method of the given canister, the Gateway fails the request.
+
+2. Else, it makes a query call to the give method, passing the `token` value given in the `streaming_strategy` as the argument.
+
+3. That query method returns a `StreamingCallbackHttpResponse`. The `body` therein is appended to the body of the HTTP response. This is repeated as long as the method returns some token in the `token` field, until that field is `null`.
```
jzxchiang1, line 2045: Slightly off-topic, but I've always been confused exactly how this streaming actually gets implemented. Is it HTTP/1.1 chunked transfer encoding or HTTP/2 streaming?

Should the protocol described in this document talk about whether HTTP/1.1 or HTTP/2 (or both) are supported?

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+=== Request encoding
+
+The HTTP request is encoded into the `HttpRequest` candid structure.
+
+The `method` field contains the HTTP methd (e.g. `HTTP`), in upper case.
+
+The `url` field contains the URL from the HTTP request line, i.e. without protocol or hostname, and including query parameters.
+
+The `headers` field contains the headers of the HTTP request.
```
jzxchiang1, line 2019: I wonder if the protocol should specify which request headers are special and will be automatically processed by any Gateway implementation.

I suppose `Host` is an example of one, but how about `Cache-Control` or any others? icx-proxy doesn't do any caching AFAIK (although boundary nodes, which run icx-proxy, apparently do). Should these kind of features be specified in the protocol?

spec/http-gateway.did
```diff
@@ -0,0 +1,34 @@
...
+  url: text;
+  headers: vec HeaderField;
+  body: blob;
+};
+
+type HttpResponse = record {
+  status_code: nat16;
+  headers: vec HeaderField;
+  body: blob;
+  upgrade : opt bool;
```
jzxchiang1, line 14: Typo: `upgrade` -> `update`

spec/http-gateway.did
```diff
@@ -0,0 +1,34 @@
...
+};
+
+type StreamingCallbackHttpResponse = record {
+  body: blob;
+  token: opt T;
+};
+
+type StreamingStrategy = variant {
+  Callback: record {
+    callback: func (T) -> (opt StreamingCallbackHttpResponse) query;
```
jzxchiang1, line 25: Unfortunately, there's a Candid bug which in practice will cause a runtime error if you define the callback like this.

See this [comment](https://forum.dfinity.org/t/cryptic-error-from-icx-proxy/6944/9?u=jzxchiang) by @chenyan-dfinity for details.

Perhaps the protocol can remain "pristine" but if users want code that actually works, they will need to change the type to:

`callback: func (?T) -> (opt StreamingCallbackHttpResponse);`

spec/index.adoc
```diff
@@ -1949,6 +1949,122 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+If the `streaming_strategy` field is set, the HTTP Gateway then uses further query calls to obtain further chunks to append to the body:
+
+1. If the function reference in the `callback` field of the `streaming_strategy` is not a method of the given canister, the Gateway fails the request.
+
+2. Else, it makes a query call to the give method, passing the `token` value given in the `streaming_strategy` as the argument.
+
+3. That query method returns a `StreamingCallbackHttpResponse`. The `body` therein is appended to the body of the HTTP response. This is repeated as long as the method returns some token in the `token` field, until that field is `null`.
+
+=== Response certification
```
jzxchiang1, line 2047: I don't feel qualified to review the certification part at the moment, so I'll defer this for someone else.

spec/http-gateway.did
```diff
@@ -0,0 +1,34 @@
...
+    token: T;
+  };
+};
+
+service : {
+  http_request: (request: HttpRequest) -> (HttpResponse) query;
+  http_request_update: (request: HttpRequest) -> (HttpResponse);
+}
```
jzxchiang1, line 33: I think it'd be clearer to make reference to the streaming callback here, but I understand why you didn't. (It's optional and also dynamically specified in `HttpResponse`.)

jzxchiang1 (changes requested):

jplevyak:
@frederikrothenberger and @Daniel-Bloom-dfinity are current working this.

spec/index.adoc
nomeata: I saw it for the first time yesterday in https://github.com/dfinity/ic/blob/a254116be472f81b85b76429ac8d987f80dd4411/typescript/service-worker/src/sw/http_request.ts#L17. I'll leave it to the Dfinity people to say if this should be there or not. But it demonstrates the need for a authorative list, at least until we have a dynamic system in place.

spec/index.adoc
nomeata: Yes, definitely! Unfortunately I don't know the precise state of affairs here, and hope that @jplevyak can fill in the blanks.

Although maybe some things need not be documented. If the boundary nodes acts on certain headers in a HTTP spec compliant way, e.g. while caching, then that should be fine even if it isn't written here. just as if some user uses their ISP's proxy and that caches.

spec/http-gateway.did
nomeata: Is that just due to bugs in some implementation that will be fixed in due time? Then I'd rather keep it as designed here, and leave temporary work-arounds out.

spec/http-gateway.did
nomeata: The name is arbitrary, right? So I wouldn't quite know how to include it.

spec/index.adoc
nomeata: This streaming is _only_ concerned with getting the data from the canister to the Gateway, to overcome the message size limit, so it's independent of any HTTP-level streaming. Note that with certification (as it is specified now), the Gateway can't even stream out the data until it has received it all. In the uncertified case it could stream it out (on the TCP level, or the HTTP level) - I don't care and leave that as as implementation choice. Note that if the Gateway is the browser itself it also doesn't make sense to stream.

But it's a good point and I'll clarify!

spec/index.adoc
nomeata: I rewrote it. Better? Or is this not a helpful paragraph?

spec/http-gateway.did
nomeata: Unfortunately(?) this field _is_ called `upgrade`, as in “upgrade the call from query to an update call”. Very confusing… oh well.

See <https://github.com/dfinity/icx-proxy/pull/6/files#diff-42cb6807ad74b3e201c5a7ca98b911c5fa08380e942be6e4ac5807f8377f87fcR259>

spec/index.adoc
```diff
@@ -1949,6 +1949,129 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+The value of the `update` field returned from `http_request_update` is ignored.
+
+=== Response decoding
+
+The Gateway assembles the HTTP response from the given `HttpResponse` record:
+
+* The HTTP response status code is taken from the `status_code` field.
+
+* The HTTP response headers are taken from the `headers` field.
```
nomeata, line 2035: Is the boundary node mangling or filtering the headers in any way that should be noted here?

origyndev:
If we are making suggestions here I’d love for there to be a “return and upgrade” feature that returns the result of the query to the client but then calls update for recording the serving of the request in the canister. The reply to this call can be discarded as the request has already been served.

nomeata:
> If we are making suggestions here

I suggest we don't, for now. Let's focus in finalizing a spec for the status quo. Might be hard enough.

Once thats done additional changes can be nicely discussed in their own PRs against master. Or maybe on the forum first.

spec/http-gateway.did
chenyan-dfinity: It's only an implementation problem. There is a more recent workaround here by abusing serde's `ignored_any`: https://github.com/dfinity/agent-rs/blob/main/ic-utils/src/interfaces/http_request.rs#L231

I think the real problem here is that `T` is undefined, it's not a proper Candid file.

nomeata: Yeah, but that's too late to avoid now, right?

chenyan-dfinity: I guess the solution is to catch up on the Candid side to add generics? In the Rust library, we kind of overcome this by deserializing the value into an untyped `IDLValue`.

In the meantime, we probably need to add a comment here to not confuse the reader.

nomeata: I wrote this in the adoc:

>  This type signature is a type signature _scheme_, and the canister can choose any type for the token type `T` used for streaming. The Gateway handles that type generically.

But yes, maybe better moved into the Candid file proper?

spec/index.adoc
frederikrothenberger: Imho, the gateway should interfere with the HTTP interaction as little as possible. There is one issue though, which is that the service worker is unable to pass on `Set-Cookie` headers, as they are `forbidden`: https://fetch.spec.whatwg.org/#forbidden-response-header-name

nomeata: This only affects service workers, but not Gateways implemented as proxies or IC-aware browsers, right?

In that case I would not forbid them here, but maybe leave a NOTE that not all Gateway implementations may be able to pass these headers on?

spec/index.adoc
Daniel-Bloom-dfinity: The filtering we have is in [dfinity/ic/ic-os/boundary-guestos/rootfs/etc/nginx/conf.d/001-ic-nginx.conf](https://github.com/dfinity/ic/blob/master/ic-os/boundary-guestos/rootfs/etc/nginx/conf.d/001-ic-nginx.conf#L154).

The short of it is we add these response headers (and thus their implications) to every HTTP call response:

```access-control-allow-credentials: true
access-control-allow-origin: *
access-control-allow-methods: GET, POST, HEAD, OPTIONS
access-control-allow-headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range,Cookie
access-control-expose-headers: Content-Length,Content-Range
x-cache-status: MISS
```

spec/http-gateway.did
nomeata: Hmm, `reserved` isn’t quite right (that is a specific Candid type) But maybe
```
type Token = /* insert service-specific type here */
```
or something like that

spec/index.adoc
nomeata: Done. Feel free to suggest edits to the wording if it should be further refined.

spec/http-gateway.did
nomeata: Refined, and added a stern warning

spec/index.adoc
nomeata: I guess the question is: Why set these? Shouldn’t the canister have control over them?

nomeata:
I’ve manually pushed a [rendering of this to `netlify`](https://ic-spec-pr-preview.netlify.app/#http-gateway), to get a preview until #9 or similar has been fixed.

spec/http-gateway.did
crusso: Probably to late to change, but would `promote` be a better name?

paulyoung:
I'll try to review this soon but in the meantime I wanted to share https://forum.dfinity.org/t/boundary-node-http-response-headers/10747 in case its of interest or provides some additional insight.

> Who else should be pinged?

Based on the above I think @lastmjs would be interested in this.

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+4. The canister handles the request and returns a HTTP response, encoded in Candid, together with additional metadata.
+5. If requested by the canister, the Gateway sends the request again via an update call to `http_request_update`.
+6. If applicable, the Gateway fetches further body data via streaming query calls.
+7. If applicable, the Gateway validates the certificate of the response.
+8. The Gateway sends the response to the HTTP client.
+
+[#http-gateway-interface]
+=== Candid interface
+
+The following interface description, in https://github.com/dfinity/candid/blob/master/spec/Candid.md[Candid syntax], describes the expecte Canister interface. You can also link:{attachmentsdir}/http-gateway.did[download the file].
```
paulyoung, line 1977:
```suggestion
The following interface description, in https://github.com/dfinity/candid/blob/master/spec/Candid.md[Candid syntax], describes the expected Canister interface. You can also link:{attachmentsdir}/http-gateway.did[download the file].
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+----
+include::{example}http-gateway.did[]
+----
+
+Only canisters that use the “Upgrade to update calls” feature need to provide the `http_request_update` method.
+
+NOTE: Canisters not using these features can completely leave out the `streaming_strategy` and/or `upgade` fields in the `HttpResponse` they return, due to how Candid subtyping works. This might simplify their code.
```
paulyoung, line 1984:
```suggestion
NOTE: Canisters not using these features can completely leave out the `streaming_strategy` and/or `upgrade` fields in the `HttpResponse` they return, due to how Candid subtyping works. This might simplify their code.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+.Canister hostname resolution
+|============================================
+| Hostname     | Canister id
+| `identity`   | `rdmx6-jaaaa-aaaaa-aaadq-cai`
+| `nns`        | `qoctq-giaaa-aaaaa-aaaea-cai`
+| `dscvr`      | `h5aet-waaaa-aaaab-qaamq-cai`
+| `personhood` | `g3wsl-eqaaa-aaaan-aaaaa-cai`
+|============================================
+
+3. Else, if `<name>` a valid textual encoding of a principal, use that principal as the canister ids.
```
paulyoung, line 2004:
```suggestion
3. Else, if `<name>` is a valid textual encoding of a principal, use that principal as the canister id.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+4. Else fail.
+
+If the hostname was of the form `<name>.ic0.app`, it is a _safe_ hostname; if it was of the form `<name>.raw.ic0.app` it is a _raw_ hostname.
+
+=== Request encoding
+
+The HTTP request is encoded into the `HttpRequest` candid structure.
+
+* The `method` field contains the HTTP methd (e.g. `HTTP`), in upper case.
```
paulyoung, line 2014:
```suggestion
* The `method` field contains the HTTP method (e.g. `HTTP`), in upper case.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+=== Response decoding
+
+The Gateway assembles the HTTP response from the given `HttpResponse` record:
+
+* The HTTP response status code is taken from the `status_code` field.
+
+* The HTTP response headers are taken from the `headers` field.
++
+NOTE: Not all Gateway implementations may be able to pass on all forms of headers. In particular, Service Workers are unable to pass on the `Set-cookie` header.
```
paulyoung, line 2036:
```suggestion
NOTE: Not all Gateway implementations may be able to pass on all forms of headers. In particular, Service Workers are unable to pass on the `Set-Cookie` header.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
++
+[NOTE]
+====
+HTTP Gateway may add addditional headers. In particular, the following headers may be set:
```
paulyoung, line 2040:
```suggestion
HTTP Gateways may add additional headers. In particular, the following headers may be set:
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+x-cache-status: MISS
+....
+====
+
+* The HTTP response body is initialized with the value of the `body` field, and further assembled as per the <<http-gateway-streaming,streaming protocol>>.
+
+[#http-gateway-streaming]
+=== Response body streaming
+
+The HTTP Gateway protocol has provisions transfer further chunks of the body data from the canister to the HTTP Gateway, to overcome the message limit of the Internet Computer. This streaming protocol is idependent of any possible streaming of data between the HTTP Gateway and the HTTP client. The gateway may assemble the response in whole before passing it on, or pass the chunks on directly, on the TCP or HTTP level, as it sees fit. When the Gateway is <<http-gateway-certification,certifying the response>>, it must not pass on uncertified chunks.
```
paulyoung, line 2056:
```suggestion
The HTTP Gateway protocol has provisions to transfer further chunks of the body data from the canister to the HTTP Gateway, to overcome the message limit of the Internet Computer. This streaming protocol is independent of any possible streaming of data between the HTTP Gateway and the HTTP client. The gateway may assemble the response in whole before passing it on, or pass the chunks on directly, on the TCP or HTTP level, as it sees fit. When the Gateway is <<http-gateway-certification,certifying the response>>, it must not pass on uncertified chunks.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+If the `streaming_strategy` field of the `HttpResponse` is set, the HTTP Gateway then uses further query calls to obtain further chunks to append to the body:
+
+1. If the function reference in the `callback` field of the `streaming_strategy` is not a method of the given canister, the Gateway fails the request.
+
+2. Else, it makes a query call to the give method, passing the `token` value given in the `streaming_strategy` as the argument.
```
paulyoung, line 2062:
```suggestion
2. Else, it makes a query call to the given method, passing the `token` value given in the `streaming_strategy` as the argument.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+3. That query method returns a `StreamingCallbackHttpResponse`. The `body` therein is appended to the body of the HTTP response. This is repeated as long as the method returns some token in the `token` field, until that field is `null`.
+
+WARNING: The type of the `token` value is chosen by the canister; the HTTP Gateway obtains the Candid type from the encoded message from the canister, and use it when passing the token back to the canister. This generic use of Candid is no covered by the Candid specification, and may not be possible in some cases (e.g. when using “future types”). Canister authors may have to use “simple” types.
```
paulyoung, line 2066:
```suggestion
WARNING: The type of the `token` value is chosen by the canister; the HTTP Gateway obtains the Candid type of the encoded message from the canister, and uses it when passing the token back to the canister. This generic use of Candid is not covered by the Candid specification, and may not be possible in some cases (e.g. when using “future types”). Canister authors may have to use “simple” types.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+3. Else, if `<name>` a valid textual encoding of a principal, use that principal as the canister ids.
+
+4. Else fail.
+
+If the hostname was of the form `<name>.ic0.app`, it is a _safe_ hostname; if it was of the form `<name>.raw.ic0.app` it is a _raw_ hostname.
+
+=== Request encoding
+
+The HTTP request is encoded into the `HttpRequest` candid structure.
```
paulyoung, line 2012:
```suggestion
The HTTP request is encoded into the `HttpRequest` Candid structure.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+
+[#http-gateway-certification]
+=== Response certification
+
+If the hostname was safe, the HTTP Gateway performs _certificate validation_:
+
+1. It searches for a response header called `Ic-Certificate` (case-insensitive).
+
+2. The value of the header must be a structured header according to RFC8941 with fields `certificate` and `tree`, both being byte sequences.
```
paulyoung, line 2076:
```suggestion
2. The value of the header must be a structured header according to RFC 8941 with fields `certificate` and `tree`, both being byte sequences.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+3. The `certificate` must be a valid certificate as per <<certification>>, signed by the root key. If the certificate contains a subnet delegation, the delegation must be valid for the given canister. The timestamp in `/time` must be recent. The subnet state tree in the certificate must reveal the canister’s  <<state-tree-certified-data,certified data>>.
```
paulyoung, line 2078:
```suggestion
3. The `certificate` must be a valid certificate as per <<certification>>, signed by the root key. If the certificate contains a subnet delegation, the delegation must be valid for the given canister. The timestamp in `/time` must be recent. The subnet state tree in the certificate must reveal the canister’s <<state-tree-certified-data,certified data>>.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+4. The `tree` must be a hash trees as per <<certification-encoding>>.
```
paulyoung, line 2080:
```suggestion
4. The `tree` must be a hash tree as per <<certification-encoding>>.
```

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+
+5. The root hash of that `tree` must match the canister’s certified data.
+
+6. The path `["http_assets",<url>]`, where `url` is the utf8-encoded `url` from the `HttpRequest` must exist and be a leaf.  Else, if it does not exist, `["http_assets","/index.html"]` must exist and be a leaf.
+
+7. That leaf must contain the SHA256 of the _decoded_ body.
```
paulyoung, line 2086:
```suggestion
7. That leaf must contain the SHA-256 hash of the _decoded_ body.
```

spec/index.adoc
Daniel-Bloom-dfinity: The short answer is "it's complicated". The longer answer is "because headers are not yet certified". As such we currently limit headers a bit to improve security. Once we get certified headers, we should not have any more need for this limitation.

spec/http-gateway.did
nomeata: Probably. Or `resend_as_update_call`. Or, in line with the pattern for the streaming callback, it could even pass a `func` ref.

In a way it's not too late: we can change this while keeping backward conpat in the implementations (there is only one so far) relatively easily (the proxy can simply keep both in it's internal expected type). I just wonder if we should do it with this PR, or maybe as a separate step, to get this in first.

spec/index.adoc
nomeata: Ok, good enough for this PR I guess.

paulyoung:
Pinging @SuddenlyHazel for visibility since this set of functionality came up in conversations today.

nomeata:
Ok, with the CLA requirement out of the way, this _could_ now be merged. Who on the DFINITY side is owning the process of shepherding and eventually merging community contributions? @Dfinity-Bjoern?

jwiegley:
@nomeata It is Björn and myself.

ali-dfinity:
update cla

spec/index.adoc
```diff
@@ -1949,6 +1949,146 @@ lookup_path(["d"],      pruned_tree) = Found "morning"
...
+[#http-gateway-certification]
+=== Response certification
+
+If the hostname was safe, the HTTP Gateway performs _certificate validation_:
+
+1. It searches for a response header called `Ic-Certificate` (case-insensitive).
+
+2. The value of the header must be a structured header according to RFC 8941 with fields `certificate` and `tree`, both being byte sequences.
+
+3. The `certificate` must be a valid certificate as per <<certification>>, signed by the root key. If the certificate contains a subnet delegation, the delegation must be valid for the given canister. The timestamp in `/time` must be recent. The subnet state tree in the certificate must reveal the canister’s <<state-tree-certified-data,certified data>>.
```
frederikrothenberger, line 2078: Currently neither icx-proxy nor the service worker check the certificate `/time` (which is bad). Can you give some more specific measure of "recent"? 5 minutes?

nomeata: That’s probably for the security team to decide. 5 min should be ample (until we get into edge node caching, then things become tricky, because a cache is hardly different from an attacker serving old data).

frederikrothenberger (approved):
Thanks for writing this spec.

Dfinity-Bjoern (approved):
That's great! Thank you very much @nomeata for contributing this!

nomeata:
Note that I am not allowed to merge, so someone else will have to press that button (and ideally presses the “Squash and merge” button and manually cleans up the PR description or copies the PR description into the commit message – mergify would do that for us but that’s not enabled at the moment.)

Daniel-Bloom-dfinity (approved):

PR #21, opened 2022-04-05
Editorial edit: Pass `Env` to `inspect_message`

nomeata:
thanks to @jwiegley for pointing it out

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2644,7 +2644,7 @@ Conditions::
     ) ∨ (
       E.content.canister_id ≠ ic_principal
       S.canisters[E.content.canister_id] ≠ EmptyCanister
-      Arg = { data = E.content.arg; caller = E.content.sender; method = E.content.method_name; time = S.time[E.content.canister_id] }
+      Arg = { data = E.content.arg; caller = E.content.sender }
       Env = {
         time = S.time[E.content.canister_id];
         balance = S.balances[E.content.canister_id]
@@ -3881,7 +3881,7 @@ Params = {
   reject_message : Text;
   sysenv : Env;
   cycles_refunded : Nat;
-  method_name : Text;
+  method_name : NoData | Text;
 }
 ExecutionState = {
   wasm_state : WasmState;
@@ -3918,6 +3918,7 @@ empty_params = {
   reject_code = 0;
   reject_message = "";
   cycles_refunded = 0;
+  method_name = NoData;
 }
 
 empty_execution_state = {
@@ -4147,7 +4148,7 @@ inspect_message = λ (method_name, wasm_state, arg, sysenv) →
       params = empty_params with {
         data = arg.data;
         caller = arg.caller;
-        method_name = arg.method_name;
+        method_name = method_name;
         sysenv
       }
       balance = sysenv.balance;
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -3574,7 +3574,7 @@ S with
       Younger_messages
 ....
 
-If the responded call context does not exist anymore, because the canister has been uninstalled since, the refundend cycles are still added to the canister balance, but
+If the responded call context does not exist anymore, because the canister has been uninstalled since, the refunded cycles are still added to the canister balance, but
 no function invocation is enqueued:
 
 Conditions::
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2267,10 +2267,8 @@ WasmState = (abstract)
 StableMemory = (abstract)
 Callback = (abstract)
 
-Arg = {
-  data : Blob
-  caller: Principal
-}
+Arg = Blob;
+CallerId = Principal;
 
 Timestamp = Nat;
 Env = {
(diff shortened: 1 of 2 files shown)


jwiegley:
And thanks to Coq for pointing this out to me. :-) The first impact of the formalization effort!

spec/index.adoc
```diff
@@ -2406,9 +2406,16 @@ Conditions::
...
       E.content.canister_id ≠ ic_principal
       S.canisters[E.content.canister_id] ≠ EmptyCanister
-      Arg = { data = E.content.arg; caller = E.content.sender; method = E.content.method_name; time = S.time[E.content.canister_id] }
+      Arg = { data = E.content.arg; caller = E.content.sender }
+      Env = {
+        time = S.time[E.content.canister_id];
+        balance = S.balances[E.content.canister_id]
+        freezing_limit = freezing_limit(S, E.content.canister_id);
+        certificate = NoCertificate;
+        status = S.status[E.content.canister_id];
```
jwiegley, line 2415: I believe you mean `canister_status` here?

nomeata: Yes, probably. I copied this from elsewhere, so it was wrong there too. I'll refine.

spec/index.adoc
```diff
@@ -3614,7 +3621,7 @@ Params = {
   reject_message : Text;
   sysenv : Env;
   cycles_refundend : Nat;
```
jwiegley, line 3623: You mean `cycles_refunded`.

spec/index.adoc
```diff
@@ -3614,7 +3621,7 @@ Params = {
...
-  method_name : Text;
+  method_name : NoData | Text;
```
jwiegley, line 3624: I wonder why these aren't `Optional Text` or `Text?`?

jwiegley (commented):
I've made a few notes throughout...

spec/index.adoc
jwiegley: Also, the function `freezing_limit` is not defined anywhere in the master version of `index.adoc`.

jwiegley: @nomeata No, it can't be the `CanStatus`, because `Env` specifies:
```status : Running | Stopping | Stopped```
Is this perhaps a reference to the `CanStatus` that is later defined?
```
CanStatus
  = Running
  | Stopping (List (CallOrigin, Nat))
  | Stopped
```
Or are they separate types that get lined up in the state transition?

File: spec/index.adoc (hunk 3 above)
Comment (nomeata) [reply spec/index.adoc:3665]: That's roughly what it means, but without the extra constructor in the non-null case to deal with everywhere. In a mechanization, I'd use an option

**Comment by nomeata** [comment]
Sorry, I didn't see the reviews. And there are conflicts now, I'll have to check if this PR is even still needed.

File: spec/index.adoc (hunk 1 above)
Comment (nomeata) [reply spec/index.adoc:2581]: Ah, yes, that’s a bit annoying. In `S` we want to keep track of the pending call contexts when the state is `Stopping`, but in `Env` (and also in the Canister status) it is just an enum. Right now we are type-wise sloppy here.

File: spec/index.adoc [hunk 4]
```diff
@@ -3731,8 +3738,8 @@ Params = {
   reject_code : 0 | SYS_FATAL | SYS_TRANSIENT | …;
   reject_message : Text;
   sysenv : Env;
-  cycles_refunded : Nat;
-  method_name : Text;
+  cycles_refundend : Nat;
```
Dfinity-Bjoern, line 3741:
```suggestion
  cycles_refunded : Nat;
```

spec/index.adoc
mraszyk: Let's introduce a type conversion function between `CanStatus` and `status : Running | Stopping | Stopped` (it's called `to_status` in the formal model).

spec/index.adoc
```diff
@@ -2509,9 +2509,16 @@ Conditions::
     ) ∨ (
       E.content.canister_id ≠ ic_principal
       S.canisters[E.content.canister_id] ≠ EmptyCanister
-      Arg = { data = E.content.arg; caller = E.content.sender; method = E.content.method_name; time = S.time[E.content.canister_id] }
+      Arg = { data = E.content.arg; caller = E.content.sender }
```
mraszyk, line 2512: Why is `caller` still part of `Arg`?

nomeata: I don’t remember :-). Would you mind taking over? I think you are currently deeper into this than I am

mraszyk: I'd be happy to take over. :) Thank you for your trust!

spec/index.adoc
mraszyk: I've introduced the conversion function and named it `simple_status` (both in textual and formal spec).

spec/index.adoc
mraszyk: I've turned `Arg` to `Blob` and have `CallerId = Principal` passed separately.

Dfinity-Bjoern (approved):
Thanks @mraszyk for shepherding!

PR #22, opened 2022-04-06
IC00 HTTP request - Return error via Reject

AlexandraZapuc:
This MR updates the candid return type for the `http_request` management method. A `Reply` will contain the `http_response`, while any error encountered in IC will be reflected via the `Reject` (reject_code, reject_message).

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -16,23 +16,12 @@ type definite_canister_settings = record {
   freezing_threshold : nat;
 };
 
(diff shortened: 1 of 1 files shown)


rumenov:
Thanks for doing this change Alexandra!

rumenov (approved):

Dfinity-Bjoern (approved):
Thanks for catching that!

PR #23, opened 2022-04-08
Update README.md

levsha:


nomeata:
Thanks, but that is just an example version number, not sure if it’s worth changing.

nomeata:
> Thanks, but that is just an example version number, not sure if it’s worth changing.

Ah, these are test PRs. Silly me :-)

levsha:
> > Thanks, but that is just an example version number, not sure if it’s worth changing.
> 
> Ah, these are test PRs. Silly me :-)

Yes. Sorry for confusion.

PR #30, opened 2022-04-09
Clean up changelog (Editorial)

nomeata:
now that this repo no longer contains what is in
https://github.com/dfinity/ic-hs it seems prudent to remove the non-spec
entries from the changelog. Will also make more sense on
https://smartcontracts.org/docs/interface-spec/index.html#changelog

### DIFF
--- a/spec/changelog.adoc
+++ b/spec/changelog.adoc
@@ -5,8 +5,11 @@
 === ∞ (unreleased)
 
 * Canister access to performance metrics
-* Expose Wasm custom sections in the state tree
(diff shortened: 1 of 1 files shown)


nomeata:
I also added two missing entries.

nomeata:
NB: I can't merge myself, unfortunately. And squash-merging would be nice :-)

nomeata:
Odd, why is the cla check still pending?

ali-dfinity:
droid update

PR #38, opened 2022-04-27
feat: implementation-specific error codes

roman-kashitsyn:
This change adds a new field to reject responses:
implementation-specific textual error code.  Tools such as DFX could use
these codes to provide better diagnostic to users and propose recovery
paths.

The spec reserves error codes starting with "IC" for the DFINITY
implementation.

In future, we might fix semantics of specific error codes in the spec.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -406,6 +406,10 @@ If the status is `rejected`, then this path contains the reject code (see <<reje
 +
 If the status is `rejected`, then this path contains a textual diagnostic message, else it is not present.
 
+* `/request_status/<request_id>/error_code` (text)
++
+If the status is `rejected`, then this path contains an implementation-specific error code (see <<error-codes>>), else it is not present.
+
 NOTE: Immediately after submitting a request, the request may not show up yet as the Internet Computer is still working on accepting the request as pending.
 
 NOTE: Request statuses will not actually be kept around indefinitely, and eventually the Internet Computer forgets about the request. This will happen no sooner than the request’s expiry time, so that replay attacks are prevented.
@@ -585,6 +589,7 @@ If the call resulted in a reject, the response is a CBOR map with the following 
 * `status` (`text`): `rejected`
 * `reject_code` (`nat`): The reject code (see <<reject-codes>>).
 * `reject_message` (`text`): a textual diagnostic message.
+* `error_code` (text): an implementation-specific textual error code (see <<error-codes>>).
 
 Canister methods that do not change the canister state can be executed more efficiently. This method provides that ability, and returns the canister’s response directly within the HTTP response.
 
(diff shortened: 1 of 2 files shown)


roman-kashitsyn:
I remember @nomeata had some opinion on this issue.

nomeata:
Quite likely that I had opinions, but I don't remember them… :-)


Are these error codes only ever provided by the system, or can they be given by the canister (in case of an explicit reject by the canister)?

roman-kashitsyn:
> Are these error codes only ever provided by the system, or can they be given by the canister (in case of an explicit reject by the canister)?

It's a good question. There were a few requests to allow custom reject codes; I guess error codes are one way to provide this feature. And if we reserve the IC namespace for system errors, there is a way to distinguish between replica errors and canister errors.

One detail that bothers me is that if a canister specifies custom error codes, we won't be able to report "ICXXX" (where ICXXX means "canister rejected your message"); we'll have to embed a custom canister message instead. I guess it's not a big deal because `reject_code` will say `CANISTER_REJECT` already in this case.

nomeata:
Yes, my thinking as well: the `CANISTER_REJECT` code clearly separates system from canister produced rejects, ans can be used by defensive clients to not get confused by a canister trying to fake a system-level reject.

roman-kashitsyn:
@Dfinity-Bjoern, we talked offline and I believe you were OK with the idea. Should we merge this?

roman-kashitsyn:
@AlexandraZapuc, I heard you might also be interested in this topic.

spec/index.adoc
```diff
@@ -735,6 +740,13 @@ The error message is guaranteed to be a string, i.e. not arbitrary binary data.
 
 When canisters explicitly reject a message (see <<system-api-requests>>), they can specify the reject message, but _not_ the reject code; it is always `CANISTER_REJECT`. In this sense, the reject code is trustworthy: If the IC responds with a `SYS_FATAL` reject, then it really was the IC issuing this reject.
 
+[#error-codes]
+=== Error codes
+
+Implementations of the API can provide additional details for rejected messages in the form of a textual label identifying the error condition.
+API clients can use these labels to handle errors programmatically or suggest recovery paths to the user.
+The specification reserves error codes matching the regular expression `IC[0-9]+` (e.g., `IC502`) for the DFINITY implementation of the API.
```
AlexandraZapuc, line 748: Do we plan to document the exact error codes somewhere?

AlexandraZapuc (approved):

spec/index.adoc
roman-kashitsyn: Yes, we will have short articles about each error code, and DFX will link to these articles. We aren't going to describe the errors in the spec for now.

Dfinity-Bjoern:
After discussion with @ulan and @dsarlis: If we depend on convention for parsing the codes anyway (after all, they're optional strings), why not just prefix the error text with "IC123:"?

roman-kashitsyn:
> After discussion with @ulan and @dsarlis: If we depend on the convention for parsing the codes anyway (after all, they're optional strings), why not just prefix the error text with "IC123:"?

I'd prefer to obtain the error code in a more structured way; this makes writing the client code more pleasurable.
In addition, if we allow canisters to specify custom reject codes, as discussed above, we won't be able to rely on convention anymore.

roman-kashitsyn:
@Dfinity-Bjoern, I discussed this issue again with @dsarlis, and we agreed that having error codes explicitly is better than relying on reject message conventions. Could we merge this change?

Dfinity-Bjoern (approved):

PR #40, opened 2022-05-03
Fix typos

ulan:
The definition of callback invocation uses undefined `Ctxt_id2`.
That looks like a typo and should be `Ctxt_id`.

This also fixes `cycles_refundend` to `cycles_refunded`.


ulan:
I also filed https://github.com/dfinity/interface-spec/issues/41

Dfinity-Bjoern (approved):
Thanks Ulan!

PR #44, opened 2022-05-05
chore: extract system API to a separate file

lwshang:
Then the file can be refer to generate binding for Rust CDK.

### Minor changes
* Several missing `;` signs were added.
* `ic0.performance_counter : (type : i32) -> (counter : i64);` has `type` as argument name which conflicts with Rust syntax. So I rename it to `counter_type`.


chenyan-dfinity:
Not sure if we want to use `.rs` for the system API. The `i32/i64` type refers to the Wasm type, not Rust type.

spec/index.adoc
```diff
@@ -945,73 +945,10 @@ In the reply callback of a <<system-api-call,inter-canister method call>>, the a
...
-
-ic0.time : () -> (timestamp : i64);                                         // *
-ic0.performance_counter : (type : i32) -> (counter : i64);                  // * s
-
-ic0.debug_print : (src : i32, size : i32) -> ();                            // * s
-ic0.trap : (src : i32, size : i32) -> ();                                   // * s
-....
+[source,rust]
+----
+include::{example}ic0.rs[]
```
nomeata, line 950:
```suggestion
include::{example}ic0.txt[]
```

jwiegley (approved):

PR #48, opened 2022-05-23
Add a maximal response size parameter for the HTTP Request feature

yotamhc:
We would like to add a new parameter for the HTTP request feature to specify the maximal size of the response.
This will be used for charging the requests and for limiting the bandwidth used by the feature.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1637,7 +1637,7 @@ For security reasons, only HTTPS connections are allowed (URLs must start with `
 
 The maximal size of a request URL is 2048 bytes.
 
-Each request should specify the maximal expected size for the response. The maximal size of a response is `2MiB`. Therefore, only the first number of bytes as specified in the request (up to 2MiB), will be returned if a response is larger than these limits. The 2MiB size limit also applies to the value returned by the `transform` function.
+Each request should specify the maximal expected size for the response. The upper limit on the size of a response is `2MiB`. Therefore, only the first number of bytes as specified in the request (up to 2MiB), will be returned if a response is larger than these limits. The 2MiB size limit also applies to the value returned by the `transform` function.
 
 
 The following parameters should be supplied for the call:
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1637,11 +1637,13 @@ For security reasons, only HTTPS connections are allowed (URLs must start with `
 
 The maximal size of a request URL is 2048 bytes.
 
-The maximal size of a response is `2MiB`. Therefore, only the first `2MiB` will be returned if a response is larger than this size. This size limit also applies to the value returned by the `transform` function.
+Each request should specify the maximal expected size for the response. The maximal size of a response is `2MiB`. Therefore, only the first number of bytes as specified in the request (up to 2MiB), will be returned if a response is larger than these limits. The 2MiB size limit also applies to the value returned by the `transform` function.
+
 
 The following parameters should be supplied for the call:
 
 - `url` - the requested URL
+- `max_response_size` - the maximal size of the response. The call will be charged based on this parameter. The raw response (before applying the `transform` function) will be truncated if necessary, but can be longer than the given value.
 - `method` - currently, only GET is supported
 - `headers` - list of HTTP request headers and their corresponding values
 - `transform` - an optional function that transforms raw responses to sanitized responses. If provided, the calling canister itself must export this function.
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1647,14 +1647,14 @@ Currently, only the `GET` method is supported for HTTP requests.
 
 For security reasons, only HTTPS connections are allowed (URLs must start with `https://`). The IC uses industry-standard root CA lists to validate certificates of remote web servers.
 
-Each request can specify the maximal expected size for the response. The upper limit on the size of a response is `2MiB` if no maximal size value is specified.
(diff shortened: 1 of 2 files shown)


spec/index.adoc
```diff
@@ -1637,11 +1637,13 @@ For security reasons, only HTTPS connections are allowed (URLs must start with `
 
 The maximal size of a request URL is 2048 bytes.
 
-The maximal size of a response is `2MiB`. Therefore, only the first `2MiB` will be returned if a response is larger than this size. This size limit also applies to the value returned by the `transform` function.
+Each request should specify the maximal expected size for the response. The maximal size of a response is `2MiB`. Therefore, only the first number of bytes as specified in the request (up to 2MiB), will be returned if a response is larger than these limits. The 2MiB size limit also applies to the value returned by the `transform` function.
```
nomeata, line 1640: Silent truncation sounds like a footgun.

Can’t we return an error in that case? HTTP error 413 isn’t quite the right thing, because that seems to be about the request size. Maybe 206 Partial Content?

Or maybe better to issue an IC-level reject here.

yotamhc: This is not that silent if the user specifies the maximal size.
The idea is to encourage users to send requests that result in smaller responses, as we are worried that this feature could use lots of bandwidth otherwise. I do see cases where the first X bytes of a bigger response are important.
We do not want to reject the call if the first X requested bytes can be served..

nomeata: But it is silent! How do I know whether the request was truncated, or just happened to have the size I was expecting? And if the user does not specify the size explicitly, it’s still 2MiB, right?

>  We do not want to reject the call if the first X requested bytes can be served..

Would that use-case be better serves using the `Ranges:` header of HTTP? This is an existing protocol lettings users request parts of a response?

We could either hope that most servers support this, or implement that ranges in the HTTP component.

Or at least add an `IC-response-truncated:` header maybe? Still a footgun, I’d say.

spec/index.adoc
```diff
@@ -1637,11 +1637,13 @@ For security reasons, only HTTPS connections are allowed (URLs must start with `
 
 The maximal size of a request URL is 2048 bytes.
 
-The maximal size of a response is `2MiB`. Therefore, only the first `2MiB` will be returned if a response is larger than this size. This size limit also applies to the value returned by the `transform` function.
+Each request should specify the maximal expected size for the response. The upper limit on the size of a response is `2MiB`. Therefore, only the first number of bytes as specified in the request (up to 2MiB), will be returned if a response is larger than these limits. The 2MiB size limit also applies to the value returned by the `transform` function.
```
qj-yu, line 1640: Shall we specify the unit expected for this field? It feels like `KB` is the intended unit, in that case, we probably want to explain the upper limit in targeted unit.

stefan-at-dfinity: ... the way I read it, the unit is `bytes`.

stefan-at-dfinity: But, yes, we should make it more explicit.

spec/index.adoc
stefan-at-dfinity: As an engineer, I see little usage in receiving only part of a response, tbh. Thus, I agree that we should fail hard—and early.

yotamhc: Users could use ranges in their request.
I meant that it is not silent as the user specifies the size and the response should have a `Content-Length` header so they can know if it was truncated. But I also see the un-usefulness in this kind of truncated response...

nomeata: Is that header guaranteed to be there? If the server streams responses probably not.

stefan-at-dfinity: Indeed, if `Transfer-Encoding` is set, `Content-Length` must not be set ...

https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html

While for most common use cases, Content-Length is set—there are sloppy http implementations out there.

Dfinity-Bjoern: I agree with @nomeata's concern. We should not discard silently.

AlexandraZapuc: I also think it's better to reject if the response does not fit the limit provided. Having this parameter that the users can set makes them more aware of the limitations.

spec/index.adoc
yotamhc: please see updated text

spec/index.adoc
yotamhc: Please see updated text

spec/index.adoc
```diff
@@ -1635,13 +1635,14 @@ Currently, only the `GET` method is supported for HTTP requests.
 
 For security reasons, only HTTPS connections are allowed (URLs must start with `https://`). The IC uses industry-standard root CA lists to validate certificates of remote web servers.
 
-The maximal size of a request URL is 2048 bytes.
-
-The maximal size of a response is `2MiB`. Therefore, only the first `2MiB` will be returned if a response is larger than this size. This size limit also applies to the value returned by the `transform` function.
+Each request can specify the maximal expected size for the response. The upper limit on the size of a response is `2MiB`.
+If the response is larger than the specified size, or than `2MiB` if no maxmial size is specified, an error will be returned.
```
AlexandraZapuc, line 1639: I would simplify the third sentence. 
 
 ```suggestion
Each request can specify the maximal expected size for the response. The upper limit on the size of a response is `2MiB` if no maximal size value is specified.
An error will be returned when the response is larger than the maximal size. 
```

Dfinity-Bjoern (approved):
Looks good – thanks!

AlexandraZapuc (approved):

PR #50, opened 2022-06-02
EXC-1122: Update idle_cycles_burned

AlexandraZapuc:
Based on previous MR  https://github.com/dfinity/interface-spec/pull/31, which introduced `idle_cycles_burned_seconds`, and a few follow-up discussions, there was reached the agreement that we should avoid `float` types in the replica. 

This MR:
* change `idle_cycles_burned_seconds` the type from float to nat.
* due to the previous change,  we lose some precision. Therefore, to increase the precision we will provide `idle_cycles_burned_hour`.
* fix: remove `freezing_threshold: nat;` from the canister status. It is already specified in the `definite_canister_settings `.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -100,8 +100,7 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
-      freezing_threshold: nat;
-      idle_cycles_burned_per_second: float64;
+      idle_cycles_burned_per_hour: nat;
   });
   delete_canister : (record {canister_id : canister_id}) -> ();
   deposit_cycles : (record {canister_id : canister_id}) -> ();
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -100,7 +100,7 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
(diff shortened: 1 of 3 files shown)


spec/ic.did
```diff
@@ -49,8 +49,7 @@ service ic : {
       module_hash: opt blob;
       memory_size: nat;
       cycles: nat;
-      freezing_threshold: nat;
-      idle_cycles_burned_per_second: float64;
+      idle_cycles_burned_per_hour: nat;
```
AlexandraZapuc, line 52: Is `cycles per hour` sufficient for the increase in precision? I was thinking maybe to use ` per day`.

Dfinity-Bjoern: It seems that based on 5 Tc/GB/Year, I get to 5000 c/Byte/year, or around 0.5 c/Byte/hour. Which would mean that by specifying the price per hour we indeed give up on some precision. I don't think it's a big deal either way, but based on the above argument I have a slight preference for the day option.

ulan:
Lgtm, thanks! +1 to `idle_cycles_burned_per_day`.

Dfinity-Bjoern (approved):
Thanks Alexandra!

PR #51, opened 2022-06-03
Public Spec: Callbacks as named, exported functions

roman-kashitsyn:
Previously, callbacks were identified by indices in an exported table,
which are hard to keep stable across upgrades. By identifying callbacks
with names, and exposing them via named function exports, those
developers who aim to have stable callbacks across upgrades have a
reasonable chance.

Since we are touching this anyways, it also changes the size of the
`env` to `i64`; this does not affect the system-side implementation
greatly, but allows canisters to do interesting tricks, such as never
re-using an `env` value, or encoding both version information and a
32-bit counter in it.

This PR is an iteration of #183, which did away with “names” for
callbacks completely and only used the `env` value for dispatch, using a
single entry point, but subsequent discussion declared that as
undesirable.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1139,7 +1139,7 @@ Subsequent calls to the following functions set further attributes of that call,
 --
   ic0.call_on_cleanup : (fun_src : i32, fun_size : i32, env : i64) -> ()
 
-This trap if the blob indicated by `fun_src/_size` is not a valid utf8-encoded name.
+This traps if the blob indicated by `fun_src/_size` is not a valid utf8-encoded name.
 
 If a cleanup callback (of type `+(env : i32) -> ()+`) is specified for this call, it is executed if and only if the `reply` or the `reject` callback was executed and trapped (for any reason).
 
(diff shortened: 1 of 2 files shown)


roman-kashitsyn:
This proposal was authored by @nomeata in the archived private repo. The patch didn't apply cleanly; I hope my resolution of conflicts didn't break anything.

nomeata:
> Since we are touching this anyways, it also changes the size of the env to i64;

8 bytes ought to be enough for everyone?

At this point, I wonder if we should just allow blobs as environments. This allows canisters to put even more information into the callback without having to touch it’s own memory, which can for example be otherwise completely stateless. It’d have to be accounted for, of course, but that seems doable.

spec/index.adoc
```diff
@@ -1112,32 +1112,39 @@ When handling an update call (or a callback), a canister can do further calls to
...
 Subsequent calls to the following functions set further attributes of that call, until the call is concluded (with `ic0.call_perform`) or discarded (by returning without calling `ic0.call_perform` or by starting a new call with `ic0.call_new`.)
 --
 
 * {blank}
 +
 --
-  ic0.call_on_cleanup : (fun : i32, env : i32) -> ()
+  ic0.call_on_cleanup : (fun_src : i32, fun_size : i32, env : i64) -> ()
+
+This trap if the blob indicated by `fun_src/_size` is not a valid utf8-encoded name.
```
jplevyak, line 1142:
```suggestion
This traps if the blob indicated by `fun_src/_size` is not a valid utf8-encoded name.
```

Dfinity-Bjoern:
Closing the PR for now because there is no active work and it is unclear that converting it to markdown is beneficial at this point. The PR can be converted and reopened at any point in time, such as when the engineering teams actually have capacity to work on the implementation.

PR #57, opened 2022-06-28
IC-530: Adding info on POST guarantees

yotamhc:


spec/index.adoc
```diff
@@ -1647,7 +1647,11 @@ The responses for all identical requests must match too. However, a web service
 
 For this reason, the calling canister can supply a transformation function, which the IC uses to let the canister sanitize the responses from such unique values. The transformation function is executed separately on the corresponding response received for a request. The final response will only be available to the calling canister.
 
-Currently, only the `GET`, `HEAD`, and `POST` methods are supported for HTTP requests. Note that when using `POST`, the calling canister must make sure that the remote server is able to handle idempotent requests sent from multiple sources. This may require, for example, to set a certain request header to uniquely identify the request.
+Currently, only the `GET`, `HEAD`, and `POST` methods are supported for HTTP requests.
```
dietersommer, line 1650: Would remove the "only"

spec/index.adoc
```diff
@@ -1647,7 +1647,11 @@ The responses for all identical requests must match too. However, a web service
...
+
+It is important to note the following for the usage of the `POST` method:
+- The calling canister must make sure that the remote server is able to handle idempotent requests sent from multiple sources. This may require, for example, to set a certain request header to uniquely identify the request.
+- There is no guarantee on which replica makes the first `POST` request. Since a single replica may be malicious, it might be that the first `POST` request to reach the remote server is maliciously modified. Therefore, it is recommended to verify the content of the request at the remote server and/or use additional integrity checks.
```
dietersommer, line 1654: Let's add a comment also that even in case no consensus is reached on the response, the modification on the server might have been done. E.g.: "Even in the case that no consensus is obtained on the response of a POST request, the server may have made a state change, for example based on the first response that reached the server."

spec/index.adoc
Dfinity-Bjoern, line 1650:
```suggestion
Currently, the `GET`, `HEAD`, and `POST` methods are supported for HTTP requests.
```

spec/index.adoc
Dfinity-Bjoern: We should try to avoid describing protocol artifacts – such as mentioning the "replica" or "consensus on response". Can we just describe that the IC may make multiple requests and that a response isn't guaranteed?

yotamhc: @Dfinity-Bjoern the problem here is that a malicious replica may even change the request, not just the response, and if it happened to be the first one, then it may really cause problems.

yotamhc: For example, if you use an API like Stripe for payments and you set an idempotency key, say to value "1", a malicious replica can change that key to "2" and submit a completely different request to the remote server. Even if it checks subsequent requests before making the change (and Strip don't do that), this would look like a completely different request.

nomeata: But a malicious replica could use any APIs anyways.

The real problem seems to be that in order to use _authenticated_ APIs, the canister has to know a secret, and keeping secrets in canister already requires trust into all replicas – that is a fact of the IC independent of the Internet Computer.


The only safe way to use an authenticated external API that I see if that API supports authentication via cryptographic signatures, and then use the ECDSA signature support to sign the request.

yotamhc: Then I wonder whether we should say anything about this point in general. E.g., should we advise against calling APIs that require API keys?

yotamhc: We can have something like:
`- There are no confidentiality guarantees on the request content, or any guarantees on the actual request that will be sent out. The request will be sent out multiple times. There is no guarantee that all copies will be identical.`

nomeata: Sounds good! Maybe

> There is no guarantee that all sent requests are as specified by the canister.

What guarantees can we give, though (assuming a majority of the nodes is honest)?

Maybe: If the canister receives a response, the then at least one request was sent that actually matches the canister's request, and that the response was to that request?

yotamhc: This depends on the remote server to some extent. E.g., if the remote server only returns 200 OK for all such requests without any identification of the request on the response, the canister will receive a valid response.

nomeata: But my claim is still true, right?

The canister will only receive a response if the (honest) majority of the nodes have received identical responses from the server. And they will only receive a response if they sent a request. And honest replicas send the correct request.

yotamhc: Yes, but this doesn't really tell about the issue that may arise from a malicious request being posted. We can choose to ignore it here, as it might not be the right place to present it.

nomeata: Right, that issue also needs to be explained. But if we only say “anything can happen” that’s a bit disheartening, so I thought what we could promise :-)

yotamhc: So how about the following:
```
- There are no confidentiality guarantees on the request content. There is no guarantee that all sent requests are as specified by the canister. If the canister receives a response, then at least one request that was sent matched the canister's request, and the response was to that request.
```

nomeata: Perfect

Dfinity-Bjoern (approved):
Thanks @yotamhc looks good now!

PR #58, opened 2022-06-29
Fix use of hash_of_map

Dfinity-Bjoern:
Fixes an inconsistency.


mraszyk:
This PR is subsumed by [that one](https://github.com/dfinity/interface-spec/pull/61) so I propose to close it.

PR #60, opened 2022-07-01
Add a note on cycle cost of invoking ecdsa APIs

ninegua:


spec/index.adoc
```diff
@@ -1623,6 +1623,8 @@ The return result is an extended public key consisting of an ECDSA `public_key`,
 This call requires that the ECDSA feature is enabled, and the `canister_id` meets the requirement of a canister id.
 Otherwise it will be rejected.
 
+Invoking this API will cost cycles. The concrete cost is yet to be determined.
```
nomeata, line 1626: Isn't that true for any of the management canister calls?

ninegua: @Dfinity-Bjoern ^^^

ninegua: Maybe there is a difference between "cycles deducted from canister" and "cycles must be send with the call". Not sure how to best put it...

nomeata: Ah, yes! If it behaves differently there it's worth pointing out. But why should it?

spec/index.adoc
```diff
@@ -1623,6 +1623,8 @@ The return result is an extended public key consisting of an ECDSA `public_key`,
 This call requires that the ECDSA feature is enabled, and the `canister_id` meets the requirement of a canister id.
```
adambratschikaye, line 1623: Not blocking this change, but for a follow-up PR:

Not sure if it makes sense to mention that the `ECDSA` feature needs to be enabled. It doesn't need to be enabled on the calling canister's subnet anyway - only on some subnet holding the key. But if it wasn't enabled on any subnet holding the key then the key wouldn't even exist...

Maybe we should say something like "the key_id must exist on some IC subnet".

spec/index.adoc
adambratschikaye: We have some management canister calls which are free (as in, the canister doesn't need to send any cycles with the request, and any cycles they do send are refunded).

PR #61, opened 2022-07-05
Upload Isabelle/HOL model and update textual specs

mraszyk:
This PR contains a translation and adjustments of Helge's Coq model for a subset of state transitions, proofs of cycle monotonicity for these transitions, and updates to the textual specs based on the formal model.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2545,8 +2545,8 @@ The following is an incomplete list of invariants that should hold for the abstr
 * Referenced call contexts exists:
 +
 ....
-∀ CallMessage M ∈ S.messages.  M.origin.calling_context ∈ S.call_contexts
-∀ ResponseMessage M ∈ S.messages. M.origin.calling_context ∈ S.call_contexts
+∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ∈ S.call_contexts
+∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ∈ S.call_contexts
 ∀ _ ↦ Ctxt ∈ S.call_contexts:
   if Ctx.needs_to_respond:
      Ctxt.origin.calling_context ∈ S.call_contexts
@@ -2953,8 +2953,8 @@ Conditions::
 ....
     S.call_contexts[Ctxt_id].needs_to_respond = true
     S.call_contexts[Ctxt_id].origin ≠ FromHeartbeat
-    ∀ CallMessage M ∈ S.messages. M.origin.calling_context ≠ Ctxt_id
-    ∀ ResponseMessage M ∈ S.messages. M.origin.calling_context ≠ Ctxt_id
+    ∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
+    ∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
     ∀ ctxt_ids ∈ dom(S.call_contexts).
         S.call_contexts[ctxt_ids].needs_to_respond
         ==> S.call_contexts[ctxt_ids].origin.calling_context ≠ Ctxt_id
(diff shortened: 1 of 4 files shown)


nomeata:
That's impressive and exiting!

Any plans to use document generation from Isabelle to replace the formal section in the spec :-)

Is Helge's Coq model the same as John's, or were that separate attempts?

nomeata:
Also, I’m a fan of “no code in the repo without CI testing it” to prevent bitrot. Is running the isabelle build on CI planned?

mraszyk:
> Is Helge's Coq model the same as John's, or were that separate attempts?

The two models were developed independently and can also be considered "incomparable":
John's model (Coq): all state transitions, no proofs about cycles
Helge's model (Coq): a subset of transitions, proofs about cycle monotonicity

nomeata:
Ah, in my optimistic enthusiasm I missed the “subset of” in the PR description :-D

mraszyk:
> Ah, in my optimistic enthusiasm I missed the “subset of” in the PR description :-D

As the immediate next step I am going to extend the model and cycle monotonicity proofs to cover all transitions. I have also formulated the transitions so that the code of the preconditions and new states can be automatically exported into Haskell/OCaml to be integrated in ic-ref in the future.

nomeata:
Amazing!

spec/index.adoc
```diff
@@ -2896,8 +2896,8 @@ else
   S with
     messages = Older_messages · Younger_messages
     balances[M.receiver] =
-      (S.balances[M.receiver] + (if Is_response then MAX_CYCLES_PER_RESPONSE else MAX_CYCLES_PER_MESSAGE))
-      - Cycles_used
+      S.balances[M.receiver] +
+      ((if Is_response then MAX_CYCLES_PER_RESPONSE else MAX_CYCLES_PER_MESSAGE) - Cycles_used)
```
mraszyk, line 2900: @bogwar Could you please confirm that this corresponds to what the mainnet implementation does?

mraszyk: Thank you for the confirmation!

spec/index.adoc
```diff
@@ -2704,14 +2704,14 @@ Conditions::
 ....
     S.messages = Older_messages · CallMessage CM · Younger_messages
     (CM.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ CM.queue)
-    S.canister_status[CM.callee] = Stopped or S.canister_status[CM.callee] = Stopping or balances[CM.callee] < freezing_limit(S, CM.callee)
+    S.canister_status[CM.callee] = Stopped or S.canister_status[CM.callee] = Stopping _ or balances[CM.callee] < freezing_limit(S, CM.callee)
```
Dfinity-Bjoern, line 2707: I don't understand this change.

spec/index.adoc
```diff
@@ -2829,9 +2829,8 @@ Conditions::
     Available = S.call_contexts[M.call_contexts].available_cycles
     ( M.entry_point = PublicMethod Name Caller Data
       Arg = { data = Data; caller = Caller }
-      (F = Mod.update_methods[Name](Arg, Env, Available)
-      or
-      (F = query_as_update(Mod.query_methods[Name], Arg, Env))
+      (F = Mod.update_methods[Name](Arg, Env, Available)) or (F = query_as_update(Mod.query_methods[Name], Arg, Env))
+      Name ∉ dom(Mod.update_methods) or Name ∉ dom(Mod.query_methods)
```
Dfinity-Bjoern, line 2833: I was wondering whether this is the right place to put the condition. It is definitely true, but it seems that's kind-of an invariant for all wasm modules and should be true after installation.

spec/index.adoc
```diff
@@ -2890,17 +2888,16 @@ then
        call_contexts[M.call_context].needs_to_respond = false
        call_contexts[M.call_context].available_cycles = 0
 
+    balances[M.receiver] = New_balance
```
Dfinity-Bjoern, line 2891: Why move this? I mean, doesn't seem to break things, just wondering.

Dfinity-Bjoern (approved):
Overall looks great, thanks @mraszyk! Would be grateful for responses to the few comments.

spec/index.adoc
mraszyk: I agree. It could move to the invariants section and be established by canister creation and upgrade. Let me give it a try!

mraszyk: Done!

spec/index.adoc
Dfinity-Bjoern: Thanks for the clarification.

PR #67, opened 2022-07-13
Fix the definitions of effective canister IDs and subnet range checking in certificates

oggy-dfin:
This fixes a couple of issues:
1. `create_canister` cannot be called by users, and the effective canister ID
should be undefined in this case. This wasn't made explicit before.

2. `provisional_create_canister_with_cycles` can be called with the effective
canister ID of `aaaaa-aa` (this is done by `dfx`, for example). However, this
makes it impossible to read the status of this call while following the rules as
specified in the interface spec. Namely, the result of read_state will be
certified by the local replica, but the subnet ranges in the certificate won't
include `aaaaa-aa`. Thus, certificate checking as currently specified will fail.
This relaxes the check such that it's not performed for `aaaaa-aa`. This is not
pretty, but it should be correct. First,
`provisional_create_canister_with_cycles` is the ONLY call for which `aaaaa-aa`
is a valid effective canister ID. Thus, for checking statuses of requests, we're
only disabling the check for one request that is used only in local deployments.
Second, for checking `/time` endpoints, I suspect we are OK with receiving the
response from any subnet. For `/subnet`, I suspect we can also let any subnet
answer that for `aaaaa-aa`, (in particular, to allow `dfx` to learn the subnet
of the local replica). For all other paths, I assume that the read_state request
will fail when using `aaaaa-aa`.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -549,7 +549,7 @@ The HTTP response to this request consists of a CBOR map with the following fiel
 
 * `certificate` (`blob`): A certificate (see <<certification>>).
 +
-If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation’s canister id range (see <<certification-delegation>>).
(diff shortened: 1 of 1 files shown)


Dfinity-Bjoern (approved):
We'll likely remove provisional canister create from the main spec in the future, but for now it seems a good idea to be consistent.

PR #68, opened 2022-07-13
define freezing_limit in terms of idle_cycles_burned_per_day

mraszyk:
Because `freeze_threshold` can be equal to zero, the equation
```
idle_cycles_burned_per_day = freezing_limit * seconds_per_day / freeze_threshold
```
might yield undefined value (if `freeze_threshold = 0`).

Hence, we make  `idle_cycles_burned_per_day` unspecified and derive `freezing_limit` from  `idle_cycles_burned_per_day` and `freeze_threshold`.


Dfinity-Bjoern (approved):
Thanks for improving consistency between spec and code!

PR #69, opened 2022-07-19
Replace set_controller with update_settings in the prose section

dsarlis:
We were still using `set_controller` as one of the management methods that we can derive an effective canister id on, but this has been superseded by `update_settings` for a while, so it's good to update that part of the interface spec.


Dfinity-Bjoern (approved):
Thank you for fixing it @dsarlis

PR #70, opened 2022-07-20
Update query call prose to cover the case of frozen canisters

dsarlis:
Canisters are currently allowed to execute query calls even if when they are frozen. This can lead to confusion as canisters with heavy or query-only workload can look as if they're completely fine and can catch canister owners off-guard as their canister can get uninstalled even they are not proactive enough to monitor its cycles balance closely. On top of this, query calls will be charged cycles anyway in the future, so disallowing queries to frozen canisters seems like a reasonable thing to enforce.

### DIFF
--- a/spec/changelog.adoc
+++ b/spec/changelog.adoc
@@ -6,6 +6,7 @@
 
 * Spec: Canister access to performance metrics
 * Spec: User delegations include a principal scope
+* Spec: Query calls are rejected when the canister is frozen
 
 [#0_18_5]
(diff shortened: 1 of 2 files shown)


Dfinity-Bjoern (approved):
Thanks Dimitris! Can you add a changelog and then merge?

dsarlis:
> Thanks Dimitris! Can you add a changelog and then merge?

@Dfinity-Bjoern can you please check if it looks good and then I can merge.

PR #71, opened 2022-07-21
chore: rename thys -> theories

roman-kashitsyn:
Rationale: the `thys` name is confusing and hard to pronounce.
Let's use `theories` instead.
This change also adds the Haskell code that the isabelle generates to .gitignore.


mraszyk (approved):
Looks good to me!

Dfinity-Bjoern (approved):
LGTM

PR #72, opened 2022-07-21
factor cycles_used out of message_execution arguments

mraszyk:


bogwar (commented):
Thanks Martin! 
It looks good, but the Nat type used in the result is mysterious. Could you also introduce CyclesUsed type; that would improve readability.

Dfinity-Bjoern:
I discussed that with @bogwar and think it makes sense. I would also appreciate @nomeata's opinion here!

mraszyk:
The corresponding replica change has been merged: https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/6230

Dfinity-Bjoern (approved):
Thanks Martin!

spec/index.adoc
```diff
@@ -2301,8 +2301,9 @@ UpdateFunc = WasmState -> Trap | Return {
   new_certified_data : NoCertifiedData | Blob
   response : NoResponse | Response;
   cycles_accepted : Nat;
+  cycles_used : Nat;
 }
-QueryFunc = WasmState -> Trap | Return Response
+QueryFunc = WasmState -> Trap | Return (Response, Nat)
```
nomeata, line 2306: Such a plain `Nat` makes it hard to read, and these types may be the first someone looks at.

Maybe introduce records for the result of running a query and for the heartbeat, with a named field?

nomeata (changes requested):
I'm confused that the signature of query and heartbeat functions are changed, but no change in the section that implements these signature, in https://internetcomputer.org/docs/current/references/ic-interface-spec/#concrete-canisters

nomeata:
(sorry for the delay, I was traveling and only  Björn's comment brought it back on my radar)

spec/index.adoc
mraszyk: I intend to introduce a type synonym Cycles for Nats that represent cycles. Would it help or should the response still be a record in your opinion?

nomeata: cycles is better, as it indicates the unit, but not the semantics. At least `CyclesConsumed`?

Also, don't trapping computations also consume cycles?

mraszyk: I wanted to avoid switching between various type synonyms (e.g., CyclesConsumed, CyclesRefunded etc.) because type conversions between type synonyms are not applicable and type (synonym) changes without type conversions could be confusing, too. What do you think?

Trapping computations consume `if Is_response then MAX_CYCLES_PER_RESPONSE else MAX_CYCLES_PER_MESSAGE` cycles because they're not refunded anything and thus there is no need explicitly track the amount of cycles consumed.

nomeata: A trapping computation still uses `MAX_CYCLES_PER_MESSAGE`? I thought this was fixed at some point… programmer expect trapping to be a cheap way out, and if it incurs the maximum cost instead that’s not great. @bogwar?

mraszyk: As for the return type `Return (Response, Nat)`, I lean towards turning it into a record `Return {response: Response, cycles_consumed: Nat}` now.

spec/index.adoc
```diff
@@ -4109,14 +4126,15 @@ callbacks = λ(callbacks, response, refunded_cycles, sysenv, available) → λ w
     if typeof(func) ≠ func (i32) -> () then Trap
 
     let es = ref { empty_execution_state with
```
mraszyk, line 4128: Is this execution state fully defined? E.g., there are no params updated.

spec/index.adoc
```diff
@@ -4109,14 +4126,15 @@ callbacks = λ(callbacks, response, refunded_cycles, sysenv, available) → λ w
...
-      wasm_state;
+      wasm_state = wasm_state;
     }
     func<es>(callbacks.on_cleanup.env)
     Return {
       new_state = es.wasm_state;
       new_calls = [];
       response = NoResponse;
       cycles_accepted = 0;
+      cycles_used = 0;
```
mraszyk, line 4137: Is this the right choice?

spec/index.adoc
mraszyk: Done.

spec/index.adoc
Dfinity-Bjoern: Do I understand correctly that this would mean: If the cleanup callback is run (and completes successfully), no cycles will be charged? In that case, it doesn't seem to be what I would expect. @bogwar ?

spec/index.adoc
Dfinity-Bjoern: At least that seems to be consistent with this part of the textual spec:
> During the execution of the cleanup function, only a subset of the System API is available (namely ic0.debug_print, ic0.trap and the ic0.stable_* functions). The cleanup function is expected to run swiftly (within a fixed, yet to be specified cycle limit) and serves to free resources associated with the callback.

The canister doesn't seem to have access to that data in the cleanup callback anyway.

spec/index.adoc
nomeata: Even if it isn’t the case right now in the implementation, I would certainly not want the spec to forbid the implementation to do more precise cycle accounting in trapping computation. This would would suggest something like
```
 QueryFunc = WasmState -> ((Trap | Return Response), Nat) 
```
and similarly for other functions.

spec/index.adoc
mraszyk: Your understanding is right. If we want to charge cycles, then I can just replace `0` by `es.cycles_used`. I was just not sure what the choice should be.

spec/index.adoc
bogwar: > A trapping computation still uses `MAX_CYCLES_PER_MESSAGE`? I thought this was fixed at some point… programmer expect trapping to be a cheap way out, and if it incurs the maximum cost instead that’s not great. @bogwar?

it looks that we have fixed that and we refund what was not consumed

spec/index.adoc
bogwar: We do charge cycles for executing the cleanup callback

mraszyk: I've just fixed that, but still the cycles consumed by the actual callback (in the try branch before `Trap`) are not considered.

spec/index.adoc
mraszyk: Let me then introduce a record for cycles_used in Trap, too:
```
Trap {
  cycles_used : Nat;
}
```

mraszyk: and update the Message execution transition...

spec/index.adoc
mraszyk: So is this resolved now? @bogwar

spec/index.adoc
nomeata: That also works of course, although I find it (on a philosophical level) less natural - trapping or returning is independent of cycle accounting. But no need to let that get in the way of a more convenient formulation :-)

spec/index.adoc
```diff
@@ -4109,14 +4126,15 @@ callbacks = λ(callbacks, response, refunded_cycles, sysenv, available) → λ w
     if typeof(func) ≠ func (i32) -> () then Trap
 
     let es = ref { empty_execution_state with
-      wasm_state;
+      wasm_state = wasm_state;
     }
     func<es>(callbacks.on_cleanup.env)
     Return {
```
mraszyk, line 4132: Interface Spec: "During the execution of the cleanup function, only a subset of the System API is available (namely ic0.debug_print, ic0.trap and the ic0.stable_* functions)."
However, we do not trap here if some of the functions have actually been invoked, e.g., for query methods, we perform the following checks:
```
  if es.cycles_accepted ≠ 0 then Trap {cycles_used = es.cycles_used;}
  if es.calls ≠ () then Trap {cycles_used = es.cycles_used;}
  if es.ingress_filter ≠ Reject then Trap {cycles_used = es.cycles_used;}
  if es.response = NoResponse then Trap {cycles_used = es.cycles_used;}
```

spec/index.adoc
mraszyk: I've fixed the method returns.

spec/index.adoc
mraszyk: They're now considered.

spec/index.adoc
nomeata: Good catch, probably simply missing.

(I am not sure if this style of “trap at the end if an unwanted getter/setter was used” is really helpful, compared to being more explicit about the context and trapping in the function directly…)

spec/index.adoc
```diff
@@ -2872,13 +2894,13 @@ State after::
 ....
 if
```
mraszyk, line 2895: Is this if-condition aligned with what the mainnet implementation does? @bogwar

spec/index.adoc
```diff
@@ -2921,8 +2943,8 @@ else
   S with
     messages = Older_messages · Younger_messages
     balances[M.receiver] =
-      S.balances[M.receiver] +
-      ((if Is_response then MAX_CYCLES_PER_RESPONSE else MAX_CYCLES_PER_MESSAGE) - Cycles_used)
+      (S.balances[M.receiver] + (if Is_response then MAX_CYCLES_PER_RESPONSE else MAX_CYCLES_PER_MESSAGE))
+      - min (R.cycles_used, (if Is_response then MAX_CYCLES_PER_RESPONSE else MAX_CYCLES_PER_MESSAGE))
```
mraszyk, line 2947: @bogwar please double check this part

spec/index.adoc
mraszyk: I've pushed a bunch of if's analogously by the query case (except that for callback cleanup, there can be no reply). I've added a note in my TODO list to handle trapping on illegal operations according to `spec/ic0.txt`, but let's do it outside this PR.

spec/index.adoc
bogwar: Yeap, that's correct (modulo the fact that we don't distinguish between requests and responses when it comes down to limiting how much we reserve/allow for execution)

bogwar (approved):
Lgtm! Great work Martin

spec/index.adoc
mraszyk: Changing `New_balance > if Is_response then 0 else freezing_limit(S, CM.callee)` to `New_balance >= freezing_limit(S, CM.callee)` looks sufficient to me for now.

PR #73, opened 2022-07-23
add CI for formal model

mraszyk:


basvandijk (approved):
LGTM.

Once we merge the interface-spec and ic-hs I would prefer to have all jobs defined with Nix. But let's nixify your `isabelle` job then.

Dfinity-Bjoern (approved):

PR #75, opened 2022-08-05
Update ic.did

lwshang:
Some types of `bitcoin` and `canister_http_request` have different definition in the IC implementation. 

Should we update the candid file in the spec or change the actual implementation? 

Also delete `user_id` since it is not used by any methods.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -19,7 +19,7 @@ type definite_canister_settings = record {
 type http_header = record { name: text; value: text };
 
 type http_response = record {
-  status: nat;
+  status: nat64;
   headers: vec http_header;
   body: blob;
 };
@@ -29,8 +29,9 @@ type ecdsa_curve = variant { secp256k1; };
 type satoshi = nat64;
 
 type bitcoin_network = variant {
-  mainnet;
-  testnet;
+  Mainnet;
+  Testnet;
+  Regtest;
 };
 
 type bitcoin_address = text;
@@ -52,8 +53,8 @@ type get_utxos_request = record {
   address : bitcoin_address;
   network: bitcoin_network;
   filter: opt variant {
-    min_confirmations: nat32;
-    page: blob;
+    MinConfirmations: nat32;
+    Page: blob;
   };
 };
 
@@ -112,12 +113,10 @@ service ic : {
   http_request : (record {
     url : text;
     max_response_bytes: opt nat64;
-    method : variant { get; head; post };
+    http_method : variant { get; head; post };
     headers: vec http_header;
     body : opt blob;
-    transform : opt variant {
-      function: func (http_response) -> (http_response) query
-    };
+    transform_method_name : opt text;
   }) -> (http_response);
 
   // Threshold ECDSA signature
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -1,5 +1,4 @@
 type canister_id = principal;
-type user_id = principal;
 type wasm_module = blob;
 
 type canister_settings = record {
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -112,7 +112,7 @@ service ic : {
   http_request : (record {
     url : text;
     max_response_bytes: opt nat64;
-    http_method : variant { get; head; post };
+    http_method : variant { GET; HEAD; POST };
     headers: vec http_header;
     body : opt blob;
     transform_method_name : opt text;


nomeata:
It's hardly a spec if the implementation defines what's true; it becomes implementation then ;-)

Don't we have spec acceptance tests? Shouldn't they prevent such accidental difference? It seems somewhere some shortcuts were taken…

Looking at the diff it seems mostly artifacts of Rust specific naming conventions (e.g. upper case enum tags). Ideally that is not the guiding light for a language agnostic interface, and I'd say it's more important to be consistent within the interface. If all the other enums and variants have lowercase names, then these here better follow suit.

spec/ic.did
```diff
@@ -29,8 +28,9 @@ type ecdsa_curve = variant { secp256k1; };
 type satoshi = nat64;
 
 type bitcoin_network = variant {
-  mainnet;
-  testnet;
+  Mainnet;
+  Testnet;
+  Regtest;
```
Dfinity-Bjoern, line 33: What does `Regtest` mean?

spec/ic.did
```diff
@@ -52,8 +52,8 @@ type get_utxos_request = record {
   address : bitcoin_address;
   network: bitcoin_network;
   filter: opt variant {
-    min_confirmations: nat32;
-    page: blob;
+    MinConfirmations: nat32;
```
Dfinity-Bjoern, line 55: That seems inconsistent with the naming conventions. Maybe we should actually fix the implementation instead? WDYT @ielashi @THLO

Dfinity-Bjoern (commented):
The HTTP API is not released, so changes there do not seem problematic. What concerns me a bit is that the changes appear to contradict our conventions.

spec/ic.did
lwshang: FYI, https://developer.bitcoin.org/examples/testing.html#regtest-mode
For local bitcoin development, `regtest` is the chain being connected. The relevant [PR here](https://github.com/dfinity/sdk/commit/8f14d6688364d182d0b9b36497fc84bda3c9012c).

spec/ic.did
```diff
@@ -112,12 +112,10 @@ service ic : {
...
     url : text;
     max_response_bytes: opt nat64;
-    method : variant { get; head; post };
+    http_method : variant { GET; HEAD; POST };
     headers: vec http_header;
     body : opt blob;
-    transform : opt variant {
-      function: func (http_response) -> (http_response) query
-    };
+    transform_method_name : opt text;
```
rumenov, line 118: Why are we changing this ? I will be on the side of using as much as type safety as possible .

dietersommer: Indeed, this is not clear, could someone clarify?

spec/ic.did
```diff
@@ -19,7 +18,7 @@ type definite_canister_settings = record {
 type http_header = record { name: text; value: text };
 
 type http_response = record {
-  status: nat;
+  status: nat64;
```
dietersommer, line 21: The team thinks that this should be changed to nat64 in the spec as it does not make sense to have an arbitrary precision number for representing HTTP status codes.

spec/ic.did
dietersommer: We will change the enumerations to lowercase in the implementation.
However, we wonder whether using "method" or "http_method" is better, considering that it is part of a record that has "httP" in its name.

spec/ic.did
```diff
@@ -112,12 +112,10 @@ service ic : {
   http_request : (record {
     url : text;
     max_response_bytes: opt nat64;
-    method : variant { get; head; post };
+    http_method : variant { GET; HEAD; POST };
```
dietersommer, line 115: We will change the enumerations to lowercase in the implementation.
However, we wonder whether using "method" or "http_method" is better, considering that it is part of a record that has "http" in its name.

spec/ic.did
dietersommer: We need to analyze whether both Motoko and Rust can support this. What constraints does this impose on the use of potential other languages like C?

spec/ic.did
nomeata: Probably not that important, but this goes against the advice by @rossberg to not use fixed-width numbers unless you do bit-fiddling or something like that.

Also note that `nat` is actually shorter on the wire than `nat64` for small values.

And even if, why `nat64` and not `nat16` or `nat32`?

nomeata: I wonder if this push to use `nat64` is mostly an artifact of the Rust bindings being designed in a way that makes the use of `nat` more inconvenient. This could probably be fixed by adding a newtype around `i64` to the Rust library that encodes/decodes to `nat`, and errors out when the incoming number is too large. Then Rust code can interact with `nat` with very little extra work whenever it wants to reject very large values anyways. WDYT, @chenyan-dfinity?

spec/ic.did
dietersommer: @nomeata : Can you help in clarifying whether and how one would implement in Rust against the current specification (lines 118-120)? Does it preclude other languages with a less expressive type system like C to be used for developing canisters?
I would really like to keep the spec as in 118-120. Thanks for any input!

spec/ic.did
chenyan-dfinity: `i128`/`u128` already maps to `int`/`nat` in Candid. The question is whether we want to optimize space for messages or the Rust memory.

We could define a newtype for `i64`, but people may still choose to ignore it. Just like we already have `candid::Int` and `candid::Nat` defined. Maybe we need an option to map all Rust number types to `nat`/`int` on the wire?

dietersommer: @nomeata : Thanks for your thoughts about this and the reference to Andreas Rossberg's opinion. So you would advise to keep it as a `nat`?

spec/ic.did
chenyan-dfinity: Motoko has no problem supporting function references. On the Rust side, you can decode it to https://docs.rs/candid/0.7.16/candid/types/reference/struct.Func.html. But there is a problem in encoding the reference type. There are some workaround implemented in `ic-proxy`, but not ideal. See this issue: https://github.com/dfinity/candid/issues/273

nomeata: I'll defer to @chenyan-dfinity, he's thenl expert here :-)

spec/ic.did
nomeata: Yes, I guess just use `i128` in rust and not worry any further :-)

dietersommer: @nomeata, @chenyan-dfinity :Sounds good! Thanks for your helpful inputs here!

spec/ic.did
Dfinity-Bjoern: I have a slight preference for just calling it "method" since it's in the context of an `http_request` and the HTTP spec refers to it as "request method" (not "HTTP method"). But I don't think it matters too much, I find it clear either way.

ielashi:
Regarding the Bitcoin API, we discussed it internally and we agree that consistency of the interface is more important, and we'll be updating the implementation to comply with the spec.

spec/ic.did
ielashi: Regtest is a tool that we're choosing to add to facilitate local development, but it's not part of production, so IMO it does not belong to the interface spec.

spec/ic.did
dietersommer: We had an extensive discussion on this internally with folks including @roman-kashitsyn, @Dfinity-Bjoern, and @dsarlis with the outcome that the "right" way of modeling this is analog to system functions like canister_init or canister_inspect_message as this best reflects the nature of a transform function, so we intend to change it to reflect that. This is in the interest of overall consistency with how things are modeled in the interface specification.

spec/ic.did
dietersommer: Agreed, than let's go with
```
method : variant { get; head; post };
```
as it is currently in the spec and change the implementation.

spec/ic.did
nomeata: I find it a very strange outcome that suddenly you are dropping to the Wasm level here.

The HTTP service is at a different level of abstraction as the low-level IC messaging protocol. `canister_init` and `canister_inspect_message` are clearly low-level canister interfaces; they are there to implement the Canister itself and build the IC inter-canister calling layer.

But the HTTP service could well be a separate canister! So it clearly is a protocol that can (and should) live on top of the lower-level, generic IC messaging service. The fact that the service’s implementation is magic and can’t be a normal canister does not affect how it should be interacted with, in my opinion.

So just like we reach all the management features not via additional Wasm-level entry points or System APIs, but by using IC-level calls, the same should be true for the HTTP component. 

And that means that that interface, including the “transform“ callback, should use the interface of ordinary inter-canister calls.

roman-kashitsyn: That's also a perfectly valid view. One minor problem is that the HTTP adapter always calls into the replica on the same machine, assuming the originating canister exposes the method. The adapter should make the transform call through boundary nodes if we want an arbitrary callback.

nomeata: Is that an observable difference? Sounds like an implementation detail/optimization to me: A replica can of course handle a query call to a canister it “owns” directly, if it wants to.

It’s a bit like inter-canister calls between canisters on the same subnet being handled without involving cross-net messaging, because it would be kinda silly to have cross-net messaging between a subnet and itself. But it wouldn't be wrong, and the abstract interface does not mandate one or the other.

roman-kashitsyn: > Is that an observable difference?

It will become observable if you specify a callback pointing into a canister on another subnet.

nomeata: Right, that's why the design so far had the extra requirement that the callback has to be to the same canister, don't it?

I admit it's a bit of a odd restriction, but much less if a severe layering violation than dropping to the wasm level here.

And with this restriction, it's not observable anymore, and the replica is welcome to resolve it locally.

dietersommer: Yes, it's essential for the trust model of the feature that the transform function is executed on the same replica that makes the network request.

dsarlis: > Right, that's why the design so far had the extra requirement that the callback has to be to the same canister, don't it?

I'm somewhat confused. If there was such a restriction on the design, then what's the point of providing a generic callback that would allow someone to call into any other canister (on the interface level at least)? Of course, the implementation can provide restrictions on who can be called (i.e. only resolve locally the request if you can) but then the interface design seems wrong to me.

Given the whole direction on the design of this feature, the choice of a system method seems the most natural to capture that:
1. This transform should only be ever called on the canister that made the original http request.
2. There's no sender, because we're not making a generic query call, but rather the system needs to invoke a special function on a canister at some point in time.

nomeata: The point of using a Candid function reference is
 * to get strong typing
 * to not have to hard code the name of the query 

The first is the interesting one, the second a nice to have.

Adding wasm layer entry points doesn't really scale. Imagine the churn it every new management canister feature was a wasm level system API function! This is the same here, I'd say.

Candid function references aren't use a lot yet, but they are there in the design for the IC programming model, and I think it's a step in the right direction to start using them.


If they are really that unwanted for some reason, at least use a normal query method with a well-known name, as in an earlier iteration of the adaptor design. That at least doesn't open up the abstract idea of what a canister is needlessly.

dsarlis: > to get strong typing

Yeah, in Motoko. If you write in Rust, it doesn't provide you with an extra benefit with the current library support.

> Adding wasm layer entry points doesn't really scale. Imagine the churn it every new management canister feature was a wasm level system API function! This is the same here, I'd say.

I agree with this point in general and I don't think we've been adding wasm entry points left and right -- we're generally cautious on when new wasm entry points are introduced.

In this case, it feels to me like the system method approach captures the intent better as I explained in my previous comment. My understanding is that we don't need a normal query method that's generally available to anyone, we need some method that the system (and the system only) can invoke when it needs to sanitise the http response.

> That at least doesn't open up the abstract idea of what a canister is needlessly.

I'm not sure I understand your point here. System methods are well established in the interface spec, we're not trying to introduce something completely new and unique.

nomeata: > Yeah, in Motoko. If you write in Rust, it doesn't provide you with an extra benefit with the current library support.

Exactly - about time to create more incentives to improve the Rust tooling with regards to inter-canister type safety :-)

> and the system only

Good point! The design with the function reference would be forward compatible to the world with real closures, a.k.a. capabilities, where the canister can generate a function reference that really can only be called by those that have it. Not sure if we ever get that, but it has been part if the vision since the beginning. So another argument for doing it this way :-)

The interface spec makes a clear abstract around the low level system API, in https://internetcomputer.org/docs/current/references/ic-interface-spec/#abstract-canisters, and then has a _separare_ section of how that maps to Wasm. Keeping that interface narrow and to the point is, I believe, useful to keep the system complexity manageable. And any feature that can be added without touching that interface should be added that way, unless there are really strong reasons to break that abstraction (e.g. a really bad performance hit).

We only have wasm canisters and that will probably stay that way, but I think a good design guidance is to imagine we had many different low level canister implementations (Wasm, Java Bytecode, Lua…), all with their own system API, but all implementing this abstract canister interface. How would we do it then? Very likely we'd stay above that interface, so that we don't have to change n different system APIs.

nomeata: BTW, I think we are debating exactly the same points as when we debated whether to use a virtual canister for management, or use system API. I had a hard time convincing Johan of the virtual canister approach, and only @ielashi managed to frame it in a way that convinced him. Maybe Islam can help coming to a conclusion here too :-)

ielashi: I am feeling the pressure now :-) I'm catching up here, so apologies in advance if I'm missing some major points.

I think we're all in agreement that we should keep the low level system API as minimal as possible, and the dilemma here is whether the transform method is a primitive that needs to be added to the system API, or whether it can be expressed using other primitives.

If I understand correctly, what makes the transform special, as Dimitris already mentioned, is the following:

1. There's no sender (and hence it's somewhat different from a query).
2. It should (ideally) not be invoked by any canister apart from the canister itself and the management canister.

IMHO, the first property isn't really necessary, and the second property can be enforced with other lower-level primitives.

Regarding "there's no sender", why is that the case? The canister is requesting from the management canister to do an HTTP request, and the _management canister_ then calls the transform method, so it seems perfectly consistent to have the management canister as the sender. Are there any reasons we need this special casing?

As for access control, which is more of a nice-to-have than a hard requirement, I believe we're also all in agreement that we should charge for queries at some point, and with that we'll need some mechanism to allow canisters to accept/reject processing queries (i.e. some kind of equivalent to `canister_inspect_message`). Once that primitive is available, then a canister can easily ensure that the transform method can only be called by itself and/or the management canister if it wishes.

With the above in mind, I'd be inclined to think the transform method is really no different than a query, and that it would be simpler to treat it as such rather than introduce a new primitive in the protocol.

As for the debate of whether to use a `func` or a `text` for the transform method, I don't have a strong opinion. I agree that `func` feels a bit too generic, but on the other hand I like the additional type safety that it can provide with the right support from a langauge's CDK, so I'd have a slight preference for that.

dsarlis: I feel there's still a misunderstanding here on what we mean when we say the management canister "calls" the transform function. 

The design is that the _replica_ calls the transform function and not really the management canister. If it was an actual call from the management canister, I'd expect it to be similar to an inter-canister call but the feature is designed in a way that the call does not resemble an inter-canister call. Perhaps it could have been designed and implemented in such a way but I simply don't see it in the current design and implementation.

That said, I feel we're just going in circles here. This discussion was brought up internally and a few people (tagged here and including myself) found the system method approach to be matching better. It seems there's another set of people that feel differently and we probably need to agree we disagree. Each side has presented their arguments, so I propose we leave it up to @Dfinity-Bjoern and other spec owners (and ultimately PR approvers) to set the direction.

nomeata: > The design is that the _replica_ calls the transform function and not really the management canister.

That sentence doesn’t make sense to me. The management canister isn’t a think, it is an abstract concept, a facade _for the replica_ (or the subnet, if you want). Just like if a canister calls the management canister to get some randomness, of course what’s happening really is the replica handles that.

But yes, I think the arguments are on the table. Poor @Dfinity-Bjoern will have to make a call :-)

Dfinity-Bjoern: I must admit @nomeata swayed me to the let-it-be-a-query direction. The main reason reason is that, from a canister perspective, the API of calling `aaaaa-aa` and then getting a "callback" (in quotes since it's not a callback in the interface spec sense) from `aaaaa-aa` makes a lot of sense. It is, in a sense, upward-compatible to providing the functionality through an actual canister. So if I have to make a call, then we go with:

- we leave it as a query that is invoked (with caller `aaaaa-aa`),
- we also keep the type-safe variant.

ielashi:
Note that the bitcoin interface in the replica has been updated to be consistent with the current spec, so the changes proposed to the Bitcoin interface in this PR can be reverted.

PR #76, opened 2022-08-09
Cut release 0.18.6

Dfinity-Bjoern:
Main changes:

* Canister access to performance metrics
* Query calls are rejected when the canister is frozen
* Support for implementation-specific error codes for requests
* Deleted call contexts do not prevent canister from reaching Stopped state
* Update effective canister id checks in certificate delegations
* Formal model in Isabelle

And a few clarifications and corrections.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -543,7 +543,10 @@ The HTTP response to this request consists of a CBOR map with the following fiel
 
 * `certificate` (`blob`): A certificate (see <<certification>>).
 +
-If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation’s canister id range (see <<certification-delegation>>), unless the `effective_canister_id` is that of the Management Canister (`aaaaa-aa`). In other words, we define `aaaaa-aa` to belong to every canister range for the purposes of certificate checking. Note that `aaaaa-aa` is a valid canister id only for calls to `provisional_create_canister_with_cycles`, so it cannot appear in `read_state` requests on production networks.
+If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation’s canister id range (see <<certification-delegation>>), unless
+- the `effective_canister_id` is that of the Management Canister (`aaaaa-aa`),
+- all requested paths have `/request_status/<request_id>` as prefix where the (single) original request referenced by `<request_id>` is an update call to the Management Canister (`aaaaa-aa`) and the method name is `provisional_create_canister_with_cycles`, and
+- whenever the certificate contains the path `/request_status/<request_id>/reply`, then its value is a Candid-encoded record with a `canister_id` field of type `principal` and the `canister_id` must be included in each delegation’s canister id range.
 
 The returned certificate reveals all values whose path is a suffix of the list of requested paths. It also always reveals `/time`, even if not explicitly requested.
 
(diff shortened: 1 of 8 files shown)


Dfinity-Bjoern:
@nomeata I cannot request a review from you but I'd still be happy to get your thoughts :)

spec/index.adoc
```diff
@@ -543,7 +543,7 @@ The HTTP response to this request consists of a CBOR map with the following fiel
 
 * `certificate` (`blob`): A certificate (see <<certification>>).
 +
-If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation’s canister id range (see <<certification-delegation>>).
+If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation’s canister id range (see <<certification-delegation>>), unless the `effective_canister_id` is that of the Management Canister (`aaaaa-aa`). In other words, we define `aaaaa-aa` to belong to every canister range for the purposes of certificate checking. Note that `aaaaa-aa` is a valid canister id only for calls to `provisional_create_canister_with_cycles`, so it cannot appear in `read_state` requests on production networks.
```
mraszyk, line 546: I don't quite the sentence starting with "Note ...". How about replacing it with "Note that `aaaaa-aa` does not satisfy the predicate `is_system_assigned` and thus a canister with this canister id can only be created by a call to `provisional_create_canister_with_cycles`, so no canister with this canister id can be created on production networks"?

Dfinity-Bjoern:
```suggestion
If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation’s canister id range (see <<certification-delegation>>), unless the `effective_canister_id` is that of the Management Canister (`aaaaa-aa`). In other words, we define `aaaaa-aa` to belong to every canister range for the purposes of certificate checking. Note that `aaaaa-aa` is a valid effective canister id only for calls to `provisional_create_canister_with_cycles`, so it cannot appear in `read_state` requests on production networks.
```

Dfinity-Bjoern: Is it clearer like that? The note only applies to calls that can be made through the HTTP interface, and `create_canister` cannot be called through that interface (same for `raw_rand`). The only call that can be made through that interface and that has `aaaaa-aa` as `effective_canister_id` is `provisional_create_canister_with_cycles`, hence the comment. Does that make sense?

mraszyk: It still does not quite make sense given the current spec:

If we talk about *calls* via HTTP interface (i.e., requests to `/api/v2/canister/<ECID>/call`), then `aaaaa-aa` (a.k.a. ic_principal) is a valid effective canister id for all calls to ic_principal that set `A.canister_id=ic_principal` in their args:
```
is_effective_canister_id(Request {canister_id = ic_principal, arg = candid({canister_id = p, …}), …}, p)
```
If `is_system_assigned ic_principal` does not hold (which is *under-specified* at the moment), then ic_principal cannot exist as a canister (both in production and in local deployment) and every such call is indeed rejected (the transition "API Request submission" cannot be taken). Otherwise, it can be a legal request that is accepted.

If we talk about *read_state* requests (i.e., requests to `/api/v2/canister/<ECID>/read_state`), then:

If `aaaaa-aa` is an instance of `CanisterID` (which is *under-specified* at the moment), then I don't see what would prevent me from making a read_state request to `/api/v2/canister/aaaaa-aa/read_state` as an anonymous principal asking for the path `["canister", "aaaaa-aa", "module_hash"]`.

Otherwise, `verify_envelope` would never return a set containing `aaaaa-aa` and thus every read_state request to `/api/v2/canister/aaaaa-aa/read_state` would be invalid (both in production and in local deployment).

mraszyk:
Could you please pick up the commit "Deleted call contexts do not prevent canister from reaching Stopped state" into the release?

mraszyk:
and #84 ?

spec/index.adoc
mraszyk: Resolved by #82 .

mraszyk (approved):

PR #77, opened 2022-08-09
Include all abstract state transitions in the formal model

mraszyk:
This PR extends the formal model of Interface Spec with

- all abstract state transitions
- complete proof on amounts of minted and burned cycles during the state transitions
- fix and prove invariants
- sync textual spec with formal model


nomeata:
Did I mention that this is great work? If not, this is great work :-)

Looking forward to a nicely polished document generation setup that will fully replace the pseudo-code section in the spec :-)

Dfinity-Bjoern (approved):
Textual spec changes LGTM

mraszyk:
An invariant violation before commit `064eae6` could look as follows. Intuitively, the steps correspond to a user call of a canister public method that calls the IC Management canister to stop itself (asynchronously without an await) and responds immediately. This may result in the call context of our canister being removed and the canister being stopped. Once the canister is stopped, the system generates a reply to the caller of the stopping request (our canister in this case). Finally, the callback invocation upon the reply needs to check if the context is not marked deleted, i.e., the corresponding field set (this would have happened, e.g., if we uninstalled our canister, in which case no callback would be executed), but the call context does not exist anymore (violation of the invariant "referenced call contexts must exist").

```
1. API Request submission
  - canister_status[R.canister_id] = Running
  - requests[R] = Received
2. Initiating canister calls
  - canister_status[R.canister_id] = Running
  - requests[R] = Processing
  - CallMessage {
      origin = FromUser {
        request = R;
      };
      caller = R.sender;
      callee = R.canister_id;
      method_name = R.method_name;
      arg = R.arg;
      transferred_cycles = 0;
      queue = Unordered;
    }
3. Call context creation: Public entry points
  - canister_status[R.canister_id] = Running
  - requests[R] = Processing
  - call_contexts[Ctxt_id] = {
      canister = R.canister_id;
      origin = FromUser {
        request = R;
      };
      needs_to_respond = true;
      deleted = false;
      available_cycles = 0;
    }
  - FuncMessage {
      call_ctxt = Ctxt_id;
      receiver = R.canister_id;
      entry_point = PublicMethod R.method_name R.sender R.arg;
      queue = Unordered;
    }
4. Message execution
  - canister_status[R.canister_id] = Running
  - requests[R] = Processing
  - call_contexts[Ctxt_id] = {
      canister = R.canister_id;
      origin = FromUser {
        request = R;
      };
      needs_to_respond = false;
      deleted = false;
      available_cycles = 0;
    }
  - CallMessage {
      origin = FromCanister {
        call_context = Ctxt_id;
        callback = call.callback;
      };
      caller = R.canister_id;
      callee = ic_principal;
      method_name = 'stop_canister';
      arg = '({canister_id: R.canister_id})';
      transferred_cycles = 0;
      queue = Queue {
        from = R.canister_id;
        to = ic_principal;
      };
    }
  - ResponseMessage {
      origin = FromUser {
        request = R;
      };
      response = res.response;
      refunded_cycles = 0;
    }
5. Stopping a canister
  - canister_status[R.canister_id] = Stopping [(FromCanister {call_context = Ctxt_id; callback = call.callback}, 0)]
  - requests[R] = Processing
  - call_contexts[Ctxt_id] = {
      canister = R.canister_id;
      origin = FromUser {
        request = R;
      };
      needs_to_respond = false;
      deleted = false;
      available_cycles = 0;
    }
  - ResponseMessage {
      origin = FromUser {
        request = R;
      };
      response = res.response;
      refunded_cycles = 0;
    }
6. Call context removal
  - canister_status[R.canister_id] = Stopping [(FromCanister {call_context = Ctxt_id; callback = call.callback;}, 0)]
  - requests[R] = Processing
  - ResponseMessage {
      origin = FromUser {
        request = R;
      };
      response = res.response;
      refunded_cycles = 0;
    }
7. Stopping a canister
  - canister_status[R.canister_id] = Stopped
  - requests[R] = Processing
  - ResponseMessage {
      origin = FromUser {
        request = R;
      };
      response = res.response;
      refunded_cycles = 0;
    }
  - ResponseMessage {
      origin = FromCanister {
        call_context = Ctxt_id;
        callback = call.callback;
      };
      response = Reply ();
      refunded_cycles = 0;
    }
8. Callback invocation
  - check not call_contexts[Ctxt_id].deleted: IC traps on missing key Ctxt_id in call_contexts!
```

Dfinity-Bjoern:
Thanks @mraszyk! Do I understand correctly that with your changes, the call context would not be removed at that point but only when the response from the management canister is in?

Do we know how the replica handles this case?

mraszyk:
With my changes, the call context will not be removed and thus the canister cannot get stopped (and thus the call context cannot be removed...). I could also confirm that the replica deadlocks in this case, too (and the call contexts exists because upgrading the canister fails with an error message relating to an open call context). The deadlock can, e.g., be broken by restarting the canister.

mraszyk:
The failed upgrade error message is produced by this piece of code in the Motoko compiler:
```
      (* check status is stopped or trap on outstanding callbacks *)
      system_call env "canister_status" ^^ compile_eq_const status_stopped ^^
      G.if0
       (G.nop)
       (ContinuationTable.count env ^^
          E.then_trap_with env "canister_pre_upgrade attempted with outstanding message callbacks (try stopping the canister before upgrade)") ^^
```
So it only seems to check for the canister status, not for the existence of open call contexts. But I still suppose that the system will take the transition from Stopping to Stopped eagerly once there are no more open call contexts left.

ulan:
> Do we know how the replica handles this case?

I think the replica should handle this case correctly:
1. We transition a canister from Stopping to Stopped only if it doesn't have any call contexts: https://github.com/dfinity/ic/blob/f20bb61039a36b9ddbbe09ede420f2b1f487c35f/rs/replicated_state/src/canister_state/system_state.rs#L649
2. We remove a call context only if it doesn't have any outstanding calls: https://github.com/dfinity/ic/blob/master/rs/replicated_state/src/canister_state/system_state/call_context_manager.rs#L425

Issue #80, opened 2022-08-19
Undefined field in `Code installation/upgrade` transitions


ulan:
`IC Management Canister: Code installation` and `IC Management Canister: Code upgrade` refer to `M.receiver`, but `M` doesn't have that field because `M` is `CallMessage` not `FuncMessage`.

I think the fix would be to replace `M.receiver` with `A.canister_id`

Links to the transitions:
- https://ic-interface-spec.netlify.app/#_ic_management_canister_code_installation
- https://ic-interface-spec.netlify.app/#ic_management_canister_code_upgrade

nomeata:
BTW, the above URL is often behind. You can link to the master version here

https://ic-interface-spec.netlify.app/#_ic_management_canister_code_installation

(At least I believe the netlify deploy is still up to date.)

ulan:
This will be fixed by https://github.com/dfinity/interface-spec/pull/77

mraszyk:
I'm closing this issue as it has been fixed in #77.

PR #81, opened 2022-08-19
Fix the canister id  in install/upgrade transitions

ulan:
This replaces an incorrect `M.receiver` with `A.canister_id` in
the code installation and upgrade transitions.

This closes Issue #80

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -3186,11 +3186,11 @@ Conditions::
     or A.mode = reinstall
     M.caller ∈ S.controllers[A.canister_id]
     Env = {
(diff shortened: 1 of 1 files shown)


mraszyk:
The fix is also contained in https://github.com/dfinity/interface-spec/pull/77

mraszyk (commented):
The changes look good to me, but they're also contained in another PR, so it might not be necessary to merge this one separately.

ulan:
Nice. I didn't see the other PR. Abandoning in favor of https://github.com/dfinity/interface-spec/pull/77

PR #82, opened 2022-08-23
update spec of ECID

mraszyk:
Update effective canister ID specification.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -608,9 +608,7 @@ NOTE: The expectation is that user-side agent code shields users and developers 
 ====
 The Internet Computer blockchain mainnet does not support `provisional_create_canister_with_cycles`. This means that using an effective canister id from a subnet's canister range leads to the request being routed to a node on the corresponding subnet and produce an error response there, while using an effective canister id that is in no subnet's canister ranges would cause an error response from the node receiving the request.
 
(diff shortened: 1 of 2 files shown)


spec/index.adoc
```diff
@@ -594,26 +594,25 @@ Canister methods that do not change the canister state can be executed more effi
...
+
+NOTE: The expectation is that user-side agent code shields users and developers from the notion of an effective canister ID, in analogy to how the System API interface shields canister developers from worrying about routing.
 
 +
 [NOTE]
 ====
-The Internet Computer blockchain mainnet does not support `provisional_create_canister_with_cycles`. This means that using an effective canister id that could be an existing canister would lead to the request being routed to a node on the corresponding subnet and produce an error response there, while using an effective canister id that cannot be an existing canister id would cause an error response from the node receiving the request -- either is fine.
+The Internet Computer blockchain mainnet does not support `provisional_create_canister_with_cycles`. This means that using an effective canister id from a subnet's canister range leads to the request being routed to a node on the corresponding subnet and produce an error response there, while using an effective canister id that is in no subnet's canister ranges would cause an error response from the node receiving the request.
```
mraszyk, line 609: "from the node receiving the request." -> which subnet is the request routed to in this case?

mraszyk: I believe that the boundary node responses with 404 because it cannot route the request.

mraszyk: Confirmed by @r-birkner

Dfinity-Bjoern (approved):
Thanks for the clarifications!

PR #83, opened 2022-08-23
Deleted call contexts do not prevent canister from reaching Stopped state

mraszyk:
Currently, a canister with open call contexts cannot move to a stopping state even if all its open call contexts are marked as deleted (e.g., after its code was uninstalled). Although it's harmless to have an empty canister with open call contexts that are all marked deleted, if somebody installed a new canister module and later wanted to upgrade it, the open call contexts that are marked deleted could prevent the transition to the stopped state, making the canister dangerous to upgrade.

To resolve the issue described in the previous paragraph, I propose to update the definition of when a canister is ready to be stopped to only consider the existence of open call contexts that are not marked deleted.

The replica MR is linked here: https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/7140

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -223,7 +223,7 @@ The canister status can be used to control whether the canister is processing ca
 
 * In status `running`, calls to the canister are processed as normal.
 * In status `stopping`, calls to the canister are rejected by the IC, but responses to the canister are processed as normal.
-* In status `stopped`, calls to the canister are rejected by the IC, and there are no outstanding responses.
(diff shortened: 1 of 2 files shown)


nomeata (approved):
Thanks! That was certainly my intention - after all, a deleted call context is no longer there, in a sense :-)

ielashi (approved):

mraszyk:
The replica MR has been merged.

bogwar (approved):

PR #88, opened 2022-09-07
feat: a spec for reserved principals

roman-kashitsyn:
This change defines a way for apps to indicate that a blob is _not_ a valid principal.
This feature is particularly useful when an app wants to re-use the textual encoding for principals but does not want to confuse the system with blobs that _accidentally_ look like a valid principal (see https://github.com/dfinity/ICRC-1/pull/55 for an example).

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -141,7 +141,7 @@ NOTE: Derived IDs are currently not explicitly used in this document, but they m
 +
 This has the form `0x04`, and is used for the anonymous caller. It can be used in call and query requests without a signature.
 
-5. _Invalid ids_
+5. _Reserved ids_
 +
 These have the form of `blob · 0xff`, with no restriction on the blob length.
 +
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -143,11 +143,11 @@ This has the form `0x04`, and is used for the anonymous caller. It can be used i
 
 5. _Reserved ids_
 +
-These have the form of `blob · 0xff`, with no restriction on the blob length.
+These have the form of `blob · 0xff`, `0 ≤ |blob| < 29`.
 +
 These ids can be useful for applications that want to re-use the <<textual-ids>> but want to indicate explicitly that the blob does not address any canisters or a user.
 
-When the IC creates a _fresh_ id, it never creates a self-authenticating id, an anonymous id or an id derived from what could be a canister or user.
+When the IC creates a _fresh_ id, it never creates a self-authenticating id, reserved id, an anonymous id or an id derived from what could be a canister or user.
 
 [#textual-ids]
 ==== Textual representation of principals
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -141,8 +141,13 @@ NOTE: Derived IDs are currently not explicitly used in this document, but they m
 +
 This has the form `0x04`, and is used for the anonymous caller. It can be used in call and query requests without a signature.
 
-When the IC creates a _fresh_ id, it never creates a self-authenticating id, an anonymous id or an id derived from what could be a canister or user.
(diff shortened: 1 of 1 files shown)


nomeata (commented):
Are they really ”invalid”? They are valid, in a way, just guaranteed to never play a particular role in this interface specification. So probably there is a better name. “reserved”? “custom”?

roman-kashitsyn:
> Are they really ”invalid”? They are valid, in a way, just guaranteed to never play a particular role in this interface specification. So probably there is a better name. “reserved”? “custom”?

I like “reserved”; this variant also crossed my mind. I hesitated a bit because I was not sure what to do in the Rust decoder:
https://github.com/dfinity/ic-types/blob/69632a79d76c09f1d2a33df988189b8d5a4015ca/src/principal.rs#L12-L13

If I name these "Invalid", I can return `Err(InvalidId(blob))`. If they are valid but “reserved”, I need to change the internal representation to be cleverer and handle arbitrary-sized arrays. I’m pretty sure I can make it work, and this will be consistent with Motoko’s behavior, where we allow arbitrary blobs as principals.

spec/index.adoc
```diff
@@ -141,8 +141,13 @@ NOTE: Derived IDs are currently not explicitly used in this document, but they m
...
 This has the form `0x04`, and is used for the anonymous caller. It can be used in call and query requests without a signature.
 
-When the IC creates a _fresh_ id, it never creates a self-authenticating id, an anonymous id or an id derived from what could be a canister or user.
+5. _Reserved ids_
++
+These have the form of `blob · 0xff`, with no restriction on the blob length.
++
+These ids can be useful for applications that want to re-use the <<textual-ids>> but want to indicate explicitly that the blob does not address any canisters or a user.
 
+When the IC creates a _fresh_ id, it never creates a self-authenticating id, an anonymous id or an id derived from what could be a canister or user.
```
mraszyk, line 150: Shouldn't the reserved IDs be mentioned in this sentence?

spec/index.adoc
```diff
@@ -141,8 +141,13 @@ NOTE: Derived IDs are currently not explicitly used in this document, but they m
 +
 This has the form `0x04`, and is used for the anonymous caller. It can be used in call and query requests without a signature.
 
-When the IC creates a _fresh_ id, it never creates a self-authenticating id, an anonymous id or an id derived from what could be a canister or user.
+5. _Reserved ids_
++
+These have the form of `blob · 0xff`, with no restriction on the blob length.
```
mraszyk, line 146: "no restriction": there is still the global restriction that any principal id is at most 29 bytes, right? That means that the blob is at most 28 bytes long.

nomeata:
Oh, you want to allow them to be larger than 32 bytes? That seem dangerous - either they are principals (even if of non-system resources), with all the rules around them, or they are not, and then one shouldn't make them look like some.

spec/index.adoc
roman-kashitsyn: Good point, done.

roman-kashitsyn:
> Oh, you want to allow them to be larger than 28 bytes?

After thinking more about it, keeping the 29-byte limit works fine for our use case. Our primary concern is constructing something that looks like a principal but isn't a principal.

If we construct a blob that is longer than 29 bytes, it's definitely _not_ a principal by definition.
If we construct a blob shorter than 29 bytes, the 0xff marker will differentiate it from all other principals.

nomeata:
I’d say, what you are constructing _is_ a principal, and we just need a mechanism to prevent clashes. But this is getting philosophical :-)

Dfinity-Bjoern (approved):
As discussed, thanks @roman-kashitsyn, @nomeata and @mraszyk for the refinement!

PR #89, opened 2022-09-07
Bitcoin fees are measured in sat/vbyte

THLO:
In Bitcoin, fees are measured in `sat/vbyte` (as opposed to `sat/byte`). This PR makes it clear that our IC method `bitcoin_get_current_fee_percentiles` is consistent with standard Bitcoin clients and returns fee percentiles in `sat/vbyte` as well.

The concept of virtual bytes (`vbytes`) and weight units is specified [here](https://en.bitcoin.it/wiki/Weight_units).

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1801,7 +1801,7 @@ The transaction fees in the Bitcoin network change dynamically based on the numb
 pending transactions.
 It must be possible for a canister to determine an adequate fee when creating a Bitcoin transaction.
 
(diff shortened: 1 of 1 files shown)


Dfinity-Bjoern (approved):
Can you add a link to a specification (like the one you have in the PR description) to the spec before you merge?

THLO:
> Can you add a link to a specification (like the one you have in the PR description) to the spec before you merge?

I thought about adding a link but decided against it because there is already a link to the [Bitcoin developer guides](https://developer.bitcoin.org/devguide/) at the beginning of the Bitcoin integration section.
There are many Bitcoin-related concepts that we don't explain here (for good reason), so I thought it would be inconsistent to add a link here.

If you feel that a link would be appropriate here, I can definitely add one. What do you think?

PR #90, opened 2022-09-10
chore: add a case to find_label

mraszyk:
Add a (forgotten) case to the definition of `find_label`:
```
find_label(l, [Leaf _]) = Absent
```

Note that `well_formed(tree)` implies that either
- `tree` is `Leaf v`: then `flatten_forks(tree) = [Leaf v]` and the new case from this PR triggers (before `find_label` would return `Unknown` which is unnecessarily conservative);
- otherwise, `flatten_forks(tree)` cannot contain any tree of the form `Leaf v`; hence we don't need to care about leaves in `find_label` beyond the case `[Leaf v]`.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1916,6 +1916,7 @@ find_label(l, _ · Labeled l1 t · _)                | l == l1     = Found t
 find_label(l, _ · Labeled l1 _ · Labeled l2 _ · _) | l1 < l < l2 = Absent
 find_label(l,                    Labeled l2 _ · _) |      l < l2 = Absent
 find_label(l, _ · Labeled l1 _ )                   | l1 < l      = Absent
+find_label(l, [Leaf _])                                          = Absent
 find_label(l, [])                                                = Absent
 find_label(l, _)                                                 = Unknown
 ....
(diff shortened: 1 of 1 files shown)


roman-kashitsyn (approved):

nomeata (commented):
Hmm, I wonder if I left that out explicitly: If you are looking up a label, and there is a leaf, then something is wrong™, in the sense that the (implied) schemata don't line up. That's somewhat different from ”query makes sense, but value is not there”.

I would not consider an implementation that aborts in that case as wrong. But I would consider it bad behavior from the replica (or a canister using the same hash trees) to out a leaf into a spot where it should prove absence of all labels.

So maybe instead add a comment explaining this?

mraszyk:
I don't see why a leaf should not prove the absence of a longer path, e.g., why a leaf with a canister module hash at the path `/canister/<canister_id>/module_hash` should not prove the absence of the path `/canister/<canister_id>/module_hash/foo`.

In my understanding, the value `Unknown` should never be returned for trees that are not pruned.

nomeata:
Fair point, it's probably fine.

Dfinity-Bjoern (approved):
I agree with the outcome of the discussion.

PR #91, opened 2022-09-15
allow a canister to query its own status

bogwar:
Currently only controllers can request `canister_status` from the management canister. This PR changes the interface spec to make this query also available to the canister itself.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1517,7 +1517,7 @@ Default value: 2592000 (approximately 30 days).
 [#ic-update_settings]
 === IC method `update_settings`
 
-Only _controllers_ of the canister can update settings. See <<ic-create_canister>> for a description of settings.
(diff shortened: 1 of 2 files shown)


nomeata (approved):
https://ic-interface-spec.netlify.app/#_ic_management_canister_canister_status needs to be updated.

Otherwise LGTM.

Probably some applications would benefit from a synchronous interface to some of the information in the canister status.

bogwar:
> https://ic-interface-spec.netlify.app/#_ic_management_canister_canister_status needs to be updated.

I've added the fix, thanks for the quick review!

> 
> Otherwise LGTM.
> 
> Probably some applications would benefit from a synchronous interface to some of the information in the canister status.

yeah, this is under discussion but requires bigger changes so for now we'll start with this smaller but still useful change.

Dfinity-Bjoern (approved):
Looks good!

PR #94, opened 2022-09-22
allow subnet id to be effective principal id

mraszyk:
To avoid the special case of update calls to the Management Canister (`aaaaa-aa`) with the method name `provisional_create_canister_with_cycles` and the effective canister id equal to that of the Management Canister (`aaaaa-aa`), I propose to change the specification and implementation in several steps as follows:

1. the specification is relaxed (this PR):
- rename effective *canister* id to effective *principal* id
- effective principal id can be equal to subnet id for requests submitted to nodes in testnets
- effective principal id can be equal to subnet id in certificate delegation checking

2. checks in `agent-rs` are relaxed according to the relaxed specification

3. IC testnet (and potentially also boundary node) implementation is incrementally updated

4. the specification is tightened (future PR) to avoid the special case

5. checks in `agent-rs` are tightened according to the tightened specification

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -550,7 +550,7 @@ The HTTP response to this request consists of a CBOR map with the following fiel
 
 * `certificate` (`blob`): A certificate (see <<certification>>).
 +
-If this `certificate` includes subnet delegations (possibly nested), then the `effective_principal_id` must be included in each delegation’s canister id range (see <<certification-delegation>>) or equal to the delegation’s subnet, unless
+If this `certificate` includes subnet delegations (possibly nested), then the `effective_principal_id` must be included in each delegation’s canister id range (see <<certification-delegation>>) or equal to the delegation’s subnet id, unless
 
 - the `effective_principal_id` is that of the Management Canister (`aaaaa-aa`),
 - all requested paths have `/time` or `/request_status/<request_id>` as prefix where the (single) original request referenced by `<request_id>` is an update call to the Management Canister (`aaaaa-aa`) and the method name is `provisional_create_canister_with_cycles`, and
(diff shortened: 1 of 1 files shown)


Dfinity-Bjoern:
I am very reluctant to do this. The main reason is that `provisional_canister_create_with_cycles` should not be in the main specification anyway. It is a feature that is not and should never be available in mainnet. Morally, we should remove it from the spec rather than have this artifact bleed into all parts of the spec.

mraszyk:
> I am very reluctant to do this. The main reason is that `provisional_canister_create_with_cycles` should not be in the main specification anyway. It is a feature that is not and should never be available in mainnet. Morally, we should remove it from the spec rather than have this artifact bleed into all parts of the spec.

I think the provisional API is necessary to bootstrap canisters after creating a new IC so it must fundamentally be part of the spec. It must have been necessary after Genesis, too, before it was disabled unless the mainnet started off with some canisters hard-coded in the state.

mraszyk:
> unless the mainnet started off with some canisters hard-coded in the state

in which case we should change the definition of the initial state of the IC because it wouldn't be possible to create any cycles from the current initial state

mraszyk:
Moreover, allowing subnet id to be effective principal id for read state requests would allow to query subnet information (i.e., the path `/subnet/`) without providing a (potentially) non-existing canister id from the canister range of a subnet to which such a request is targetted.

mraszyk:
Not needed anymore to avoid the special case of update calls to the Management Canister (`aaaaa-aa`) with the method name `provisional_create_canister_with_cycles` and the effective canister id equal to that of the Management Canister (`aaaaa-aa`).

PR #95, opened 2022-09-26
add a remark on utf-8 encoding of custom section names

mraszyk:
The additional constraint is implemented in ic-ref and on the mainnet and tested in ic-ref-test so it is added for documentation purposes.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -879,6 +879,7 @@ In order for a WebAssembly module to be usable as the code for the canister, it 
 * If it exports any functions called `canister_update <name>` or `canister_query <name>` for some `name`, the functions must have type `+() -> ()+`.
 * It may not export both `canister_update <name>` and `canister_query <name>` with the same `name`.
 * It may not export other methods the names of which start with the prefix `canister_` besides the methods allowed above.
+* Every custom section name of the form `icp:public <name>` or `icp:private <name>` must be encoded in UTF-8 (the actual content of the custom section need not be encoded in UTF-8).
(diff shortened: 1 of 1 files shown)


spec/index.adoc
```diff
@@ -879,6 +879,7 @@ In order for a WebAssembly module to be usable as the code for the canister, it
 * If it exports any functions called `canister_update <name>` or `canister_query <name>` for some `name`, the functions must have type `+() -> ()+`.
 * It may not export both `canister_update <name>` and `canister_query <name>` with the same `name`.
 * It may not export other methods the names of which start with the prefix `canister_` besides the methods allowed above.
+* Every custom section name of the form `icp:public <name>` or `icp:private <name>` must be encoded in UTF-8 (the actual content of the custom section need not be encoded in UTF-8).
```
roman-kashitsyn, line 882: I think using WebAssembly section already implies that section names are UTF-8 encoded:

> All [names](https://webassembly.github.io/spec/core/binary/values.html#binary-name) are represented in [Unicode](https://www.unicode.org/versions/latest/) encoded in UTF-8. Names need not be unique.
-- https://webassembly.github.io/spec/core/appendix/custom.html

mraszyk: Your link refers to a particular custom section whose name is `name`. The generic custom sections are defined [here](https://webassembly.github.io/spec/core/binary/modules.html#binary-customsec). The name of a custom section is specified [here](https://webassembly.github.io/spec/core/syntax/values.html#syntax-name) and it must indeed be UTF-8 encoded, so I'm closing this PR.

PR #96, opened 2022-09-26
http_request: Add transform context

yotamhc:
It has been decided to add a blob that represents any necessary context for the transform function, as an argument for the http_request method, and for the transform function.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -115,9 +115,9 @@ service ic : {
     method : variant { get; head; post };
     headers: vec http_header;
     body : opt blob;
-    transform_context: blob;
-    transform : opt variant {
-      function: func (http_response, blob) -> (http_response) query
+    transform : opt record {
+      function: func (record { response: http_response; context: blob }) -> (http_response) query;
+      context: blob
     };
   }) -> (http_response);
 
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -115,8 +115,9 @@ service ic : {
     method : variant { get; head; post };
     headers: vec http_header;
     body : opt blob;
+    transform_context: blob;
     transform : opt variant {
-      function: func (http_response) -> (http_response) query
+      function: func (http_response, blob) -> (http_response) query
     };
   }) -> (http_response);
 
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1683,6 +1683,8 @@ The returned response (and the response provided to the `transform` function, if
 The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
 When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
 
+NOTE: The `transform` function is invoked by the management canister (`aaaaa-aa`). This information can be used by developers to implement access control mechanism for this function.
+
 [#ic-provisional_create_canister_with_cycles]
 === IC method `provisional_create_canister_with_cycles`
 
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1672,8 +1672,7 @@ The following parameters should be supplied for the call:
 - `method` - currently, only GET, HEAD, and POST are supported
 - `headers` - list of HTTP request headers and their corresponding values
 - `body` - optional, the content of the request's body
-- `transform_context` - a byte-encoded context that is provided to the `transform` function, along with the response
-- `transform` - an optional function that transforms raw responses to sanitized responses. If provided, the calling canister itself must export this function.
+- `transform` - an optional record that includes a function that transforms raw responses to sanitized responses, and a byte-encoded context that is provided to the function upon invocation, along with the response to be sanitized. If provided, the calling canister itself must export this function.
 
 The returned response (and the response provided to the `transform` function, if specified) contains the following fields:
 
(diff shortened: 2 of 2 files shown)


Dfinity-Bjoern:
Is there a size restriction, or does the canister pay for the space used by the context blob?

yotamhc:
> Is there a size restriction, or does the canister pay for the space used by the context blob?

The canister pays, the size of this blob is added to the size of request, which is one parameter of the cost function. Should this be specified in the spec? So far we did not mention pricing here.

nomeata:
If only we had proper closures ;-)

Dfinity-Bjoern (approved):

spec/ic.did
```diff
@@ -115,8 +115,9 @@ service ic : {
     method : variant { get; head; post };
     headers: vec http_header;
     body : opt blob;
+    transform_context: blob;
     transform : opt variant {
-      function: func (http_response) -> (http_response) query
+      function: func (http_response, blob) -> (http_response) query
```
crusso, line 120: For consistency, should the two argument sequence `(http_repsonse, blob)`  be packed into a record `(response: http_response; context: blob)`? 

This is legal Candid as is, but everywhere else in the IC management canister interface uses a single record of arguments and results, never multiple arguments.

crusso:
```suggestion
      context : blob;
      function: func { response : http_response; context:  blob} -> (http_response) query
```
Since the transform itself is optional, perhaps the context should only be specified as part of the variant. Why require a context when no transform is provided? This way, you supply neither of both.

(and delete transform_context at l. 118)

spec/ic.did
```diff
@@ -115,8 +115,9 @@ service ic : {
     method : variant { get; head; post };
     headers: vec http_header;
     body : opt blob;
+    transform_context: blob;
     transform : opt variant {
```
crusso, line 119:
```suggestion
      transform : opt record { context : blob;
                                             function: func (record { response : http_response; context:  blob}) -> (http_response) query }
```
Since the transform itself is optional, perhaps the context should only be specified as part of the variant. Why require a context when no transform is provided? This way, you supply neither of both.

(and delete `transform_context` at l. 118)

nomeata: Good point!

spec/ic.did
```diff
@@ -115,8 +115,9 @@ service ic : {
     method : variant { get; head; post };
     headers: vec http_header;
     body : opt blob;
-    transform : opt variant {
-      function: func (http_response) -> (http_response) query
+    transform : opt record {
+      function: func (record { response: http_response; context: blob }) -> (http_response) query;
+      context: blob
```
crusso, line 120:
```suggestion
    transform : opt record {
      function : func (record {response : http_response; context : blob}) -> (http_response) query;
      context : blob
```
(adjust spacing)

crusso (commented):
LGTM

Does the spec mention what the caller of the transform function will be once invoked? I suspect an implementation will want to be able to do access control checks on the caller to prevent DOS attacks.

crusso:
> If only we had proper closures ;-)

My recollection is that you and Andreas started a design for this in Candid but never reached a satisfactory conclusion, right?

crusso:
Hmm, do we ever envision changing the API the transform function? If so, we might want to add a variant around the record

yotamhc:
> Does the spec mention what the caller of the transform function will be once invoked? I suspect an implementation will want to be able to do access control checks on the caller to prevent DOS attacks.

Currently it does not. Should it? (cc @Dfinity-Bjoern )

Dfinity-Bjoern:
> > Does the spec mention what the caller of the transform function will be once invoked? I suspect an implementation will want to be able to do access control checks on the caller to prevent DOS attacks.
> 
> Currently it does not. Should it? (cc @Dfinity-Bjoern )

I think it makes sense, yes. And I suppose it's `aaaaa-aa` in the implementation?

spec/index.adoc
```diff
@@ -1682,6 +1683,8 @@ The returned response (and the response provided to the `transform` function, if
 The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
 When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
 
+NOTE: The `transform` function is invoked by the management canister (`aaaaa-aa`). This information can be used by developers to implement access control mechanism for this function.
```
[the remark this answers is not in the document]
yotamhc: Thanks! I pushed a fix, please re-read.

spec/index.adoc
```diff
@@ -1682,6 +1683,8 @@ The returned response (and the response provided to the `transform` function, if
 The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
 When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
 
+NOTE: The `caller_id` for a valid invocation of the `transform` function is `aaaaa-aa`. This information can be used by developers to implement access control mechanism for this function.
```
AlexandraZapuc, line 1686: This is consistent with the spec.
 
 ```suggestion
NOTE: The identity of the caller for a valid invocation of the `transform` function is `aaaaa-aa`. This information can be used by developers to implement access control mechanism for this function.
```

AlexandraZapuc (approved):

PR #101, opened 2022-10-12
refine specification of access control for read state requests

mraszyk:


nomeata (approved):
LGTM

spec/index.adoc
```diff
@@ -556,18 +556,35 @@ If this `certificate` includes subnet delegations (possibly nested), then the `e
...
- * `/canisters/<canister_id>/controllers`. Can be requested by anyone if `<canister_id>` matches `<effective_canister_id>`. The order may vary depending on the implementation.
- * `/canisters/<canister_id>/metadata/<name>`. Can be requested by anyone if `<canister_id>` matches `<effective_canister_id>` and `<name>` is a public custom section. If `<name>` is a private custom section, it can only be requested by the controllers of the canister.
+ * `/time`. Can always be requested.
+ * `/subnet`. Can always be requested.
+ * `/request_status/<request_id>`. Can be requested if no path with such a prefix exists in the state tree or
+   - the sender of the original request referenced by `<request_id>` is the same as the sender of the read state request and
+   - `<effective_canister_id>` is a valid effective canister id for the original request referenced by `<request_id>` and
+   - the effective canister id of the original request referenced by `<request_id>` and the effective canister id `<effective_canister_id>` of the read state request belong to canister ranges of the same subnet.
```
Dfinity-Bjoern, line 568: The condition "on the same subnet" seems at odds with the general approach of hiding the subnets where possible. Does it maybe make sense to _guarantee_ correct output if the `request_id` and `effective_request_id` are the same and not guarantee behavior otherwise?

mraszyk: Now that I looked into the Rust implementation of this check (it doesn't seem possible to get the original message arguments from the ingress history which would be needed to check the second condition above), we might need to store the effective canister id of the original request in the ingress history and thus the original formulation (effective canister id of the original request must match the effective canister id of this read state request) might actually work.

mraszyk: We might actually take the original condition (the effective canister id of the original request matches the effective canister id of this read state request) if we store the effective canister id of the original request in the ingress history.

Dfinity-Bjoern (approved):

PR #103, opened 2022-10-15
canister history

mraszyk:
This change introduces canister history implemented in the ticket [[EXC-1180]](https://dfinity.atlassian.net/browse/EXC-1180).


spec/history.cddl
```diff
@@ -0,0 +1,58 @@
...
+  code-deployment /
+  controllers-change
+
+canister-creation = 0
+code-uninstall = 1
+
+code-deployment = {
+  mode : code-deployment-mode
+  module_hash : bytes
+}
```
nomeata, line 37: This looks strangly non-uniform. Some fields are numbers, some are records with a `mode`.

Moreover, it’s not really a `can-change-type` (NB: type) if it also has additional data! Why not `can-change`?

Or, maybe more idiomatic, one could simply inline the fields into the `can-change` type.

spec/history.cddl
```diff
@@ -0,0 +1,58 @@
...
+
+code-deployment-mode =
+  install /
+  reinstall /
+  upgrade
+
+install = 0
+reinstall = 1
+upgrade = 2
```
nomeata, line 46: Using strings is too expensive? We use strings otherwise in our CDDLs, do we?

spec/index.adoc
```diff
@@ -443,6 +443,19 @@ If the canister has a https://webassembly.github.io/spec/core/binary/modules.htm
 +
 It is recommended for the canister to have a custom section called "icp:public candid:service", which contains the UTF-8 encoding of https://github.com/dfinity/candid/blob/master/spec/Candid.md#core-grammar[the Candid interface] for the canister.
 
+* `/canister/<canister_id>/history` (blob):
++
+The history of the canister. The value consists of a CBOR value encoding the canister history according to the following CDDL (the https://tools.ietf.org/html/rfc8610[Concise Data Definition Language (CDDL)] is a data description language for CBOR). You can also link:{attachmentsdir}/history.cddl[download the file].
```
nomeata, line 448: This puts all history into one leaf of the state tree.

Just wondering: Wouldn’t it be beneficial to put each change into the state tree individually, so once can query the latest only in a certified way?

In that case, I’d go for
```
/canister/<canister_id>/history/<change_index>/
```
where `<change_index>` is simply the (LEB-encoded) number of the index. 

But not sure if it’s important to support large histories efficiently.

spec/index.adoc
```diff
@@ -2283,6 +2297,7 @@ Arg = Blob;
 CallerId = Principal;
 
 Timestamp = Nat;
+CanHistoryIndex = Nat;
```
nomeata, line 2300: Question about naming: Saying “CanHistoryIndex” works, but is a bit round-about – the number makes sense even if we don’t record the history, and I’d lean towards something like `CanisterGeneration`. But not important, of course.

spec/index.adoc
```diff
@@ -2521,6 +2540,37 @@ CanStatus
...
+    }
+  | FromCanister {
+      canister_id : CanisterId;
+      canister_history_index : CanHistoryIndex;
+    }
+CodeDeploymentMode
+  = Install
+  | Reinstall
+  | Upgrade
+CanChangeType
```
nomeata, line 2555: Why not simply
```suggestion
CanChange
```
(it’s more than the `Type`)

spec/index.adoc
mraszyk: I'm not sure if `CanisterGeneration` should include controller changes which `CanisterHistoryIndex` (or `CanHistoryIndex` for short) does.

spec/index.adoc
mraszyk: Because then it clashes with the record including timestamp and change origin which are independent of the type of the change (e.g., code deployment or controllers change). Inlining this type into `CanChange` below would also only work in the textual spec, not in the actual code as far as I can see.

spec/index.adoc
nomeata: Fair point. Maybe ok then.

spec/history.cddl
```diff
@@ -0,0 +1,58 @@
+can-history = tagged<{
+  total_num_changes : unsigned
+  changes : [* can-change]
+}>
+
+can-change = {
```
ielashi, line 6: nit: I suggest expanding `can` and say `canister` for clarity? Otherwise, it can be confusing, as it's not clear whether this means it's a "canister change" or if it's something that "can change".

spec/index.adoc
ielashi: Could this not substantially increase the cost of certification? The history is trimmed to the point where it's feasible to return the entire history in a single response, so exposing all of it in one endpoint shouldn't add much complication to the caller,

spec/index.adoc
ielashi: How about `CanisterVersion`? I think it still makes sense for the version of a canister to change when controllers are changed.

Actually, now that I think about it, would it make sense to have some kind of semantic versioning, for example a canister can have a version `x.y`, where `x` signifies a "major" change where the canister is upgraded, installed, reinstalled, and `y` signifies a "minor" change like a controller change, settings change, etc?

spec/index.adoc
ielashi: An alternative would be to rename the `CanChange` below to `CanChangeEntry`, and rename this to `CanChange`?

spec/index.adoc
mraszyk: > Could this not substantially increase the cost of certification?

Not once [[MR-304]](https://dfinity.atlassian.net/browse/MR-304) is done.

ielashi: Works for me if performance won't be a problem. I guess we'll also need a `/canister/<canister_id>/history/latest/` that returns the latest change?

mraszyk: I was thinking of `/canister/<canister_id>/history/total_num_changes` to give the total number of changes.

mraszyk: > Works for me if performance won't be a problem.

Could we then store the entire history (and charge for it)?

spec/index.adoc
mraszyk: The issue is how we version the legacy canisters. Currently, their index would be 0 because they have 0 change entries in their history.

ielashi: We have this issue for all approaches, right? With a single history index, we'll also start at zero for existing canisters, I presume?

mraszyk: > We have this issue for all approaches

Not quite. I don't view index 0 as the first entry in the history, but as the number of entries in the history snapshot we want to reference (i.e., index 0 means "before history is recorded", index 1 means the first entry in the history etc.).

spec/history.cddl
```diff
@@ -0,0 +1,39 @@
...
+  change : canister-change
+}
+
+canister-change-origin = {
+  from : "user"
+  user_id : principal
+  //
+  from : "canister"
+  canister_id : principal
+  canister_history_index : unsigned
```
Dfinity-Bjoern, line 18: I think this is more fundamental than it may initially appear: What we are effectively saying here is that a canister is not purely identified by its canister id, but by the canister id _and_ the version. Then if we're consistent, we may ask questions like:
- Shouldn't we also expose this in the `caller` in inter-canister calls?
- There was once the idea of sending a message conditionally on the canister running a certain hash, shouldn't we allow sending a message conditionally on the target canister being at a certain version?
- Shouldn't the certified variables of a canister specify the version of the canister that produced them?

Of course we don't have to do all of this at once, but I think the change is (also witnessed by the fact that we'd actually have to change the message routing protocol) more fundamental than it initially seems.

Would appreciate other thoughts – so let me also ping @nomeata !

nomeata: I'd say the canister evolution is an internal to the canister, and something that, at least in general, should be of no concern to other parties.

Note that the universal canister can emulator all (or most) those changes without a history change. I'd hide whether a state change affects just the data or also the code. At least initially.

mraszyk: @nomeata you were already proposing something similar [here](https://github.com/dfinity/interface-spec/pull/19/files#diff-4de6dbd4bce188297247b78b84639ec28500b6e90a783241539e95c9e33fe4d0R2109)

nomeata: Ah, yes: When handling a callback the canister should know the generation that sent the callback, for the reasons explained in #19. But this is still internal (like the `env`), and not visible to the other canister. It seems that Björn is pondering whether to expose the generation to the other side.

I wonder if `canister_history_index` can be used as in #19, despite the fact that it can increase during the lifetime of one instance, or if we’d still need a separate instance generation counter. But I think it works: the canister can store the current index in `canister_init`, and can then reliably distinguish the current generation (history index ≥ the stored one) from old ones (history index < the stored one).

spec/index.adoc
mraszyk: During today's canisterOS meeting, we decided to keep a single index, in particular, because semantic versioning would typically bump the minor for code fixes and major for API changes whereas here we would bump the minor for non-code changes and major for all code changes.

spec/history.cddl
mraszyk: I think you're right that exposing `canister_history_index` via System API (e.g., `ic0.canister_history_index`) would enable to distinguish callbacks from the current generation from callbacks from past ones. Still, this would go beyond the scope of this PR.

My main motivation for introducing `canister_history_index` is the ability to implement the following notion: Given a canister C, identify all canisters D such that the following holds: for every canister E that is a (reflexive and transitive) controller of C, we have that E has been (reflexively and transitively) deployed by D.

If this property holds, then the canister D has “full” control over the current and future deployments of C.

Hence, it enables checking whether a canister (C) is managed by some DAO, e.g., SNS governance (canister D).

Here, checking if "E has been (reflexively and transitively) deployed by D" requires `canister_history_index` or sufficient gaps between the time-stamps of the individual deployments in the chain between D and E. In the latter case, however, it is not clear what is "sufficient" and, according to @daniel-wong-dfinity-org from the NNS team, "not relying on time for synchronization is highly preferred".

nomeata: I agree that having the history index here is sensible.

nomeata: This is a  useful discussion, and I think I see a clear path forward here that generalizes beyond the imminent need.

Multiple times we noticed a need to somehow identify and discern different points in the “canister-local timeline”:

* In the context of #19 we want to recognize canister generations (changes to the canister _code_).

* Here, the SNN want to recognize canister configurations (changes to the canister code and/or certain system metadata).

* application developers face difficulties with making sense of parallel update calls, or update calls and query calls, because of asynchronicity and potentially reaching different replica. They would dearly have some form of canister state counter included in the response to learn their real ordering. (Changes to the canister user state)

I have no reason to believe that this is the end, and likely more use cases will occur.

We could add three different concepts as we try to address these issues. But that's redundant and short-sighted design, because all these use cases are also solvable with a _finer_ counter. 

So my recommendation is to finally add the “canister state counter”. A simple number, bumped whenever anything about the canister state in the broad sense (code, system state, user state) changes. One counter to rule them all.

For example then it suffices for the SNN use case it all history entries include the canister state counter, and also the canister origin (as shown above): it can tell which interval it falls in.

This also nicely avoids the problem that we don't have a good name for CanisterHistoryIndex :-)

spec/index.adoc
mraszyk: In the canisterOS meeting, we decided to store the individual change entries as separate paths in the system state tree. This has a number of benefits:
- in the future, we can easily increase the maximum number of entries kept in the history (we'll use 1000 in the implementation for now) without increasing the maximum size of a single value in the system state tree
- a dashboard caching canister history can periodically poll for the total number of changes and only fetch the missing change entries

while it is still easy to fetch the entire history by requesting the prefix `/canister/<canister_id>/history`.

spec/history.cddl
mraszyk: Agreed! This suggestion sounds really useful!

Dfinity-Bjoern: Yes, that looks really good! Just to throw in another possible application of this: For "secure" queries (meaning: ask multiple replicas, aggregate the result) we have to make sure we execute the query at the right state. So far we only have block height (not nice, exposes protocol internals) or subnet time (not nice since not guaranteed to be strictly increasing). The canister state counter es *exactly* the right tool for that job as well.

nomeata: Exactly!

spec/history.cddl
```diff
@@ -0,0 +1,34 @@
+canister-change-entry = tagged<{
+  timestamp : timestamp
+  origin : canister-change-origin
+  change : canister-change
+}>
+
+canister-change-origin = {
+  from : "user"
```
peterpeterparker, line 8: doesn't the `from: user | canister` field give a hint if a principal is which is the opposite of the design of not knowing what a principal is?

spec/history.cddl
```diff
@@ -0,0 +1,34 @@
...
+  canister_id : principal
+  canister_history_index : unsigned
+  //
+  from : "system"
+}
+
+canister-change = {
+  kind : "canister-creation" / "code-uninstall"
+  //
+  kind : "code-install" / "code-reinstall" / "code-upgrade"
```
peterpeterparker, line 21: shouldn't the enum match the `install_code` spec?

https://github.com/dfinity/interface-spec/blob/f29fb083893c4d19de6373e1da6f74203a28679f/spec/ic.did#L93

spec/history.cddl
nomeata: I’d create one PR to introduce the counter concept itself, and then separate PRs (such as this one) to expose/use the counter in various places

(Maybe the plain system API exposure to the canister can be in the first PR, but I would not lump it together with unrelated big changes like the Canister History log.)

spec/history.cddl
nomeata: I wouldn’t say so. The design behind a principal on its own is that it is opaque (as possible). An in some applications, we (or some of us) like to keep the difference hidden – e.g. update calls vs. inter-canister calls. But for certain use cases (like this one) the distinction is important, and then it’s ok to put that information next to the principal.

(It would be fishy if we wouldn’t have the user/canister tag here and instead say ”parse the principal”; _that_ would be in opposition to treating principals as opaque identifiers as far as possible.)

peterpeterparker: Thanks for the clarification. It would be probably neat to add the keyword "(as possible)" in the docs (https://internetcomputer.org/docs/current/references/ic-interface-spec#principal). It makes the concept more clear in few words.

spec/history.cddl
mraszyk: I don't see this as a strict requirement since there also exist canister change kinds that are not members of this variant, e.g., `canister-creation` or `code-uninstall`. Do you see a reason why these three modes need to be encoded literally?

peterpeterparker: Not a did expert but assuming the two informations are linked, if I would implement this in object base language, the kind would probably be an extension of the spec because otherwise the types would have to be mapped. As the map would become an important part of the conversion and tracability of the information, I would rather like to spare it.

But then indeed, if the `canister-change { kind }` and `install_code { mode }` are not related at all, then it can be different.

Being said, I am not a scientist or backend architect, so don't mind me if you rather like to use a custom variant.

spec/history.cddl
mraszyk: Here's the [PR](https://github.com/dfinity/interface-spec/pull/106).

Dfinity-Bjoern:
Adding @roman-kashitsyn to discuss whether there is a potential issue with growing the tree.

spec/index.adoc
```diff
@@ -437,6 +437,23 @@ If the canister is not empty, it exists and contains the SHA256 hash of the curr
...
+The system can drop the oldest canister change entries from the list to keep its length bounded (at least 1000 change entries are guaranteed to remain in the list), but the indices of the remaining change entries are not affected.
+
+Every canister change entry consists of the system timestamp at which the change was performed, its origin (a user, a canister, or system), and its kind (canister creation, code change, or controllers change). The change origin can be from system if a canister is uninstalled because it ran out of cycles. Otherwise, the change origin includes the principal (called _originator_ in the following) that initiated the change and, if the originator is a canister, the index into the originator's canister history when the originator initiated the change (represented as the total number of the originator's changes at that time). Code changes are described by their kind (code uninstall, code install, code reinstall, code upgrade) and, if applicable, the SHA-256 hash of the newly deployed canister module. Controllers changes are described by the full new set of the canister controllers after the change.
+
+* `/canister/<canister_id>/total_num_changes` (natural):
```
roman-kashitsyn, line 453: Maybe
```suggestion
* `/canister/<canister_id>/state_counter` (natural):
```
?

spec/index.adoc
```diff
@@ -3994,27 +3994,6 @@ S with
     canister_state_counter[CanisterId] = N1
 ....
 
-==== Trimming canister history
```
derlerd-dfinity: I think it would still make sense to trim the history entries in the replicated state at 1k entries to avoid unbounded growth of the state, no?

mraszyk: You're right. Let me add this back.

Dfinity-Bjoern (approved):
LGTM, please merge as you see fit

spec/ic.did
```diff
@@ -16,6 +15,35 @@ type definite_canister_settings = record {
...
+    mode : variant {install; reinstall; upgrade};
+    module_hash : blob;
+  };
+  controllers_change : record {
+    controllers : vec principal;
+  };
+};
+
+type canister_change_entry = record {
+  timestamp : nat64;
```
mraszyk, line 41: @alin-at-dfinity suggested to call this timestamp_nanos. Agreed?

Dfinity-Bjoern: Sounds good.

spec/ic.did
```diff
@@ -106,6 +142,15 @@ service ic : {
       cycles: nat;
       idle_cycles_burned_per_day: nat;
   });
+  canister_info : (record {
+      canister_id : canister_id;
+      num_requested_changes : opt nat64;
+  }) -> (record {
+      total_num_changes : nat64;
+      changes : vec canister_change;
```
mraszyk, line 150: @roman-kashitsyn 's suggestion:
```
changes_total : nat64;
recent_changes : vec canister_change;
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
...
+    user_id : principal;
+  };
+  canister_change_from_canister : record {
+    canister_id : principal;
+    canister_version : opt nat64;
+  };
+};
+
+type canister_change_details = variant {
+  canister_creation : record {
```
Dfinity-Bjoern, line 29:
```suggestion
  creation : record {
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
...
+    controllers : vec principal;
+  };
+  canister_code_uninstall;
```
Dfinity-Bjoern, line 32:
```suggestion
  code_uninstall;
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
...
+  canister_code_deployment : record {
```
Dfinity-Bjoern, line 33:
```suggestion
  code_deployment : record {
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
...
+    mode : variant {install; reinstall; upgrade};
+    module_hash : blob;
+  };
+  canister_controllers_change : record {
```
Dfinity-Bjoern, line 37:
```suggestion
  controllers_change : record {
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
   freezing_threshold : nat;
 };
 
+type canister_change_origin = variant {
+  canister_change_from_user : record {
```
Dfinity-Bjoern, line 19:
```suggestion
  from_user : record {
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
...
+    user_id : principal;
+  };
+  canister_change_from_canister : record {
```
Dfinity-Bjoern, line 22:
```suggestion
  from_canister : record {
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
...
+  };
+  canister_controllers_change : record {
+    controllers : vec principal;
+  };
+};
+
+type canister_change = record {
+  timestamp_nanos : nat64;
+  canister_version : nat64;
+  change_origin : canister_change_origin;
```
Dfinity-Bjoern, line 45:
```suggestion
  origin : canister_change_origin;
```

spec/ic.did
```diff
@@ -16,6 +15,37 @@ type definite_canister_settings = record {
...
+  change_details : canister_change_details;
```
Dfinity-Bjoern, line 46:
```suggestion
  details : canister_change_details;
```

spec/ic.did
Dfinity-Bjoern: Interface spec meeting says:
- change `changes` to `recent_changes`
- leave `total_num_changes`

Dfinity-Bjoern (approved):
Discussed in interface spec meeting

PR #105, opened 2022-10-20
Add custom DNS argument to HTTP outcalls

yotamhc:
This PR adds an optional argument for HTTP outcalls to specify a custom DNS resolver that will be used with the corresponding call. This is now under development.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -111,6 +111,7 @@ service ic : {
   raw_rand : () -> (blob);
   http_request : (record {
     url : text;
+    custom_dns: opt text;
     max_response_bytes: opt nat64;
     method : variant { get; head; post };
     headers: vec http_header;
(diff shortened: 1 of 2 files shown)


Dfinity-Bjoern (approved):
Looks good overall, can you add an example of the formatting you expect? E.g. I've seen cases where IPv6 addresses are required to be in brackets and some where they're not, and I think it would be good to clarify.

yotamhc:
> Looks good overall, can you add an example of the formatting you expect? E.g. I've seen cases where IPv6 addresses are required to be in brackets and some where they're not, and I think it would be good to clarify.

Done. But let's wait with merging this. Marking as a draft for now.

yotamhc:
Closing this PR, as we will not do this for now.

PR #106, opened 2022-10-23
Introduce a canister state counter

mraszyk:
A *canister state counter* (represented as an unsigned integer) guaranteed to be incremented upon every change of the canister's code and settings is introduced. It is available to the canister via a system API call.

This PR only introduces the concept of a canister state counter. Its application will follow in separate PRs. Here we outline the anticipated applications of the canister state counter:

- used instead of *canister generation counter* proposed [here](https://github.com/dfinity/interface-spec/pull/19) for the sake of safe instantaneous (i.e., without stopping the canister) upgrades that handle all outstanding callbacks properly
- used instead of *canister history index* [here](https://github.com/dfinity/interface-spec/pull/103) to identify the canister's code version that triggered a change (e.g., code deployment) of another canister

Note. In the future, we might additionally guarantee that the state counter increases upon further changes of the canister's state. This is backwards-compatible because already now the state counter can be increased by the system at any time.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2307,7 +2307,6 @@ MethodCall = {
   arg: Blob;
   transferred_cycles: Nat;
   callback: Callback;
-  canister_state_counter: CanisterStateCounter;
 }
 
 UpdateFunc = WasmState -> Trap { cycles_used : Nat; } | Return {
@@ -4422,7 +4421,6 @@ ic0.call_new<es>(
       on_reject = Closure { fun = reject_fun; env = reject_env }
       on_cleanup = NoClosure
     };
-    canister_state_counter = es.params.sysenv.canister_state_counter;
   }
 
 ic0.call_data_append<es> (src : i32, size : i32) =
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1103,7 +1103,7 @@ Status `3` (stopped) can be observed, for example, in `canister_pre_upgrade` and
 [#system-api-canister-state-counter]
 === Canister state counter
 
-For each canister, the system maintains a _state counter_. Upon canister creation, it is set to 0, and it is incremented upon every change to the canister's state (except for changes affecting only the canister's balance or time), i.e., upon every (successful and unsuccessful) message execution (ingress message, inter-canister call, callback execution) on that canister, successful management canister call (`update_settings`, `install_code`, `uninstall_code`, `stop_canister`, `start_canister`, `deposit_cycles`, and `provisional_top_up_canister`) on that canister, change of the canister status of that canister from stopping to stopped, code uninstallation due to that canister running out of cycles, successful `inspect_message` execution for ingress messages, and charging canisters for (compute and memory) resource allocation and usage.
+For each canister, the system maintains a _state counter_. Upon canister creation, it is set to 0, and it is guaranteed to be incremented upon every change to the canister's state (except for changes affecting only the canister's time), i.e., upon every (successful and unsuccessful) message execution (ingress message, inter-canister call, callback execution) on that canister, successful management canister call (`update_settings`, `install_code`, `uninstall_code`, `stop_canister`, `start_canister`, `deposit_cycles`, and `provisional_top_up_canister`) on that canister, change of the canister status of that canister from stopping to stopped, code uninstallation due to that canister running out of cycles, successful `inspect_message` execution for ingress messages, and charging canisters for (compute and memory) resource allocation and usage. The system can arbitrarily increment the state counter also if the canister's state does not change.
 
 * `ic0.canister_state_counter : () -> i64`
 +
@@ -3768,7 +3768,7 @@ S with
 ....
 
 
-==== Time progressing and cycle consumption
+==== Time progressing, cycle consumption, and canister state counter increments
 
 Time progresses. Abstractly, it does so independently for each canister, and in unspecified intervals.
 
@@ -3808,6 +3808,19 @@ State after::
 ....
 S with
     system_time = T1
+....
+
+Finally, the canister state counter can be incremented arbitrarily:
+
+Conditions::
+....
+    N0 = S.canister_state_counter[CanisterId]
+    N1 > N0
+....
+State after::
+....
+S with
+    canister_state_counter[CanisterId] = N1
 ....
 
 ==== Query call
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1103,7 +1103,7 @@ Status `3` (stopped) can be observed, for example, in `canister_pre_upgrade` and
 [#system-api-canister-state-counter]
 === Canister state counter
 
-For each canister, the system maintains a _state counter_. Upon canister creation, it is set to 0, and it is incremented upon every change to the canister's state (except for changes affecting only the canister's balance or time), i.e., upon every successful message execution on that canister, successful management canister call (`update_settings`, `install_code`, `uninstall_code`, `stop_canister`, `start_canister`) on that canister, canister status (running, stopping, stopped) change of that canister, and code uninstallation due to that canister running out of cycles.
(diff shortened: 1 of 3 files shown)


spec/index.adoc
```diff
@@ -1100,6 +1100,15 @@ Status `1` indicates running, `2` indicates stopping, and `3` indicates stopped.
 +
 Status `3` (stopped) can be observed, for example, in `canister_pre_upgrade` and can be used to prevent accidentally upgrading a canister that is not fully stopped.
 
+[#system-api-canister-state-counter]
+=== Canister state counter
+
+For each canister, the system maintains a _state counter_. Upon canister creation, it is set to 0, and it is incremented upon every change to the canister's state (except for changes affecting only the canister's balance or time), i.e., upon every (successful and unsuccessful) message execution (ingress message, inter-canister call, callback execution) on that canister, successful management canister call (`update_settings`, `install_code`, `uninstall_code`, `stop_canister`, `start_canister`, `deposit_cycles`, and `provisional_top_up_canister`) on that canister, change of the canister status of that canister from stopping to stopped, code uninstallation due to that canister running out of cycles, successful `inspect_message` execution for ingress messages, and charging canisters for (compute and memory) resource allocation and usage.
```
nomeata, line 1106: I'd say ”at least”, but allow spurious bumps. This may be useful for the system, e.g. the system may bump the counter any time. The only guarantee is that if one of the events listed here has happened, the counter has strictly increased.

For example it is a reasonable implementation (imho) to have one counter for all canisters.

Or one that bumps the counter upon every management calls, even unsuccessful ones.

In the formal spec, this could be modeled by a spurious counter bump transition.

mraszyk: > it is a reasonable implementation (imho) to have one counter for all canisters.

This is indeed a good idea, making it possible to also bump the (single in the implementation) state counter if time progresses. I wonder if bumping the state counter whenever time progresses shouldn't be guaranteed, too, given the possibility to have a single canister state counter.

nomeata: Good question. Would the difference be observable? Under what circumstances can you observe a changed time when we don't already observe a change of the state counter?

Probably query calls… but that's really a good reason not to require a state bump just because time has progressed?

mraszyk: A query call returning `ic0.time` could return different results for the same canister state counter. Hence, we'd need to check the state counter *and time-stamp* when comparing query call results across replicas.

nomeata: I see. Still probably not useful to bump the counter - I expect the counter to be part of the certified state, and it's probably good to not have to touch that for a canister that only serves queries.

So to reproduce a query call one would have to fix state (via the couner) plus all other inputs (time, of course arg and caller, maybe balance)?

mraszyk: > So to reproduce a query call one would have to fix state (via the couner) plus all other inputs (time, of course arg and caller, maybe balance)?

Indeed. Changes to the canister's balance are reflected in the state counter, so one would only need to fix the state counter, time-stamp, and query call arguments (arg and sender).

nomeata: > Changes to the canister's balance are reflected in the state counter

Are they really? Isn’t there the continuous, time-dependent cycle cost?

mraszyk: > Isn’t there the continuous, time-dependent cycle cost?

Indeed. But this is reflected in the formal model by bumping the counter for the cycle consumption transition and can be reflected in the implementation by bumping the counter upon call to `charge_canisters_for_resource_allocation_and_usage` in the scheduler.

nomeata: Hmm, ok. I kinda assumed that canisters without calls are completely untouched and the current cycle balance is calculated on the fly when observed, to avoid state updates. But if that is not the case (and maybe it doesn't even work), then requiring that balance changes imply counter bumps is probably reasonable.

ulan:
In order to support of query calls in query methods (a.k.a. inter-canister query calls), we need keep track of state changes in queries. This means that we can have two different canister states (one evolving via updates, another via queries) that have the same counter and the same canister id. Would that be a problem?

Would it make sense to add some kind of a prefix to the counter in order to distinguish these states? Something like `(Context, Counter)` where `Context = Update | Query CallContextId`.

ulan:
> Note. We do not bump the canister state counter if only the time-stamp increases because this would potentially require an update of the canister state counter for all canisters in every round 

Does it mean that the counter cannot be used for secure queries? Because the current time is observable by the canister via the `ic0.time()` API.

ulan:
> used to derive the actual execution order of update and query calls submitted in parallel

What is the context here?  So far we don't specify the execution order of update and query calls.

nomeata:
> Would that be a problem?

It's something to watch out for, but I don't expect it to be a problem per se. The canister state counter is effectively a per-canister local strictly increasing unitless clock, and if you fork the evolution of the canister (as you kinda do in query calls), time starts ticking independently for them

mraszyk:
> > Note. We do not bump the canister state counter if only the time-stamp increases because this would potentially require an update of the canister state counter for all canisters in every round
> 
> Does it mean that the counter cannot be used for secure queries? Because the current time is observable by the canister via the `ic0.time()` API.

For secure queries, we'd need to fix the state counter and time-stamp for every query call (and also for every call from within this query call). But if we go for a single counter per replica, then we might be able to update the counter also as time progresses (in which case I'd make this part of the spec).

mraszyk:
> > used to derive the actual execution order of update and query calls submitted in parallel
> 
> What is the context here? So far we don't specify the execution order of update and query calls.

Sure, it's not specified and I don't propose to fix any particular order, but using the state counter (if returned along the response message) we can determine the actual order of the update and query calls (or the callback triggering the response).

Dfinity-Bjoern:
> > > Note. We do not bump the canister state counter if only the time-stamp increases because this would potentially require an update of the canister state counter for all canisters in every round
> > 
> > 
> > Does it mean that the counter cannot be used for secure queries? Because the current time is observable by the canister via the `ic0.time()` API.
> 
> For secure queries, we'd need to fix the state counter and time-stamp for every query call (and also for every call from within this query call). But if we go for a single counter per replica, then we might be able to update the counter also as time progresses (in which case I'd make this part of the spec).

I think the granularity of counters that one would see would still be at the level of "a block" – meaning the queries would still be executed on a consistent certified state. So I don't see a finer-grained counter to be a problem as long as we keep executing queries on certified state.

spec/index.adoc
```diff
@@ -1100,6 +1100,15 @@ Status `1` indicates running, `2` indicates stopping, and `3` indicates stopped.
 +
 Status `3` (stopped) can be observed, for example, in `canister_pre_upgrade` and can be used to prevent accidentally upgrading a canister that is not fully stopped.
 
+[#system-api-canister-state-counter]
+=== Canister state counter
+
+For each canister, the system maintains a _state counter_. Upon canister creation, it is set to 0, and it is guaranteed to be incremented upon every change to the canister's state (except for changes affecting only the canister's time), i.e., upon every (successful and unsuccessful) message execution (ingress message, inter-canister call, callback execution) on that canister, successful management canister call (`update_settings`, `install_code`, `uninstall_code`, `stop_canister`, `start_canister`, `deposit_cycles`, and `provisional_top_up_canister`) on that canister, change of the canister status of that canister from stopping to stopped, code uninstallation due to that canister running out of cycles, successful `inspect_message` execution for ingress messages, and charging canisters for (compute and memory) resource allocation and usage. The system can arbitrarily increment the state counter also if the canister's state does not change.
```
dsarlis, line 1106: > change of the canister status of that canister from stopping to stopped

What about change of canister status from running to stopping? Or stopped to running? You are incrementing the counter in the formal model section in these cases. More generally, are you trying to capture every possible case where a change in the canister state might happen or is this list supposed to be more like examples?

dsarlis:
> > > used to derive the actual execution order of update and query calls submitted in parallel
> > 
> > 
> > What is the context here? So far we don't specify the execution order of update and query calls.
> 
> Sure, it's not specified and I don't propose to fix any particular order, but using the state counter (if returned along the response message) we can determine the actual order of the update and query calls (or the callback triggering the response).

I think the point of the question was "What is the use case"? Naively, I could say you can use `ic0.time()` for a similar purpose (of course it's not as fine grained as the counter you're proposing here because ingress messages executed on the same round would all report the same time).

spec/index.adoc
mraszyk: The cases you describe are covered by "successful IC management canister calls". Going from stopping to stopped is an internal system transition not associated to a particular IC management canister call.

spec/index.adoc
```diff
@@ -3134,6 +3150,7 @@ S with
       controllers[A.canister_id] = A.settings.controllers
     if A.settings.freezing_threshold is not null:
       freezing_threshold[A.canister_id] = A.settings.freezing_threshold
+    canister_state_counter[A.canister_id] = S.canister_state_counter[A.canister_id] + 1
```
mraszyk, line 3153: Let's just increment the state counter for every change of the canister's settings unless it's easier to increment it only for changes of its controllers.

Dfinity-Bjoern (approved):
As agreed in a meeting on the topic – thanks @mraszyk !

PR #107, opened 2022-10-24
Rename the transform context to arg

rumenov:
Going over the spec it seems we use the word "context" more in the situation of a callback. 

Also in the current spec it does seem that we already use the `arg` convention for the `install_code` method and within HTTPS API as part of the CBOR request. Those occurrences seem semantically closer to the transform context intent than a callback intent.


rumenov:
Discussed on slack and it seems that the proposal doesn't make much sense if we concentrate mainly on the naming of the arguments.

PR #109, opened 2022-10-25
update http_request specification

mraszyk:


spec/index.adoc
```diff
@@ -1658,19 +1658,20 @@ It is important to note the following for the usage of the `POST` method:
...
-The `2MiB` size limit also applies to the value returned by the `transform` function.
+Each request can specify the maximal expected size for the response body from the remote HTTP server. The upper limit on the size of a response body defaults to `2MiB` if no maximal size value is specified.
+An error will be returned when the response body is larger than the maximal size.
 
 The following parameters should be supplied for the call:
 
-- `url` - the requested URL. The URL may specify a custom port number.
-- `max_response_bytes` - optional, specifies the maximal size of the response in bytes. Any value less than or equal to `2MiB` is accepted. The call will be charged based on this parameter. If not provided, the maximum of `2MiB` will be used.
+- `url` - the requested URL. The URL must be valid according to https://www.ietf.org/rfc/rfc3986.txt[RFC-3986] and its length must not exceed 65534. The URL may specify a custom port number.
+- `max_response_bytes` - optional, specifies the maximal size of the response *body* in bytes. Any value less than or equal to `2MiB` is accepted. The call will be charged based on this parameter. If not provided, the maximum of `2MiB` will be used.
```
yotamhc, line 1667: Is it really just accounting for the body size? @tthebst

spec/index.adoc
```diff
@@ -1680,6 +1681,8 @@ The returned response (and the response provided to the `transform` function, if
 The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
 When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
 
+The size of the response (from the remote server if no `transform` function is specified or after `transform`) in Candid encoding (including `status`, `headers`, and `body`) must not exceed 2045952 (`2MiB - 50KiB`).
```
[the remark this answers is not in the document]
yotamhc: I don't know what we usually do. I find it hard to tell.

spec/index.adoc
mraszyk: I've tested that in this [PR](https://github.com/dfinity/ic-hs/pull/100).

tthebst: Currently yes. The limit is only used to limit the response that we receive over the wire. It is used to account for the bandwidth cost. But thinking about it it might also be a good idea to use it to limit the size of the transform.

mraszyk: > The limit is only used to limit the response that we receive over the wire.

The headers and status line are also received over the wire, but not accounted for.

spec/index.adoc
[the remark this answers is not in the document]
rumenov: @Dfinity-Bjoern , can you please elaborate what are your thoughts ?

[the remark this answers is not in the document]
rumenov: I don't think we need this paragraph. 
 Does candid do any compression ? How much would the size of the http response would change after the candid encoding ?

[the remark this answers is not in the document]
rumenov: I would formulate this in the spirit that the transform should not produce data bigger than our consensus blocks otherwise the operation would fail without going much into candid or strict limits.

spec/index.adoc
```diff
@@ -1675,27 +1675,31 @@ It is important to note the following for the usage of the `POST` method:
 
 For security reasons, only HTTPS connections are allowed (URLs must start with `https://`). The IC uses industry-standard root CA lists to validate certificates of remote web servers.
 
-Each request can specify the maximal expected size for the response from the remote HTTP server. The upper limit on the size of a response defaults to `2MiB` if no maximal size value is specified.
+The *size* of an HTTP request or response is the total number of bytes representing the names and values of HTTP headers and the HTTP body. Each request can specify a maximal size for the response from the remote HTTP server. The upper limit on the maximal size for the response defaults to `2MB` (`2,000,000B`) if no maximal size value is specified.
```
rumenov, line 1678: If someone specifies 5MB it will again default to 2MB which is not very clear from this text.

rumenov, line 1678: If someone specifies 5MB it will again default to 2MB which is not very clear from this text.

mraszyk: If someone specifies 5MB, then the request gets rejected [here](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/types/types/src/canister_http.rs?L202). I tried to emphasize the maximum allowed value of 2MB in a recent commit.

mraszyk: Duplicate.

spec/index.adoc
```diff
@@ -1675,27 +1675,32 @@ It is important to note the following for the usage of the `POST` method:
...
+
 The returned response (and the response provided to the `transform` function, if specified) contains the following fields:
 
 - `status` - the response status (e.g., 200, 404)
 - `headers` - list of HTTP response headers and their corresponding values
 - `body` - the response's body
 
-The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
-When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
+The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc. The maximal size of the blob produced by the `transform` function is `2MB` (`2,000,000B`). When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
```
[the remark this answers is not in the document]
yotamhc: Also

> When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
This is mentioned below in a separate paragraph. So I'd either remove it from here or remove the other paragraph (I prefer keeping the separate paragraph as it is more informative, or merging the two).

yotamhc (approved):
Looks good but please see my comments before merging

spec/index.adoc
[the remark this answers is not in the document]
mraszyk: I changed "The maximal size of the blob" to "The maximal number of bytes representing the response". I agree that talking about a blob is confusing given `ic.did`. On the other hand, the slack we have above `2MB` might not be enough for the transform function, e.g., because a custom candid encoder in the transform might include a huge type table (if no transform function is provided, then the system uses a candid encoder for which the slack is sufficient).

I removed the note on caller's identity in transform because the sentence "The identity of the caller for a valid invocation of the `transform` function is `aaaaa-aa`." is wrong. The caller is `aaaaa-aa` only if the transform function is invoked by the system for canister http_request calls, but a valid invocation need not be performed by the system only.

I added the (useful) sentence "This information can be used by developers to implement access control mechanism for this function." from the note into the main text.

spec/index.adoc
```diff
@@ -1675,29 +1675,34 @@ It is important to note the following for the usage of the `POST` method:
...
 
 - `status` - the response status (e.g., 200, 404)
 - `headers` - list of HTTP response headers and their corresponding values
 - `body` - the response's body
 
-The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc.
-When the transform function was invoked due to a canister HTTP request, the caller's identity is the principal of the management canister.
+The `transform` function may, for example, transform the body in any way, add or remove headers, modify headers, etc. The maximal number of bytes representing the response produced by the `transform` function is `2MB` (`2,000,000B`).
+
+When the transform function was invoked by the system due to a canister HTTP request, the caller's identity is the principal of the management canister. This information can be used by developers to implement access control mechanism for this function.
```
yotamhc, line 1700:
```suggestion
When the transform function is invoked by the system due to a canister HTTP request, the caller's identity is the principal of the management canister. This information can be used by developers to implement access control mechanism for this function.
```

spec/index.adoc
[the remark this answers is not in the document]
yotamhc: Thanks! Just added a small writing fix but otherwise it looks good to me.

Dfinity-Bjoern (approved):
Thanks @mraszyk – I agree this is what we discussed.

32 KiB for a header name seems very generous, but I guess it also doesn't hurt.

PR #111, opened 2022-11-03
Introduce canister global timer

mraszyk:


spec/index.adoc
```diff
@@ -962,6 +964,14 @@ For periodic or time-based execution, the WebAssembly module can export a functi
 
 NOTE: While an implementation will likely try to keep the interval between `canister_heartbeat` invocations to within a few seconds, this is not formally part of this specification.
 
+==== Global timer
+
+For time-based execution, the WebAssembly module can export a function with name `canister_global_timer`. This function is called if the canister has set its global timer (using the System API function `ic0.global_timer_set`) and the current time (as returned by the System API function `ic0.time`) has passed the value of the global timer. Once the function `canister_global_timer` is scheduled, the canister's global timer is reset. The global timer is also reset upon changes to the canister's Wasm module (calling `install_code`, `uninstall_code` methods of the management canister or if the canister runs out of cycles). In particular, the function `canister_global_timer` won't be scheduled again unless the canister sets the global timer again (using the System API function `ic0.global_timer_set`). The global timer scheduling algorithm is implementation-defined.
```
dfinity-berestovskyy, line 969: Shall we maybe try to split this into two paragraphs? Like setup in one and scheduling in the following...

spec/index.adoc
```diff
@@ -962,6 +964,14 @@ For periodic or time-based execution, the WebAssembly module can export a functi
...
+
+`canister_global_timer` is triggered by the IC, and therefore has no arguments, no caller, and cannot reply or reject.
```
dfinity-berestovskyy, line 971: Maybe it's worth explicitly mentioning that we can still initiate new calls?

spec/index.adoc
```diff
@@ -990,7 +1000,7 @@ The comment after each function lists from where these functions may be invoked:
 * `C`: from a cleanup callback
 * `s`: the `(start)` module initialization function
 * `F`: from `canister_inspect_message`
-* `H`: from `canister_heartbeat`
+* `H`: from `canister_heartbeat` or `canister_global_timer`
```
dfinity-berestovskyy, line 1003: Inside the code we've decided to rename this `heartbeat` or `global timer` into `system task`, as they share a lot of code. Not sure if it's relevant for the spec, but JFYI.

spec/index.adoc
```diff
@@ -2973,6 +3047,9 @@ then
     if res.new_certified_data ≠ NoCertifiedData:
       certified_data[M.receiver] = res.new_certified_data
 
+    if res.new_global_timer ≠ NoGlobalTimer:
```
dfinity-berestovskyy, line 3050: Should we always set it, without any conditions?

spec/index.adoc
mraszyk: > we can still initiate new calls

You mean it's possible to initiate new calls from the `canister_global_time` function? Then this remark would apply analogously to `canister_heartbeat`.

spec/index.adoc
mraszyk, line 969: Could you please confirm that the following text is accurate: "Once the function `canister_global_timer` is scheduled, the canister's global timer is reset. The global timer is also reset upon changes to the canister's Wasm module (calling `install_code`, `uninstall_code` methods of the management canister or if the canister runs out of cycles). In particular, the function `canister_global_timer` won't be scheduled again unless the canister sets the global timer again (using the System API function `ic0.global_timer_set`)."

spec/index.adoc
mraszyk: While this does not matter now, this formulation is a bit more general as it would allow to distinguish whether a change was made or not. Changes to the certified variables are handled analogously.

spec/index.adoc
mraszyk: I see 4 occurrences of the regular expression `heartbeat.*or.*timer` scattered throughout the spec. We could abbreviate, but then we'd need to introduce the abbreviation properly so I'm not sure about this change.

spec/index.adoc
dfinity-berestovskyy: Yes, we can initiate new calls both from the heartbeat and the timer. Up to you though, as technically we are correct.

spec/index.adoc
dfinity-berestovskyy: Yeah, to me `deactivated` sounds straightforward. But reset is probably also fine...

spec/index.adoc
```diff
@@ -4264,6 +4265,8 @@ query_methods[method] = λ (arg, caller, sysenv) → λ wasm_state →
   if es.calls ≠ [] then Trap {cycles_used = es.cycles_used;}
   if es.ingress_filter ≠ Reject then Trap {cycles_used = es.cycles_used;}
   if es.response = NoResponse then Trap {cycles_used = es.cycles_used;}
+  if es.new_certified_data ≠ NoCertifiedData then Trap {cycles_used = es.cycles_used;}
+  if es.new_global_timer ≠ NoGlobalTimer then Trap {cycles_used = es.cycles_used;}
```
nomeata, line 4269: Maybe this pattern for error handling doesn’t quite scale anymore :-)

Maybe it is more direct to have an `es.context` that is an enumeration of all the various contexts for system calls (like that table in the spec), and the assert the right context in each system call.

Or remove all that noise from here and _just_ point to the table, i.e. make the trap implicit.

mraszyk: I've had this on my TODO list and your comment made me implement this. Thank you!

PR #118, opened 2022-11-15
Changes to the Bitcoin integration API: Sending transactions and computing fees

THLO:
This PR cleans up and extends some parts of the Bitcoin integration API:

- The plan is no longer to resubmit transactions periodically. The functionality of the `send_transaction` endpoint is updated accordingly.
- The fee API was underspecified in that it didn't specify what method is used to compute the fee percentiles. The text is updated to make this explicit.


Dfinity-Bjoern (approved):
Thanks @THLO!

PR #120, opened 2022-11-18
rename canister state counter to canister version

mraszyk:
The name "canister state counter" could be confusing because the counter is not incremented upon every change of the canister state. Hence, we agreed to rename it to "canister version".

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1116,14 +1116,14 @@ Status `1` indicates running, `2` indicates stopping, and `3` indicates stopped.
 +
 Status `3` (stopped) can be observed, for example, in `canister_pre_upgrade` and can be used to prevent accidentally upgrading a canister that is not fully stopped.
 
-[#system-api-canister-state-counter]
-=== Canister state counter
+[#system-api-canister-version]
+=== Canister version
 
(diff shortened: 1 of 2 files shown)


nomeata:
It’s not? I thought that was the conclusion that it would be most useful if we _do_ have a state counter, and it would subsume meta data versions etc.? Did that plan get dropped along the way somewhere?

mraszyk:
> It’s not? I thought that was the conclusion that it would be most useful if we _do_ have a state counter, and it would subsume meta data versions etc.? Did that plan get dropped along the way somewhere?

This plan did not get dropped entirely, but we agreed to start off with a counter that only increases upon changes to code and settings while still specifying that it might also increase in further cases. This way, we can further refine this (now coarse-grained) counter in the future.

nomeata:
I see. Too bad that we are not fixing the query-consitency-problem as well while we are at it, but I am used to DFINITY prioritizing those itches that are actively felt by the people on the team at the moment, over itches that are only felt by others, or worse, only hypothesized to be felt by them. (Certified variable design, I am looking at you :-))

Anyways, I hope the rename won’t get in the way of eventually getting the fine grained counter, and that nobody in the meantime relies on the state counting _not_ incrementing between code/setting changes :-)

Dfinity-Bjoern (approved):
Thanks!

PR #124, opened 2022-12-01
Bitcoin general availability release

dietersommer:
* Maximum page size for returned UTXOs
* Links to pricing and further detailed documentation

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1759,7 +1759,7 @@ NOTE: The IC Bitcoin API is considered EXPERIMENTAL. Canister developers must be
 
 The Bitcoin functionality is exposed via the management canister.
 Information about Bitcoin can be found in the https://developer.bitcoin.org/devguide/[Bitcoin developer guides].
-Invoking the functions of the Bitcoin API will cost cycles. The concrete costs for each function are yet to be determined.
+Invoking the functions of the Bitcoin API will cost cycles. We refer the reader to the [Bitcoin documentation](https://internetcomputer.org/docs/current/developer-docs/integrations/bitcoin/bitcoin-how-it-works) for further relevant information and the [IC pricing page](https://internetcomputer.org/docs/current/developer-docs/deploy/computation-and-storage-costs) for information of pricing for the feature for Bitcoin Testnet and Mainnet.
 
 [#ic-bitcoin_get_utxos]
 === IC method `bitcoin_get_utxos`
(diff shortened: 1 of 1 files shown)


spec/index.adoc
```diff
@@ -1797,7 +1797,7 @@ For a newly discovered block, a regular Bitcoin (full) node therefore provides a
 than the Bitcoin component, which implies that it is advisable to set the number of confirmations to a reasonably large value, such as 6,
 to gain confidence in the correctness of the returned UTXOs.
 
-There is an upper bound of 100,000 UTXOs that can be returned in a single request. For addresses that contain sufficiently many UTXOs,
+There is an upper bound of 1,000 UTXOs that can be returned in a single request. For addresses that contain sufficiently many UTXOs,
```
ielashi, line 1800: I actually think we should put this higher at, say, 10k. The implementation can still cap it at 1,000, but having the upper bound here by 10k allows us to increase the implementation's cap without breaking the spec.

dietersommer: Sounds good to me, @THLO, are you fine with 10K as well?

dprats (approved):

spec/index.adoc
```diff
@@ -1759,7 +1759,7 @@ NOTE: The IC Bitcoin API is considered EXPERIMENTAL. Canister developers must be
 
 The Bitcoin functionality is exposed via the management canister.
 Information about Bitcoin can be found in the https://developer.bitcoin.org/devguide/[Bitcoin developer guides].
-Invoking the functions of the Bitcoin API will cost cycles. The concrete costs for each function are yet to be determined.
+Invoking the functions of the Bitcoin API will cost cycles. We refer the reader to the [Bitcoin documentation](https://internetcomputer.org/docs/current/developer-docs/integrations/bitcoin/bitcoin-how-it-works) for further relevant information and the [IC pricing page](https://internetcomputer.org/docs/current/developer-docs/deploy/computation-and-storage-costs) for information of pricing for the feature for Bitcoin Testnet and Mainnet.
```
THLO, line 1762:
```suggestion
Invoking the functions of the Bitcoin API will cost cycles. We refer the reader to the [Bitcoin documentation](https://internetcomputer.org/docs/current/developer-docs/integrations/bitcoin/bitcoin-how-it-works) for further relevant information and the [IC pricing page](https://internetcomputer.org/docs/current/developer-docs/deploy/computation-and-storage-costs) for information on pricing both for the Bitcoin mainnet and testnet.
```

spec/index.adoc
THLO: Yes, that's the number @ielashi and I agreed on.

THLO (approved):

spec/index.adoc
dietersommer: Set to 10,000 as agreed.

Dfinity-Bjoern (approved):

PR #128, opened 2022-12-07
Add CBOR and CDDL intro sections

nathanosdev:
I found the section regarding encoding of certificates a bit difficult to follow at first, namely because I was not familiar with CDDL. Google searches will return links regarding "Common Development and Distribution License". Of course if you read the whole spec from start to finish then you will eventually find more information regarding CDDL, but I don't believe this is the way that everyone will read the spec. So I propose adding intro sections near the beginning of the document regarding CDDL and CBOR that can be referenced anywhere these technologies are mentioned.

Other smaller changes in this PR:
- replaced links to obsolete CBOR spec with the current spec
- Fixed one instance of "internet computer" -> "Internet Computer"

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -345,7 +345,7 @@ where `s` is the SHA-256 hash of the `seed` used in the public key and `m` is th
 --
 
 [#supplementary-technologies]
-== Supplementary Technologies
+=== Supplementary Technologies
 
 [#cbor]
 === CBOR
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -371,7 +371,7 @@ Particular concepts to note from the spec are:
 * https://www.rfc-editor.org/rfc/rfc8949#self-describe[CBOR Self-Describe]
 
 [#cddl]
-=== CDDL
+==== CDDL
 
 The https://tools.ietf.org/html/rfc8610[Concise Data Definition Language (CDDL)] is a data description language for CBOR. It is used at various points throughout this document to describe how certain data structures are encoded with CBOR.
 
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -301,7 +301,7 @@ You can also view the wrapping in https://lapo.it/asn1js/#MF4wDAYKKwYBBAGDuEMBAQ
 ====
 --
 
-* The signature is a CBOR value consisting of a data item with major type 6 (“Semantic tag”) and tag value `55799` (see https://tools.ietf.org/html/rfc7049#section-2.4.5[Self-Describe CBOR]), followed by a map with three mandatory fields:
+* The signature is a CBOR (see <<cbor>>) value consisting of a data item with major type 6 (“Semantic tag”) and tag value `55799`, followed by a map with three mandatory fields:
   - `authenticator_data` (`blob`): WebAuthn authenticator data.
   - `client_data_json` (`text`): WebAuthn client data in JSON representation.
   - `signature` (`blob`): Signature as specified in the https://www.w3.org/TR/webauthn/#signature-attestation-types[WebAuthn w3c recommendation], which means DER encoding in the case of an ECDSA signature.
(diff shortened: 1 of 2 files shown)


spec/index.adoc
```diff
@@ -344,6 +344,44 @@ where `signing_canister_id` is the id of the signing canister and `reconstruct`
...
+55799({
+  "content": {
+    "request_type": "call",
+    "canister_id": h'ABCD01',
+    "method_name": "say_hello",
+    "arg": h'0061736D01000000'
+  },
+  "sender_sig": h'DEADBEEF',
+  "sender_pubkey": h'B7A3C12DC0C8C748AB07525B701122B88BD78F600C76342D27F25E5F92444CDE'
+})
```
mraszyk, line 373: The other fields (arg, sender_sig, sender_pubkey) are not valid either (I suppose). Do we want this example to be a realistic one?

nathanosdev: I copied the example from the "CBOR encoding of requests and responses" section. I don't think the example necessarily needs to be realistic, but I also don't think it should be confusing or misleading. The only thing I want to demonstrate here is the before/after of decoding using cbor.me. Maybe it's more sensible to use a basic example from the CBOR spec: https://www.rfc-editor.org/rfc/rfc8949.html#name-examples-of-encoded-cbor-da

I think `82 61 61 a1 61 62 61 63` -> `["a", {"b": "c"}]` could be a nice example, what do you think?

spec/index.adoc
```diff
@@ -344,6 +344,44 @@ where `signing_canister_id` is the id of the signing canister and `reconstruct`
 where `s` is the SHA-256 hash of the `seed` used in the public key and `m` is the SHA-256 hash of the payload.
 --
 
+[#supplementary-technologies]
+== Supplementary Technologies
+
+[#cbor]
+=== CBOR
+
+https://www.rfc-editor.org/rfc/rfc8949[Concise Binary Object Representation(CBOR)] is a data format with a small code footprint, small message size and an extensible interface. CBOR is used extensively throughout the Internet Computer as the primary format for data exchange between components within the system.
```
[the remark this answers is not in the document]
nathanosdev: Thanks!

nomeata (commented):
Thanks, this is helpful!

spec/index.adoc
mraszyk: I'd prefer an unrelated example in that case (if we do not make the request a proper one).

nathanosdev: I've replaced it with the generic example now, thank you

mraszyk (approved):
Looks good to me. Thank you!

spec/index.adoc
```diff
@@ -344,6 +344,35 @@ where `signing_canister_id` is the id of the signing canister and `reconstruct`
...
+....
+82 61 61 a1 61 62 61 63
+....
+
+Can be converted into the following CBOR diagnostic format:
+....
+["a", {"b": "c"}]
+....
+
+Particular concepts to note from the spec are:
```
Dfinity-Bjoern, line 367:
```suggestion
Particular concepts to note from the spec are:

```

spec/index.adoc
```diff
@@ -344,6 +344,36 @@ where `signing_canister_id` is the id of the signing canister and `reconstruct`
 where `s` is the SHA-256 hash of the `seed` used in the public key and `m` is the SHA-256 hash of the payload.
 --
 
+[#supplementary-technologies]
+== Supplementary Technologies
```
Dfinity-Bjoern, line 348:
```suggestion
=== Supplementary Technologies
```

spec/index.adoc
```diff
@@ -344,6 +344,36 @@ where `signing_canister_id` is the id of the signing canister and `reconstruct`
 where `s` is the SHA-256 hash of the `seed` used in the public key and `m` is the SHA-256 hash of the payload.
 --
 
+[#supplementary-technologies]
+=== Supplementary Technologies
+
+[#cbor]
+=== CBOR
```
Dfinity-Bjoern, line 351:
```suggestion
==== CBOR
```

spec/index.adoc
```diff
@@ -344,6 +344,36 @@ where `signing_canister_id` is the id of the signing canister and `reconstruct`
...
+....
+
+Particular concepts to note from the spec are:
+
+* https://www.rfc-editor.org/rfc/rfc8949#name-specification-of-the-cbor-e[Specification of the CBOR Encoding]
+* https://www.rfc-editor.org/rfc/rfc8949#name-major-types[CBOR Major Types]
+* https://www.rfc-editor.org/rfc/rfc8949#self-describe[CBOR Self-Describe]
+
+[#cddl]
+=== CDDL
```
Dfinity-Bjoern, line 374:
```suggestion
==== CDDL
```

Dfinity-Bjoern (approved):

PR #129, opened 2022-12-19
Add optional canister_id argument to provisional_create_canister_with_cycles

dsarlis:
This PR adds an optional `specified_id` argument to `provisional_create_canister_with_cycles` which is going to be used by dfx in order to allow users to install third-party canister dependencies for their canisters locally under the same ids that these canisters have on mainnet.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -3753,6 +3753,7 @@ S with
       canister_id = A.specified_id
     else:
       canister_id = CanisterId
+    canisters[canister_id] = EmptyCanister
     time[canister_id] = CurrentTime
     global_timer[canister_id] = 0
     if A.settings.controllers is not null:
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1764,7 +1764,7 @@ NOTE: Currently, the Internet Computer mainnet only supports URLs that resolve t
 [#ic-provisional_create_canister_with_cycles]
 === IC method `provisional_create_canister_with_cycles`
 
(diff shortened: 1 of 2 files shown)


spec/index.adoc
```diff
@@ -3743,34 +3743,39 @@ Conditions::
     M.arg = candid(A)
     is_system_assigned CanisterId
     CanisterId ∉ dom(S.canisters)
+    A.specified_id ∉ dom(S.canisters)
+
 ....
 State after::
 ....
 S with
-    canisters[CanisterId] = EmptyCanister
```
mraszyk, line 3750: this line got lost

nomeata:
Didn't we have that originally even a few years back? :-)

It will certainly simplify the Motoko test setup once we can install canisters in drun and ic-ref-run with the same canister IDs, pinging @ggreif

dsarlis:
> Didn't we have that originally even a few years back? :-)
> 
> It will certainly simplify the Motoko test setup once we can install canisters in drun and ic-ref-run with the same canister IDs, pinging @ggreif

If you say so... I don't remember having this actually.

Dfinity-Bjoern (approved):
LGTM

mraszyk:
> Didn't we have that originally even a few years back? :-)
> 
> It will certainly simplify the Motoko test setup once we can install canisters in drun and ic-ref-run with the same canister IDs, pinging @ggreif

It should be possible to get the same canister ids without this feature, too. It used to be the case that ic-ref would yield different canister IDs because it was converting integers to canister IDs via little endian instead of big endian.

nomeata:
> It used to be the case that ic-ref would yield different canister IDs because it was converting integers to canister IDs via little endian instead of big endian.

In a way that was by design: I often made ic-ref behave intentionally different, but still within the spec, so that applications tested with it are less likely to depend on unspecified behavior. But it comes at a cost of course.

PR #130, opened 2022-12-22
canister creation can fail after providing specified_id

mraszyk:
If a canister ID C is taken by a canister created with specified_id = Some C, then subsequent canister creation with specified_id = None can fail if the last_generated_canister_id is equal to C which is already taken.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1764,7 +1764,7 @@ NOTE: Currently, the Internet Computer mainnet only supports URLs that resolve t
 [#ic-provisional_create_canister_with_cycles]
 === IC method `provisional_create_canister_with_cycles`
 
-As a provisional method on development instances, the `provisional_create_canister_with_cycles` method is provided. It behaves as `create_canister`, but initializes the canister’s balance with `amount` fresh cycles (using `DEFAULT_PROVISIONAL_CYCLES_BALANCE` if `amount = null`). If `specified_id` is provided, the canister is created under this id.
+As a provisional method on development instances, the `provisional_create_canister_with_cycles` method is provided. It behaves as `create_canister`, but initializes the canister’s balance with `amount` fresh cycles (using `DEFAULT_PROVISIONAL_CYCLES_BALANCE` if `amount = null`). If `specified_id` is provided, the canister is created under this id. Note that canister creation using `create_canister` or `provisional_create_canister_with_cycles` with `specified_id = null` can fail after calling `provisional_create_canister_with_cycles` with provided `specified_id`.
 
 Cycles added to this call via `ic0.call_cycles_add128` are returned to the caller.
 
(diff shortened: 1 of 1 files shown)


spec/index.adoc
```diff
@@ -1764,7 +1764,7 @@ NOTE: Currently, the Internet Computer mainnet only supports URLs that resolve t
 [#ic-provisional_create_canister_with_cycles]
 === IC method `provisional_create_canister_with_cycles`
 
-As a provisional method on development instances, the `provisional_create_canister_with_cycles` method is provided. It behaves as `create_canister`, but initializes the canister’s balance with `amount` fresh cycles (using `DEFAULT_PROVISIONAL_CYCLES_BALANCE` if `amount = null`). If `specified_id` is provided, the canister is created under this id.
+As a provisional method on development instances, the `provisional_create_canister_with_cycles` method is provided. It behaves as `create_canister`, but initializes the canister’s balance with `amount` fresh cycles (using `DEFAULT_PROVISIONAL_CYCLES_BALANCE` if `amount = null`). If `specified_id` is provided, the canister is created under this id. Note that canister creation using `create_canister` or `provisional_create_canister_with_cycles` with `specified_id = null` can fail after calling `provisional_create_canister_with_cycles` with provided `specified_id`. In that case, canister creation should be retried.
```
Dfinity-Bjoern, line 1767: My first intuition was that this is an ugly wart but it is acceptable since it's not going to appear in prod but only in the local canister development environment, so it seemed acceptable. But then, I realized that a common case of using this feature maybe the installation of the NNS canisters, which happen to have the same canister ids that the canisters in the local canister development environment also get (since they're the lowest `nat64`). In that sense: I would be fine accepting this spec change, but are we sure we want to have that behavior? We may see lots of problem reports from developers with potentially non-reproducible errors.

mraszyk: @dsarlis Could you please explain your reasoning why you prefer that behavior?

dsarlis: What Bjoern describes will not happen in the case where someone uses dfx, as it will be started with the right option that separates the ranges used for canisters with specified ids and the canisters that get auto-assigned ids (and therefore we won't have conflicts). This _can_ be a problem if you try to create canisters with specified ids when the ranges are **not** properly assigned. E.g. if someone is trying to use this on a testnet or in a system test without turning on the right option. Even if you turned it on, we could still have issues as we designed the solution for the dfx use case only but it can be extended to work for testnets and/or system tests if we wish without clashes.

In short, getting an error for a canister creation after you have created a canister with `specified_id` would happen only in test scenarios where someone goes outside the designed norm. Since we're talking about test scenarios anyway, getting an error and retrying (you'll eventually get the next free id and succeed) from the client side is very much acceptable (and preferable to questionable alternatives of adding special error handling on the replica production code for test scenarios).

Dfinity-Bjoern (approved):

PR #131, opened 2022-12-23
RUN-509: Enable Canister Timers in pre-upgrade

dfinity-berestovskyy:
Canister Timers are allowed in post-upgrade and init and it is easy to assume that they are also okay in pre-upgrade.


mraszyk (approved):
LGTM

Dfinity-Bjoern (approved):
LGTM

PR #139, opened 2023-02-16
Make replicated query semantics more explicit

yvonneanne:
Several people have asked about whether send a call request type to a query method changes the canister state. In this PR I try to make it more explicit that the state will not be changed.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -562,7 +562,7 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
 
 The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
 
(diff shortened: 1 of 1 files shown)


spec/index.adoc
```diff
@@ -562,7 +562,7 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
 
 The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
 
-This request type can _also_ be used to call a query method. A user may choose to go this way, instead of via the faster and cheaper <<http-query>> below, if they want to get a _certified_ response.
+This request type can _also_ be used to call a query method. A user may choose to go this way, instead of via the faster and cheaper <<http-query>> below, if they want to get a _certified_ response. Note that the canister change will not be changed by sending a call request type for a query method.
```
ggreif, line 565:
```suggestion
This request type can _also_ be used to call a query method. A user may choose to go this way, instead of via the faster and cheaper <<http-query>> below, if they want to get a _certified_ response. Note that the canister state will not be changed by sending a call request type for a query method.
```

nomeata (approved):

bogwar (approved):

PR #140, opened 2023-02-21
enable calling stable memory system API in WASM start function

mraszyk:
The mainnet's WASM execution engine is going to convert calls to stable memory system API into native memory operations which makes it challenging to prevent calls to stable memory system API in the WASM start function. Since we do not see any important reason to prevent such calls in the WASM start function, we adjust the specification to enable them.

### DIFF
--- a/spec/ic0.txt
+++ b/spec/ic0.txt
@@ -45,14 +45,14 @@ ic0.call_cycles_add : (amount : i64) -> ();                                 // U
 ic0.call_cycles_add128 : (amount_high : i64, amount_low: i64) -> ();        // U Ry Rt T
 ic0.call_perform : () -> ( err_code : i32 );                                // U Ry Rt T
 
-ic0.stable_size : () -> (page_count : i32);                                 // *
-ic0.stable_grow : (new_pages : i32) -> (old_page_count : i32);              // *
-ic0.stable_write : (offset : i32, src : i32, size : i32) -> ();             // *
-ic0.stable_read : (dst : i32, offset : i32, size : i32) -> ();              // *
-ic0.stable64_size : () -> (page_count : i64);                               // *
-ic0.stable64_grow : (new_pages : i64) -> (old_page_count : i64);            // *
-ic0.stable64_write : (offset : i64, src : i64, size : i64) -> ();           // *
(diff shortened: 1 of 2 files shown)


nomeata:
There were some good Wasmy reasons from @rossberg why we usually don't allow API access from the start function. @crusso, do you remember?

 Maybe stable memory API is fine, since its supposed to be a stop gap for direct memory instructions. 

Why is canister_init not good enough, though?

mraszyk:
> Maybe stable memory API is fine, since its supposed to be a stop gap for direct memory instructions.

I didn't get this comment.
 
> Why is canister_init not good enough, though?

With an upcoming feature, it'd be more difficult to prevent such calls to stable memory API in the start function so let's just allow them.

nomeata:
Sorry for being vage. I think the argument was something like that the Wasm standard makes no guarantee that imported functions are available and fully working before the start function is run, and I think if you have multiple wasm modules that mutually import each other, then some start function has to be run first.

It's mostly theoretical, I guess, in our environment we know that the system API works already in start.

Maybe don't worry then :-)

adambratschikaye:
I looked into this a little bit - it does seem to be true that the spec leaves it to the embedder to decide how imports are resolved, so it would be possible for a given embedder to allow mutual imports between two modules. But on the other hand, the spec does specify that imports can be called from the start function (see [these tests](https://github.com/WebAssembly/spec/blob/2558d66b716b8b85e028a595a2a51a3a9063c0f7/test/core/start.wast#L80-L95)). It also looks like Wasmtime itself wasn't set up to allow mutual imports - when instantiating a module, it's expected that all it's imports are provided (see [here](https://docs.wasmtime.dev/api/wasmtime/struct.Instance.html#method.new) or [here](https://docs.wasmtime.dev/api/wasmtime/struct.Linker.html#method.instantiate)). I think you could still hack it together by instantiating separate Wasmtime `Instance`s and then providing implementations that call each other, but I don't really think that's something we would want to do.

Dfinity-Bjoern (approved):
Discussed in the interface spec meeting, after checking the above arguments and taking into account that there are other `ic0` imports that are callable in `s`, this does not seem to change anything fundamental.

PR #141, opened 2023-02-28
feat: Add a system api call to check whether a given principal is a controller

dsarlis:
This PR adds a new system api call to allow a canister to check whether a given principal is one of its own controllers. This is a useful check for canisters and it has come up in various contexts, for example when canisters want to perform some access control on their methods based on whether the caller is a controller.

### DIFF
--- a/spec/ic0.txt
+++ b/spec/ic0.txt
@@ -62,6 +62,7 @@ ic0.data_certificate_copy : (dst: i32, offset: i32, size: i32) -> ();       // *
 ic0.time : () -> (timestamp : i64);                                         // *
 ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U Ry Rt C T
 ic0.performance_counter : (counter_type : i32) -> (counter : i64);          // * s
+ic0.is_controller: (src: i32, size: i32) -> ( result: i32);                 // * s
 
 ic0.debug_print : (src : i32, size : i32) -> ();                            // * s
 ic0.trap : (src : i32, size : i32) -> ();                                   // * s
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -2379,6 +2379,7 @@ Timestamp = Nat;
 CanisterVersion = Nat;
 Env = {
   time : Timestamp;
+  controllers : List Principal;
   global_timer : Nat;
   balance : Nat;
   freezing_limit : Nat;
@@ -2805,6 +2806,7 @@ Conditions::
       S.canisters[E.content.canister_id] ≠ EmptyCanister
       Env = {
         time = S.time[E.content.canister_id];
+        controllers = S.controllers[E.content.canister_id];
         global_timer = S.global_timer[E.content.canister_id];
         balance = S.balances[E.content.canister_id]
         freezing_limit = freezing_limit(S, E.content.canister_id);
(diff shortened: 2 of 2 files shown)


spec/index.adoc
```diff
@@ -4157,6 +4169,7 @@ ExecutionState = {
   wasm_state : WasmState;
   params : Params;
   response : NoResponse | Response;
+  controllers: List Principal;
```
dsarlis, line 4172: I wasn't sure if I need to somehow set this to the actual list of controllers or if this is fine given this is a model. Please let me know if I need to add something more.

Dfinity-Bjoern:
Discussed in the interface spec meeting; we weren't sure about possible use cases for a more generic API that allows to get all controllers. @bogwar will get input from the developer forum.

dsarlis:
> Discussed in the interface spec meeting; we weren't sure about possible use cases for a more generic API that allows to get all controllers. @bogwar will get input from the developer forum.

I'm fine with asking but let's make sure we ask the right question. What I mean: if we ask, do you want a more generic API that gives you the list of controllers, nobody is gonna say no (people will always say give me more basically). But if we ask: between the two options, is the `is_controller` enough for almost everything and what does a more generic API add (if anything), that's a better way to phrase this imo. The more generic API is also more complicated, it will need 3 methods to properly return a "list" of controllers, so we should have really strong use cases to go for it. cc @bogwar

bogwar:
Definitely — thanks for the suggestion Dimitris

spec/ic0.txt
```diff
@@ -62,6 +62,7 @@ ic0.data_certificate_copy : (dst: i32, offset: i32, size: i32) -> ();       // *
 ic0.time : () -> (timestamp : i64);                                         // *
 ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U Ry Rt C T
 ic0.performance_counter : (counter_type : i32) -> (counter : i64);          // * s
+ic0.is_controller: (src: i32, size: i32) -> ( result: i32);                 // * s
```
mraszyk, line 65: Is it really gonna be available in the WASM _start_ function? @dsarlis

spec/index.adoc
mraszyk: Fixed.

spec/ic0.txt
dsarlis: I was thinking it's a general system api that can be useful (similar to performance counter or retrieving your balance) so I didn't see a reason not to. I don't feel strongly though, if you believe we don't need it, we can drop start here.

mraszyk: I'm fine either way, too, just wanted to make sure it's a deliberate choice.

bogwar (approved):
Martin has fixed the "controllers in the Env" issue, so it's good to go

spec/index.adoc
```diff
@@ -1462,6 +1462,17 @@ The system resets the counter at the beginning of each <<entry-points>> invocati
...
+[#system-api-controller-check]
+=== Controller check
+
+The canister can check whether a given principal is one of its controllers.
+
+`+ic0.is_controller : (src : i32, size: i32) -> (result: i32)+`
+
+Checks whether the principal identified by `src`/`size` is one of the controllers of the canister. If yes, then a value of 1 is returned, otherwise a 0 is returned. It can be called multiple times.
+
+This system call traps if `src+size` exceeds the size of the WebAssembly memory.
```
mraszyk, line 1474:
```suggestion
This system call traps if `src+size` exceeds the size of the WebAssembly memory or the principal identified by `src`/`size` is not a valid binary encoding of a principal.
```

PR #142, opened 2023-02-28
feat: specify representation-independent hash for integers

roman-kashitsyn:
This change extends the specification of representation-independent hashing to handle signed integers.

Motivation: DFINITY's ICRC-1 ledger hashes blocks and transactions using the representation-independent hash from the Interface Spec. The new approve transactions allow negative amounts, so we must extend the spec to be able to hash such transactions.


Dfinity-Bjoern (approved):
Looks good, thanks!

PR #143, opened 2023-03-03
IC-1462 extend the error space for update requests

rumenov:
More context can be found at https://dfinity.atlassian.net/browse/IC-1462.

To summarize, the current problem we are facing is that IC platform errors are mixed with canister owner errors which causes confusion among engineers how different errors should be handled.

This proposal is an attempt to have have a clear separation between canister owner errors and IC platform errors.

### DIFF
--- a/spec/ic.did
+++ b/spec/ic.did
@@ -116,8 +116,8 @@ service ic : {
     headers: vec http_header;
     body : opt blob;
     transform : opt record {
-      function : func (record {response : http_response; context : blob}) -> (http_response) query;
-      context : blob
+      function : func (record {response : http_response; arg : blob}) -> (http_response) query;
+      arg : blob
     };
   }) -> (http_response);
 
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -530,7 +530,11 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
 * `method_name` (`text`): Name of the canister method to call
 * `arg` (`blob`): Argument to pass to the canister method
 
-The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
+The HTTP response to this request can have the following responses:
+* 202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use <<http-read-state,`read_state`>> to determine the status of the call.
+* 200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the user error. This return type
+is used when the error can be handled by the canister owner (e.g. when a canister is out of cycles) 
+* 4xx or 5xx HTTP error. 
 
 This request type can _also_ be used to call a query method. A user may choose to go this way, instead of via the faster and cheaper <<http-query>> below, if they want to get a _certified_ response.
 


spec/index.adoc
```diff
@@ -530,7 +530,11 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
 * `method_name` (`text`): Name of the canister method to call
 * `arg` (`blob`): Argument to pass to the canister method
 
-The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
+The HTTP response to this request can have the following responses:
+* 202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use <<http-read-state,`read_state`>> to determine the status of the call.
+* 200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the user error. This return type
+is used when the error can be handled by the canister owner (e.g. when a canister is out of cycles)
```
mraszyk, line 536: The previous sentence talks about "user" error (sounds like it's triggered by the request sent by the user) while the last sentence talks about "canister owner" to fix the error (although canister running out of cycles needn't be only fixed by the controller). Could you please clarify what types of error are expected there? @rumenov

rumenov: I am not an expert [on execution errors](https://github.com/dfinity/ic/blob/master/rs/http_endpoints/public/src/common.rs#L80) but I believe most of the issues here should be returned as OK 200

@dsarlis any suggestions how to improve the wording here ?

dsarlis: If we do this, wouldn't then it mean that 4xx and 5xx can be returned only when the request is malformed (or something along those lines) or the replica is unavailable respectively? And therefore, any IC specific error would be returned via 200 with non-empty body containing the details of the error?

spec/ic.did
```diff
@@ -116,8 +116,8 @@ service ic : {
     headers: vec http_header;
     body : opt blob;
     transform : opt record {
-      function : func (record {response : http_response; context : blob}) -> (http_response) query;
-      context : blob
+      function : func (record {response : http_response; arg : blob}) -> (http_response) query;
+      arg : blob
```
dsarlis, line 120: Why this change?

spec/index.adoc
dsarlis: Oh wait, too early in the morning, I misread the bullet points. I think we can simply say something like:
```
200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the IC specific error encountered.
```
and completely drop the last part. Some of the errors could be handled only by the canister owner or they could be handled by the client too, so trying to be more specific doesn't help much.

dsarlis: Also, I think we should specify the body in the case where an error is returned. I'd expect it would be quite similar to the error case for a query call. So, something like:

```
The body is a CBOR map with the following fields:
* `status` (`text`): `rejected`
* `reject_code` (`nat`): The reject code (see <<reject-codes>>).
* `reject_message` (`text`): a textual diagnostic message.
* `error_code` (text): an optional implementation-specific textual error code (see <<error-codes>>).
```
(maybe we can skip `status` in this case though as it would only be non-empty in the error case anyway).

spec/ic.did
rumenov: Good question :) I don't actually know why it showed up. Maybe it is some old change that slipped into my branch.

spec/index.adoc
rumenov: @dsarlis can you please take another look ?

spec/index.adoc
```diff
@@ -560,7 +560,15 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
...
 
-The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
+The HTTP response to this request can have the following responses:
+
+* 202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use <<http-read-state,`read_state`>> to determine the status of the call.
+* 200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the IC specific error encountered. The body is a CBOR map with the following fields:
+** `reject_code` (`nat`): The reject code (see <<reject-codes>>).
+** `reject_message` (`text`): a textual diagnostic message.
+** `error_code` (text): an optional implementation-specific textual error code (see <<error-codes>>).
+* 4xx HTTP status when the is error caused by the client.
```
mraszyk, line 570:
```suggestion
* 4xx HTTP status when the error is caused by the client.
```

spec/index.adoc
```diff
@@ -560,7 +560,15 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
...
 * `arg` (`blob`): Argument to pass to the canister method
 
-The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
+The HTTP response to this request can have the following responses:
+
+* 202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use <<http-read-state,`read_state`>> to determine the status of the call.
+* 200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the IC specific error encountered. The body is a CBOR map with the following fields:
+** `reject_code` (`nat`): The reject code (see <<reject-codes>>).
+** `reject_message` (`text`): a textual diagnostic message.
+** `error_code` (text): an optional implementation-specific textual error code (see <<error-codes>>).
```
mraszyk, line 569:
```suggestion
** `error_code` (`text`): an optional implementation-specific textual error code (see <<error-codes>>).
```

spec/index.adoc
dsarlis: Lgtm!

dsarlis (approved):

mraszyk (approved):
LGTM

spec/index.adoc
```diff
@@ -560,7 +560,15 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
...
-The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
+The HTTP response to this request can have the following responses:
+
+* 202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use <<http-read-state,`read_state`>> to determine the status of the call.
+* 200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the IC specific error encountered. The body is a CBOR map with the following fields:
+** `reject_code` (`nat`): The reject code (see <<reject-codes>>).
+** `reject_message` (`text`): a textual diagnostic message.
+** `error_code` (`text`): an optional implementation-specific textual error code (see <<error-codes>>).
+* 4xx HTTP status when the error is caused by the client.
+* 5xx HTTP status when the server is aware that it has encountered an error or is otherwise incapable of performing the request.
```
alin-at-dfinity, line 571: Personal opinion (feel free to ignore): I'dd add some examples and some clarifications for the 4xx and 5xx cases. Something along the lines of (although I'm not sure the examples are entirely accurate):

```suggestion
* 4xx HTTP status for client errors (e.g. malformed request or canister not found). Retrying the request will likely have the same outcome.
* 5xx HTTP status when the server has encountered an error or is otherwise incapable of performing the request. The request might succeed if retried at a later time.
```

alin-at-dfinity (approved):
LGTM, thanks.

roman-kashitsyn (approved):

Dfinity-Bjoern:
The errors that we want to return as 200s in the future – were they returned as 202 or as 4xx/5xx? If we returned a 4xx/5xx before and now we signal errors through a 200 with a CBOR, isn't that a backward-incompatible change? Meaning: shouldn't we rather do this as an `/api/v3/...` HTTP call?

dsarlis:
> The errors that we want to return as 200s in the future – were they returned as 202 or as 4xx/5xx? If we returned a 4xx/5xx before and now we signal errors through a 200 with a CBOR, isn't that a backward-incompatible change? Meaning: shouldn't we rather do this as an `/api/v3/...` HTTP call?

Yeah I think that's a good point. My understanding is that certain 5xx errors would become 200 + CBOR body, so we should make it in a backwards compatible way. If you think `/api/v3/...` is the best way, I am all for it.

cc @rumenov for his opinion as the proposer of the change.

rumenov:
> The errors that we want to return as 200s in the future – were they returned as 202 or as 4xx/5xx? If we returned a 4xx/5xx before and now we signal errors through a 200 with a CBOR, isn't that a backward-incompatible change? Meaning: shouldn't we rather do this as an `/api/v3/...` HTTP call?

There was actually a breaking change ~6 months ago which moved 202 codes to 5xx/4xx. Now we want partially to move those 5xx/4xx to 200.

Not sure if it worth doing 'api/v3/..' if you think we should do it. What is the correct way to do this ? I assume there must be some boundary node changes as well.

spec/index.adoc
```diff
@@ -560,7 +560,15 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
 * `method_name` (`text`): Name of the canister method to call
 * `arg` (`blob`): Argument to pass to the canister method
 
-The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
+The HTTP response to this request can have the following responses:
+
+* 202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use <<http-read-state,`read_state`>> to determine the status of the call.
+* 200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the IC specific error encountered. The body is a CBOR map with the following fields:
```
sesi200, line 566: I'm not a fan of HTTP 2xx indicating an error. 2xx generally indicate success - see also [here](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes).

spec/index.adoc
```diff
@@ -560,7 +560,15 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
...
+** `reject_code` (`nat`): The reject code (see <<reject-codes>>).
+** `reject_message` (`text`): a textual diagnostic message.
+** `error_code` (`text`): an optional implementation-specific textual error code (see <<error-codes>>).
+* 4xx HTTP status for client errors (e.g. malformed request or canister not found). Retrying the request will likely have the same outcome.
+* 5xx HTTP status when the server has encountered an error or is otherwise incapable of performing the request. The request might succeed if retried at a later time.
```
sesi200, line 571: Would this include e.g. `not enough cycles`? In this case, I'd like to see specific documented cases that can be 
matched against, similar to the above `error_code` because  some errors can be treated in some way by the developer.

On second thought, why not also allow this for any non-200 error code? There will always be use cases for differentiating between different error conditions, and only using HTTP status codes feels too limiting

spec/index.adoc
rumenov: It was success in the sense the request reached the execution environment. Compared to all alternatives this is maybe the least confusing we found given we want to distinguish between canister/execution errors and IC platform errors.

spec/index.adoc
rumenov: `not enough cycles` would be 2xx error

> On second thought, why not also allow this for any non-200 error code? There will always be use cases for differentiating between different error conditions, and only using HTTP status codes feels too limiting

Not sure what you exactly mean by this .

spec/index.adoc
sesi200: `execution pre-processing error` is this something like failed to decode candid or rejected by guard method? 

Anything above 5xx has no canonical meaning, why not just use 7xx for it? If I as a dev see 200, I am sure everything worked out fine, which it obviously didn't

sesi200: I just saw `All valid status codes are within the range of 100 to 599, inclusive.` in the RFC... blegh.

Looking through the RFC, I feel like for a pre-processing error a lot would be more appropriate than 200. The first sentence of [200](https://www.rfc-editor.org/rfc/rfc9110.html#name-200-ok) is `The 200 (OK) status code indicates that the request has succeeded.`, which is in direct conflict to a pre-processing **error** IMO. [202](https://www.rfc-editor.org/rfc/rfc9110.html#name-202-accepted) feels better, but still it says `accepted for processing, but the processing has not been completed` which is partially true, but processing _has_ completed, with an error. Some codes in the 4xx range seem more valid on first glance (e.g. [406 Not acceptable](https://www.rfc-editor.org/rfc/rfc9110.html#name-406-not-acceptable) or [412 Precondition failed](https://www.rfc-editor.org/rfc/rfc9110.html#name-412-precondition-failed)), but they are talking about headers, which also seems wrong for this case. Maybe [422 Unprocessable content](https://www.rfc-editor.org/rfc/rfc9110.html#name-422-unprocessable-content) would be valid - it talks about `unable to process the contained instructions`.

spec/index.adoc
sesi200: Some of the 4xx or 5xx status codes have multiple root causes, some of which are recoverable for the client, but at the moment they can only be differentiated with the error message, which may change at any point in time. If we had an error code for these too, I'd be much more comfortable to implement automatic recovery processes.

spec/index.adoc
rumenov: > If I as a dev see 200, I am sure everything worked out fine, which it obviously didn't

Even today, the spec explicitly says that receiving 200 or even 202 means nothing. You should pool the state to make sure your update call made it through. Given the async nature of the protocol I don't think we have much choice here.

spec/index.adoc
rumenov: ideally figuring out if it is recoverable or not the client should use only the status code of the 4xx or 5xx error and not the body. This is part of the HTTP spec to.

spec/index.adoc
nathanosdev: I agree with Severin that 2xx codes should not be used for platform errors. Aside from developer confusion, it also reduces compatibility with existing (and future) HTTP aware tooling that understands HTTP codes. Differentiating between platform errors and canister errors using the body fields sounds better, in my opinion.

alin-at-dfinity: There is a difference to be made between the IC/subnet/replica (the application, if you want) and the HTTP API used to access it. The HTTP status codes (whether 2xx, 4xx or 5xx) only refer to the API, which acts as a proxy to the application. To use the envelope and content analogy at the top of the *Call* section, the status codes refer to the delivery of the envelope.

The alternative (and what we have now, for the most part) is to try and make the HTTP API entirely transparent. But that means either:
* Using the same status code to report on both API and IC errors. E.g. using 404 for "no such API endpoint", "no such canister" and "no such canister method". Similarly, and more confusingly, returning 5xx errors both when the API encounters a server error; when there's an internal replica/sandbox error; and when the canister logic produces an error or panics.
* Or designating some made-up 4xx and 5xx status codes to reflect replica or canister errors. At which point (i) they are not immediately understandable to an HTTP client; and (ii) they are not all that different from returning a 2xx status and a body clarifying the error. Furthermore, a 200 or 202 response to a `call` endpoint by no means guarantees that the subnet will ever induct the ingress message into a block. I.e. we already have the case where a replica replies with 200 or 202, only for the ingress message to disappear without a trace.

Maybe part of the confusion also stems from explicitly stating that 202 stands for success and 200 for error. It would probably be cleaner and clearer to say that a 2xx response means that the API delivered the envelope to the replica and the outcome, if any, is in the body of the response.

sesi200: > Even today, the spec explicitly says that receiving 200 or even 202 means nothing. [...] Given the async nature of the protocol I don't think we have much choice here.

Fair enough. I still don't like how 2xx means nothing, but I guess then we have found the least bad option.

spec/index.adoc
sesi200: Then can we please guarantee that one cause of error = one status code? Otherwise it's back to matching on the body since there are no guarantees - having an `error_code` field would make this a lot easier.

One use case I have in mind: a non-controller asks for the canister status. I want to show the user an explanation of controllers / nice error message. On the replica, I can match on HTTP 403, but on ic-ref, it returns 400, where I have to match on the body for "wrong sender" because anything could return a 400.

Similarly, AFAIK 503 only happens if a query happens to hit a replica that is behind, therefore I should retry my request, but who knows? It could be a problem where retrying does not help

rikonor (approved):
No objection from BN and Trust teams. 👍

sesi200 (approved):
LGTM from CDK and Agent side

rumenov:
@Dfinity-Bjoern, can you please take one more look

We discussed the changes with the BN (@rikonor ) and SDK (@sesi200 ) and we didn't flag any problems with the proposed changes.

spec/index.adoc
```diff
@@ -560,7 +560,15 @@ In order to call a canister, the user makes a POST request to `/api/v2/canister/
...
 
-The HTTP response to this request has an empty body and HTTP status 202, or a HTTP error (4xx or 5xx). Paranoid agents should not trust this response, and use <<http-read-state,`read_state`>> to determine the status of the call.
+The HTTP response to this request can have the following responses:
+
+* 202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use <<http-read-state,`read_state`>> to determine the status of the call.
+* 200 HTTP status with non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the IC specific error encountered. The body is a CBOR map with the following fields:
+** `reject_code` (`nat`): The reject code (see <<reject-codes>>).
+** `reject_message` (`text`): a textual diagnostic message.
+** `error_code` (`text`): an optional implementation-specific textual error code (see <<error-codes>>).
+* 4xx HTTP status for client errors (e.g. malformed request or canister not found). Retrying the request will likely have the same outcome.
```
mraszyk, line 570: > Retrying the request will likely have the same outcome.

I see http status codes like 429 Too Many Requests in the overview of http status codes which should be possible to retry. So I wonder we could rephrase this sentence, e.g., by listing the status codes that do not make sense to retry?

@rumenov

spec/index.adoc
mraszyk, line 566: Could we please list the cases explicitly? Currently, the reference implementation and its test suite does not handle this so I would like to know what to change there?

@rumenov

rumenov: I don't think this is reasonable. The [list is ever](https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/http_endpoints/public/src/common.rs?L80) growing.

rumenov: > I see. Would they all become 200? Then it could make sense to list which 4xx/5xx are not transient.

I believe they all become 200. @dsarlis to confirm

spec/index.adoc
rumenov: I don't think it makes sense. Then we will be repeating the HTTP protocol standard.

mraszyk: Sure, I don't want to repeat the HTTP protocol standard. Instead I want to explicitly list status codes produced by the IC due to IC-specific failures (canister not found is an example that likely has the same outcome unless the canister is created in the meantime as thus should be listed as a case to which the sentence "Retrying the request will likely have the same outcome" applies).

spec/index.adoc
dsarlis: All these errors would become 200 (plus the cbor body that has the IC error details).

> Then it could make sense to list which 4xx/5xx are not transient.

I'm not sure I understood this part.

spec/index.adoc
rumenov: Given this - https://github.com/dfinity/interface-spec/pull/143#discussion_r1151721927, there won't be any execution errors that map to 4xx/5xx errors.

mraszyk: Perfect! I've made a suggestion to drop "canister not found" and refine the sentence on retrying.

spec/index.adoc
mraszyk: > Then it could make sense to list which 4xx/5xx are not transient.

Given that all IC-specific errors would become 200, my comment is irrelevant (I wanted to list IC-specific errors - like canister not found - that still return 4xx/5xx instead of 200).

spec/index.adoc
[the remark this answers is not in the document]
alin-at-dfinity: The suggestion seems too generic: "doing X might fail".

If anything, I'd rather mention 429 explicitly. Or say something along the lines of "The client should not retry the request without modification" (in some cases, said modification would be to e.g. use a different URL or authenticate; in other cases, it would simply mean "wait until the conditions are right").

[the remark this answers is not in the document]
mraszyk: I mentioned 429 explicitly now. I did not go for the sentence "The client should not retry the request without modification" because with 429 you can make the very same request again.

rumenov:
@Dfinity-Bjoern are we waiting for anything else on this MR ? I think everything is resolved by now.

mraszyk:
> @Dfinity-Bjoern are we waiting for anything else on this MR ? I think everything is resolved by now.

We might want to wait with merging this until the code changes are ready to be merged.

Dfinity-Bjoern (approved):
Looks good now!

levifeldman:
> To summarize, the current problem we are facing is that IC platform errors are mixed with canister owner errors which causes confusion among engineers how different errors should be handled.

> This proposal is an attempt to have have a clear separation between canister owner errors and IC platform errors.

I thought the way to separate between an IC platform error and a canister owner error is that reject codes 1,2, and 3 are for IC platform errors and reject codes 4 and 5 are for canister owner errors. 

What additional benefits does this bring for separating between canister owner errors and IC platform errors? 

If there is an additional separation that this brings, should it be done for query calls too?

rumenov:
@levifeldman,

First I want to clarify, that the spec change is only relevant when the [canister_inspect_message](https://internetcomputer.org/docs/current/references/ic-interface-spec#system-api-inspect-message) fails.

This said, today when `canister_inspect_message` fails we return a HTTP error code as part of the response (no reject codes are being included as part of the body). The spec change suggests that we return a 200 with a body that contains the reject code, similar to what happens when a query is being executed. 


I am not sure I completely understand 

> It seems there can still be an IC platform error - reject codes 1, 2, and 3 - in the read_state call, after this call returns 202, and the spec as written in this PR still allows for canister errors - reject codes 4 and 5 - to be returned from a 200 response.

However, please not that the spec change is under the `call` section and not under the `read_state` section.

> If there is an additional separation that this brings, should it be done for query calls too?

The spec change actually adopts the exact same behaviour as what we have for query calls.

dsarlis:
To add a bit to Rosti's reply: I think there might be a confusion because of the wording used.

The goal was to make it more clear when we have IC platform errors, in the sense of replica code hitting some unexpected error case (and which are important for the operators of the network but not end users) and IC protocol-level errors, i.e. errors that are intrinsic to IC behaviour that is captured by the reject codes in the spec (and which is more useful to end users).

As Rosti explained, the change actually makes things more consistent to query calls by returning the IC protocol specific errors inside the body of the HTTP reply.

levifeldman:
Hi @rumenov, I see, thanks! Before there was no way to return the reject code when canister_inspect_message fails and now with this change there is a way. Cool!

I was thinking that IC-platform-errors were the reject codes 1,2, and 3, and the canister-owner-errors were reject codes 4 and 5. Thanks @dsarlis for the clarification that we are talking about a separation between the IC-protocol-level errors which are the reject codes as a whole, and between the network-operator replica errors. 

Thank you both!

levifeldman:
Hi @rumenov, I see, thanks! Before there was no way to return the reject code when canister_inspect_message fails and now with this change there is a way. Cool!

I was thinking that IC-platform-errors were the reject codes 1,2, and 3, and the canister-owner-errors were reject codes 4 and 5. Thanks @dsarlis for the clarification that we are talking about a separation between the IC-protocol-level errors which are the reject codes as a whole, and between the network-operator replica errors.

Thank you both.

PR #144, opened 2023-03-06
specify composite queries

mraszyk:
This PR introduces a new public entry point of canisters: composite queries.


spec/index.adoc
```diff
@@ -4015,30 +4071,120 @@ Conditions::
...
+              let R'' = F'(W')
+              if R'' = Trap trap''
+              then Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles - trap''.cycles_used)
+              else if R'' = Return {new_state = W''; new_calls = Calls''; response = Response''; cycles_used = Cycles_used''}
+              then
+                 W := W''
+                 Cycles := Cycles - Cycles_used''
+                 if Response'' = NoResponse
+                 then
+                    Calls := Calls'' · Calls // DFS call graph traversal
```
nomeata, line 4160: Do we want to tie down the precise order in which calls are handles, or do we want to underspecify this, so that the implementation can make smart shoices (e.g. less context switches between different canisters), and that it’s forward compatible to a world where composite queries work cross-subnet?

spec/index.adoc
```diff
@@ -4520,19 +4683,19 @@ The pseudo-code below does _not_ explicitly enforce the restrictions of which im
 
 ....
 ic0.msg_arg_data_size<es>() : i32 =
-  if es.context ∉ {I, U, Q, Ry, F} then Trap {cycles_used = es.cycles_used;}
+  if es.context ∉ {I, U, Q, CQ, Ry, F} then Trap {cycles_used = es.cycles_used;}
```
nomeata, line 4686: Unrelated, but: I wonder how much we gain from repeating the “context check” in every call in this section, compared to just stating once that “all functions trap if not called in an allowed context according to table XX”

mraszyk: We gain a bit more precision.

spec/index.adoc
```diff
@@ -4015,30 +4071,120 @@ Conditions::
...
+
+==== Composite query call
+
+Canister composite query calls to `/api/v2/canister/<ECID>/composite_query` can be executed directly. They can only be executed against canisters which have a status of `Running` and are also not frozen.
+
+During the execution of a composite query call, a certificate is provided to the canister that is valid, contains a current state tree (or “recent enough”; the specification is currently vague about how old the certificate may be), and reveals the canister’s <<system-api-certified-data,Certified Data>>.
+
+We define an auxiliary method that handles calls from composite query methods by performing a DFS call graph traversal.
+....
+composite_query_helper(S, Cycles, Caller, Canister_id, Method_name, Arg) =
```
mraszyk, line 4108: Do we need to track a stack of canisters called before to prevent reentrancy? Or is reentrancy allowed after all? @ulan

spec/index.adoc
nomeata: is there really more information there? Or just more precision in the sense that readers are reminded about this check?

mraszyk: > Or just more precision in the sense that readers are reminded about this check?

This is what I meant. Or rather that the pseudo-code is more self-contained, i.e., does not rely on additional assumptions outside the actual code.

spec/index.adoc
mraszyk: If wanted, I'd then replace
```
              let [Call; _] = Calls
```
by
```
              let Calls' · Call · Calls''  = Calls
              Calls := Calls' · Calls''
```

ulan: I guess for the spec it makes sense to underspecify the order.
In the implementation we will likely do the DFS traversal both now and in the future with cross-subnet queries.

spec/index.adoc
ulan: Yes, reentrancy is allowed and we converged to the design where a query that's reentering the canister gets a fresh copy of the canister state. In other words, it doesn't see the changes done by previous queries of the current query call graph.

mraszyk: Thank you for confirming! This means that nothing has to be changed here.

spec/index.adoc
mraszyk: Makes sense! Done.

ulan (commented):
Thanks a lot for writing the specification, Martin! I have some initial questions.

spec/index.adoc
```diff
@@ -67,11 +67,11 @@ To get some consistency in this document, we try to use the following terms with
...
 
-The public entry points of canisters are called _methods_. Methods can be declared to be either _update methods_ (state mutation is preserved) or _query methods_ (state mutation is discarded, no further calls can be made).
+The public entry points of canisters are called _methods_. Methods can be declared to be either _update methods_ (state mutation is preserved), _query methods_ (state mutation is discarded, no further calls can be made), or _composite query_ methods (state mutation is discarded, calls can be made to canisters on the same subnet).
 
 Methods can be _called_, from _caller_ to _callee_, and will eventually incur a _response_ which is either a _reply_ or a _reject_. A method may have _parameters_, which are provided with concrete _arguments_ in a method call.
 
-External calls can be update calls, which can call both kinds of methods, and query calls, which can _only_ call query methods.
+External calls can be update calls, which can call both kinds of methods, query calls, which can _only_ call query methods, and composite query calls, which can _only_ call query and composite query methods.
```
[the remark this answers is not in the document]
mraszyk: Good catch! Fixed.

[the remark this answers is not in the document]
mraszyk: I removed the additional HTTP endpoint. Now we have just the single notion of a query call and the two notions of a query method and a composite query method.

spec/index.adoc
```diff
@@ -481,11 +481,12 @@ The concrete mechanism that users use to send requests to the Internet Computer
...
+* At `/api/v2/canister/<effective_canister_id>/composite_query` the user can perform (synchronous, non-state-changing) composite query calls.
 * At `/api/v2/status` the user can retrieve status information about the Internet Computer.
 
 In these paths, the `<effective_canister_id>` is the <<textual-ids,textual representation>> of the <<http-effective-canister-id,_effective_ canister id>>.
 
-Requests to `/api/v2/canister/<effective_canister_id>/call`, `/api/v2/canister/<effective_canister_id>/read_state` and `/api/v2/canister/<effective_canister_id>/query` are POST requests with a CBOR-encoded request body, which consists of a authentication envelope (as per <<authentication>>) and request-specific content as described below.
+Requests to `/api/v2/canister/<effective_canister_id>/call`, `/api/v2/canister/<effective_canister_id>/read_state`, `/api/v2/canister/<effective_canister_id>/query`, and `/api/v2/canister/<effective_canister_id>/composite_query` are POST requests with a CBOR-encoded request body, which consists of a authentication envelope (as per <<authentication>>) and request-specific content as described below.
```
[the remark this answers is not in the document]
mraszyk: Done. (please see above for more comments)

spec/index.adoc
```diff
@@ -67,12 +67,12 @@ To get some consistency in this document, we try to use the following terms with
 
 We avoid the term “client”, as it could be the client of the Internet Computer or the client inside the distributed network that makes up the Internet Computer. Instead, we use the term _user_ to denote the external entity interacting with the Internet Computer, even if in most cases it will be some code (sometimes called “agent”) acting on behalf of a (human) user.
 
-The public entry points of canisters are called _methods_. Methods can be declared to be either _update methods_ (state mutation is preserved) or _query methods_ (state mutation is discarded, no further calls can be made).
+The public entry points of canisters are called _methods_. Methods can be declared to be either _update methods_ (state mutation is preserved), _query methods_ (state mutation is discarded, no further calls can be made), or _composite query_ methods (state mutation is discarded, calls can be made to canisters on the same subnet).
```
Dfinity-skaestle, line 70:
```suggestion
The public entry points of canisters are called _methods_. Methods can be declared to be either _update methods_ (state mutation is preserved), _query methods_ (state mutation is discarded, no further calls can be made), or _composite query_ methods (state mutation is discarded, calls can be made to query and composite query methods of canisters on the same subnet).
```

spec/index.adoc
```diff
@@ -922,8 +922,8 @@ In order for a WebAssembly module to be usable as the code for the canister, it
 * If it exports a function called `canister_inspect_message`, the function must have type `+() -> ()+`.
 * If it exports a function called `canister_heartbeat`, the function must have type `+() -> ()+`.
 * If it exports a function called `canister_global_timer`, the function must have type `+() -> ()+`.
-* If it exports any functions called `canister_update <name>` or `canister_query <name>` for some `name`, the functions must have type `+() -> ()+`.
-* It may not export both `canister_update <name>` and `canister_query <name>` with the same `name`.
+* If it exports any functions called `canister_update <name>`, `canister_query <name>`, or `canister_composite_query <name>` for some `name`, the functions must have type `+() -> ()+`.
```
Dfinity-skaestle, line 925: Did we want to mark the new functionality as experimental somewhere?

spec/index.adoc
```diff
@@ -1471,15 +1473,15 @@ For each canister, the IC keeps track of “certified data”, a canister-define
 +
 The canister can update the certified data with this call. The passed data must be no larger than 32 bytes. This can be used any number of times.
 
-When executing a query method via a query call (i.e. in non-replicated state), the canister can fetch a certificate that authenticates to third parties the value last set via `ic0.certified_data_set`.
+When executing a query or composite query method via a query call (i.e. in non-replicated state), the canister can fetch a certificate that authenticates to third parties the value last set via `ic0.certified_data_set`.
```
Dfinity-skaestle, line 1476: For composite queries, this is only currently supported in the root canister of the query call graph.

spec/index.adoc
```diff
@@ -1471,15 +1473,15 @@ For each canister, the IC keeps track of “certified data”, a canister-define
...
 
 * `+ic0.data_certificate_present : () -> i32+`
 +
 returns `1` if a certificate is present, and `0` otherwise.
 +
-This will return `1` when called from a query method when invoked via a query call.
+This will return `1` when called from a query or composite query method via a query call.
```
Dfinity-skaestle, line 1482:
```suggestion
This will return `1` when called from a query or composite query method via a query call. In composite queries, it is only `1` in the root canister of the composite query call tree, i.e. the initial call triggered from the user and not further calls triggered from other canisters.
```

spec/index.adoc
```diff
@@ -1471,15 +1473,15 @@ For each canister, the IC keeps track of “certified data”, a canister-define
...
 +
-This will return `0` if the query method is executed within replicated execution (e.g. when invoked via an update call or inter-canister call).
+This will return `0` for update methods and if a query method is executed within replicated execution (e.g. when invoked via an update call or inter-canister call).
```
Dfinity-skaestle, line 1484:
```suggestion
This will return `0` for update methods, if a query method is executed within replicated execution (e.g. when invoked via an update call or inter-canister call) and from within composite query calls for all but the initial method invoked by the user.
```

spec/index.adoc
```diff
@@ -2399,6 +2401,12 @@ QueryFunc = WasmState -> Trap { cycles_used : Nat; } | Return {
   response : Response;
   cycles_used : Nat;
 }
+CompositeQueryFunc = WasmState -> Trap { cycles_used : Nat; } | Return {
+  new_state : WasmState;
```
Dfinity-skaestle, line 2405: Should there be a `new_state` here?
We are not committing any changes from composite query calls.

spec/index.adoc
```diff
@@ -3996,9 +4016,74 @@ S with
...
+     Method_name ∈ dom(Mod.composite_query_methods) and
+     Cycles >= MAX_CYCLES_PER_MESSAGE
+  then
+     let W = S.canisters[Canister_id].wasm_state
+     let R = F(Arg, Caller, Env)(W)
+     if R = Trap trap
+     then Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles - trap.cycles_used)
+     else if R = Return {new_state = W'; new_calls = Calls; response = Response; cycles_used = Cycles_used}
+     then
+        W := W'
```
Dfinity-skaestle, line 4052: We don't apply changes from composite queries to the canister state.

spec/index.adoc
```diff
@@ -3996,9 +4016,74 @@ S with
...
+              if Cycles < MAX_CYCLES_PER_RESPONSE
+              then
+                 Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles) // composite query out of cycles
+              let F' = Mod.callbacks(Call.callback, Response', 0, Env, 0)
+              let R'' = F'(W')
+              if R'' = Trap trap''
+              then Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles - trap''.cycles_used)
+              else if R'' = Return {new_state = W''; new_calls = Calls''; response = Response''; cycles_used = Cycles_used''}
+              then
+                 W := W''
```
Dfinity-skaestle, line 4074: We don't commit `W''` here.

spec/index.adoc
```diff
@@ -3996,9 +4016,74 @@ S with
...
+           while Calls ≠ []
+           do
+              let Calls' · Call · Calls''  = Calls
+              Calls := Calls' · Calls''
+              if S.canister_subnet[Canister_id] ≠ S.canister_subnet[Call.callee]
+              then
+                 Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles) // calling to another subnet
+              let (Response', Cycles_used') = composite_query_helper(S, Cycles, Canister_id, Call.callee, Call.method_name, Call.arg)
+              Cycles := Cycles - Cycles_used'
+              if Cycles < MAX_CYCLES_PER_RESPONSE
```
Dfinity-skaestle, line 4065: There are some more checks in the code, e.g. a maximum call graph depth. Do we need to add them to the spec too?

spec/index.adoc
```diff
@@ -3996,9 +4016,74 @@ S with
...
+     then Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles - trap.cycles_used)
+     else if R = Return {new_state = W'; new_calls = Calls; response = Response; cycles_used = Cycles_used}
+     then
+        W := W'
+        Cycles := Cycles - Cycles_used
+        if Response = NoResponse
+        then
+           while Calls ≠ []
+           do
+              let Calls' · Call · Calls''  = Calls
```
Dfinity-skaestle, line 4058: The order of executing calls is different in the code. Does that matter?

spec/index.adoc
```diff
@@ -67,12 +67,12 @@ To get some consistency in this document, we try to use the following terms with
...
+The public entry points of canisters are called _methods_. Methods can be declared to be either _update methods_ (state mutation is preserved), _query methods_ (state mutation is discarded, no further calls can be made), or _composite query_ methods (state mutation is discarded, calls can be made to canisters on the same subnet).
 
 Methods can be _called_, from _caller_ to _callee_, and will eventually incur a _response_ which is either a _reply_ or a _reject_. A method may have _parameters_, which are provided with concrete _arguments_ in a method call.
 
-External calls can be update calls, which can call both kinds of methods, and query calls, which can _only_ call query methods.
-Inter-canister calls can also call both kinds of methods. Note that calls from a canister to itself also count as "inter-canister".
+External calls can be update calls, which can call update and query methods, and query calls, which can _only_ call query and composite query methods.
+Inter-canister calls can call update and query methods (just like update calls). Note that calls from a canister to itself also count as "inter-canister".
```
Dfinity-skaestle, line 75: This is a complete list of calls that are allowed.

update -> {update, query}
query -> {}
composite query -> {query, composite query}

Should we specify this in a bit more detail?

spec/index.adoc
mraszyk: We underspecify the exact order in the spec. I have originally used dfs call graph traversal in the pseudocode here, but got a review to underspecify that, i.e., allow multiple ways of traversing the call graph.

spec/index.adoc
mraszyk: This change is not committed, but necessary to process callbacks properly. Otherwise, the callbacks would run on the original canister state and not see changes made before `.await`.

spec/index.adoc
mraszyk: This change is not committed, but necessary to process callbacks properly. Otherwise, the callbacks would run on the original canister state and not see changes made before `.await`.

spec/index.adoc
mraszyk: We should hint at all places where the execution could fail. Could you provide a list here?

spec/index.adoc
mraszyk: The change is not committed, but necessary to process callbacks properly. Otherwise, the callbacks would run on the original canister state and not see changes made before .await.

spec/index.adoc
mraszyk: > currently

Is it going to change soon? It's not difficult to model, but it would clutter the pseudocode below.

nomeata: I guess that is implicit in “via a query call”, but could be made clearer indeed.

nomeata:
Not sure if this is on the agenda already, but since the caller needs to know whether a method is a query or a composite query method (makes a difference when doing a call from an update method), this requires extending candid to include that information in the interface, in <https://github.com/dfinity/candid/blob/master/spec/Candid.md#structure-1>, extending
```
<funcann>  ::= oneway | query
```
(Pinging @chenyan-dfinity, @rossberg and @crusso.)

spec/index.adoc
mraszyk: Added a note at two places.

spec/index.adoc
mraszyk: The pseudocode below gives a certificate to every call of a (composite) query method, not just the one called by the user.

spec/index.adoc
mraszyk: Now I've specified this a few lines above:
```
The public entry points of canisters are called _methods_. Methods can be declared to be either _update methods_ (state mutation is preserved, can call update and query methods of arbitrary canisters), _query methods_ (state mutation is discarded, no further calls can be made), or _composite query_ methods (state mutation is discarded, can call query and composite query methods of canisters on the same subnet).
```

mraszyk:
> Or do you plan to just ignore this issue and support update → compquery soon enough (so that query vs. compquery becomes an implementation detail not visible to the caller, as arguably it should)?

Very good point! @nomeata Replicated execution of composite query methods is not gonna be available any time soon to the best of my knowledge.

spec/index.adoc
Dfinity-skaestle: Getting certified variables work in the non-root canister isn't that easy, as some of the code has to be restructured. Some voices think that it's a simple rewrite, others think it might be more tricky.

Probably we need to try it and see how much work it is before we can give a timeline.

I'm fine with keeping the pseudocode simple, for now.

spec/index.adoc
mraszyk: Yes, thank you for providing the additional limits. I've implemented them in the pseudocode.

spec/index.adoc
ulan: +1 to making the ceritificate available everywhere in the call graph and keeping the spec simple

spec/index.adoc
ulan: Let's keep the spec simple. Certificates not being available in the call graph is more of a defect in the implementation.

spec/index.adoc
```diff
@@ -3996,9 +4021,93 @@ S with
...
+           do
+              if Depth = MAX_CALL_DEPTH_COMPOSITE_QUERY
+              then
+                 Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles) // max call graph depth exceeded
+              let Calls' · Call · Calls''  = Calls
+              Calls := Calls' · Calls''
+              if S.canister_subnet[Canister_id] ≠ S.canister_subnet[Call.callee]
+              then
+                 Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles) // calling to another subnet
+              let (Response', Cycles_used') = composite_query_helper(S, Cycles, Depth + 1, Canister_id, Call.callee, Call.method_name, Call.arg)
```
ulan, line 4084: I think the function returns the remaining cycles, not used cycles, right?

mraszyk: Very good catch! Thank you!

spec/index.adoc
```diff
@@ -3996,9 +4021,93 @@ S with
...
+              Cycles := Cycles - Cycles_used'
+              if Cycles < MAX_CYCLES_PER_RESPONSE
+              then
+                 Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles) // composite query out of cycles
+              let F' = Mod.callbacks(Call.callback, Response', 0, Env, 0)
```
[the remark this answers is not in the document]
mraszyk: Given that
```
ic0.call_cycles_add : (amount : i64) -> ();                                 // U Ry Rt T
ic0.call_cycles_add128 : (amount_high : i64, amount_low: i64) -> ();        // U Ry Rt T
```
which also matches the implementation I thought that no cycles can ever be attached in a composite query execution.

But they could be actually attached to calls made from callbacks (Ry, Rt). We can also use the following API in callbacks:
```
ic0.certified_data_set
ic0.global_timer_set
ic0.msg_cycles_accept
ic0.msg_cycles_accept128
```

I believe that we should introduce a pair of new modes, e.g., CRy and CRt, to denote callbacks in composite query contexts and make the above API unavailable in CRy and CRt. What do you think?

The alternative (your second option) could be confusing as we would allow sending cycles but we would never actually commit cycles sent in such a way. Moreover, we would then also need to allow setting certified variables and global timer although there would be no practical effect of that (except for retrieving the new certified variables in the same query - without signatures - and retrieving the previous value of global timer by setting a new value that is never gonna be used).

@ulan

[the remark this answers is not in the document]
mraszyk: This suggests that refunded cycles are set to zero for callbacks in composite queries: https://sourcegraph.com/github.com/dfinity/ic/-/blob/rs/execution_environment/src/query_handler/query_context.rs?L411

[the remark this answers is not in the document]
ulan: > I believe that we should introduce a pair of new modes, e.g., CRy and CRt, to denote callbacks in composite query contexts and make the above API unavailable in CRy and CRt. What do you think?

I agree. This seems to be the cleanest solution.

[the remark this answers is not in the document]
mraszyk: I've added the new modes (and also one for callback cleanup in composite queries).

spec/index.adoc
```diff
@@ -3996,9 +4025,93 @@ S with
...
+              balance = S.balances[Canister_id];
+              freezing_limit = freezing_limit(S, Canister_id);
+              certificate = Cert;
+              status = simple_status(S.canister_status[Canister_id]);
+              canister_version = S.canister_version[Canister_id];
+            }
+  if S.canisters[Canister_id] ≠ EmptyCanister and
+     S.canister_status[Canister_id] = Running and
+     S.balances[Canister_id] >= freezing_limit(S, Canister_id) and
+     Method_name ∈ dom(Mod.composite_query_methods) and
```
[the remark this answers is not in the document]
mraszyk: Very good catch! I've fixed the pseudocode now.

Dfinity-skaestle (approved):

spec/index.adoc
mraszyk: Given that we want to release composite queries soon and the certificate is not available everywhere in the call graph, I've made this commit (https://github.com/dfinity/interface-spec/pull/144/commits/f1ec4cd267c9a6cb103598f73bc5b9d832c54cde) to specify the current behavior. Could you please double check? @ulan

spec/index.md
```diff
@@ -72,11 +72,11 @@ To get some consistency in this document, we try to use the following terms with
...
 
 Methods can be *called*, from *caller* to *callee*, and will eventually incur a *response* which is either a *reply* or a *reject*. A method may have *parameters*, which are provided with concrete *arguments* in a method call.
 
-External calls can be update calls, which can call both kinds of methods, and query calls, which can *only* call query methods. Inter-canister calls can also call both kinds of methods. Note that calls from a canister to itself also count as "inter-canister".
+External calls can be update calls, which can *only* call update and query methods, and query calls, which can *only* call query and composite query methods. Inter-canister calls can call update and query methods (just like update calls). Note that calls from a canister to itself also count as "inter-canister".
```
oggy-dfin, line 79: I think this isn't correct anymore, since a composite query can issue an inter-canister calls.

spec/index.md
```diff
@@ -606,7 +606,7 @@ The HTTP response to this request can have the following responses:
 
 -   5xx HTTP status when the server has encountered an error or is otherwise incapable of performing the request. The request might succeed if retried at a later time.
 
-This request type can *also* be used to call a query method. A user may choose to go this way, instead of via the faster and cheaper [Request: Query call](#http-query) below, if they want to get a *certified* response. Note that the canister state will not be changed by sending a call request type for a query method.
+This request type can *also* be used to call a query method (but not a composite query method). A user may choose to go this way, instead of via the faster and cheaper [Request: Query call](#http-query) below, if they want to get a *certified* response. Note that the canister state will not be changed by sending a call request type for a query method.
```
crusso, line 609: Won't the balance change, even if the rest of the state does not?

spec/index.md
```diff
@@ -1659,15 +1675,15 @@ For each canister, the IC keeps track of "certified data", a canister-defined bl
 
     The canister can update the certified data with this call. The passed data must be no larger than 32 bytes. This can be used any number of times.
 
-When executing a query method via a query call (i.e. in non-replicated state), the canister can fetch a certificate that authenticates to third parties the value last set via `ic0.certified_data_set`.
+When executing a query or composite query method via a query call (i.e. in non-replicated mode), the canister can fetch a certificate that authenticates to third parties the value last set via `ic0.certified_data_set`. The certificate is not available in composite query method callbacks and in query and composite query methods evaluated on canisters other than the target canister of the query call.
```
crusso, line 1678: "other than the target canister of the query call"
I'm curious, why the restriction?

spec/index.md
```diff
@@ -4474,9 +4510,94 @@ S with
...
+         else if R = Return {new_state = W'; new_calls = Calls; response = Response; cycles_used = Cycles_used}
+         then
+            W := W'
+            if Cycles_used > MAX_CYCLES_PER_MESSAGE
+            then
+               Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles - MAX_CYCLES_PER_MESSAGE) // single message execution out of cycles
+            Cycles := Cycles - Cycles_used
+            if Response = NoResponse
+            then
+               while Calls ≠ []
```
oggy-dfin, line 4564: Don't we need something about not allowing calls for queries here?

spec/index.md
bogwar: I think canister state refers strictly to the canister memory and not system side state pertaining to the canister.

spec/index.md
bogwar: It's a deficiency of the current implementation afaik

Dfinity-Bjoern:
As discussed in interface spec meeting: we should resolve the open issues and comments and then it's ready to go in.

spec/index.md
```diff
@@ -4474,9 +4510,94 @@ S with
...
+                  freezing_limit = freezing_limit(S, Canister_id);
+                  certificate = Cert;
+                  status = simple_status(S.canister_status[Canister_id]);
+                  canister_version = S.canister_version[Canister_id];
+                }
+      if S.canisters[Canister_id] ≠ EmptyCanister and
+         S.canister_status[Canister_id] = Running and
+         S.balances[Canister_id] >= freezing_limit(S, Canister_id) and
+         (Method_name ∈ dom(Mod.query_methods) or Method_name ∈ dom(Mod.composite_query_methods)) and
+         Cycles >= MAX_CYCLES_PER_MESSAGE
```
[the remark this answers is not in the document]
crusso: Resolved offline.

spec/index.md
```diff
@@ -4474,9 +4510,94 @@ S with
...
 
-During the execution of a query call, a certificate is provided to the canister that is valid, contains a current state tree (or "recent enough"; the specification is currently vague about how old the certificate may be) and reveals the canister's [Certified Data](#system-api-certified-data).
+We define an auxiliary method that handles calls from composite query methods by performing a call graph traversal. It can also be (trivially) invoked for query methods that do not make further calls.
+
+    composite_query_helper(S, Cycles, Depth, Root_canister_id, Caller, Canister_id, Method_name, Arg) =
+      let Mod = S.canisters[Canister_id].module
+      let Cert <- { Cert | verify_cert(Cert) and
```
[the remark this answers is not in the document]
crusso: Resolved offline - its a non-det selection from a set of Certs/

spec/index.md
```diff
@@ -4474,9 +4510,94 @@ S with
...
+               do
+                  if Depth = MAX_CALL_DEPTH_COMPOSITE_QUERY
+                  then
+                     Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles) // max call graph depth exceeded
+                  let Calls' · Call · Calls''  = Calls
+                  Calls := Calls' · Calls''
+                  if S.canister_subnet[Canister_id] ≠ S.canister_subnet[Call.callee]
+                  then
+                     Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles) // calling to another subnet
+                  let (Response', Cycles') = composite_query_helper(S, Cycles, Depth + 1, Root_canister_id, Canister_id, Call.callee, Call.method_name, Call.arg)
```
crusso, line 4574: I guess that the use of recursion with original S, not updated canister state, ensures that re-entrant calls see the original canister state, and not the transient changes from the outer call, right?

mraszyk: Indeed.

spec/index.md
```diff
@@ -4474,9 +4510,94 @@ S with
...
+         S.canister_status[Canister_id] = Running and
+         S.balances[Canister_id] >= freezing_limit(S, Canister_id) and
+         (Method_name ∈ dom(Mod.query_methods) or Method_name ∈ dom(Mod.composite_query_methods)) and
+         Cycles >= MAX_CYCLES_PER_MESSAGE
+      then
+         let W = S.canisters[Canister_id].wasm_state
+         let F = if Method_name ∈ dom(Mod.query_methods) then Mod.query_methods[Method_name] else Mod.composite_query_methods[Method_name]
+         let R = F(Arg, Caller, Env)(W)
+         if R = Trap trap
+         then Return (Reject (CANISTER_ERROR, <implementation-specific>), Cycles - trap.cycles_used)
```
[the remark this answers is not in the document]
mraszyk: The implementation may wrap the trap messages with additional information, e.g., that what follows is a trap in the canister code.

spec/index.md
mraszyk: > not allowing calls for queries

a query method cannot make a call because it cannot use the System API to create and submit a call

spec/index.md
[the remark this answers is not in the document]
mraszyk: This restriction is an artifact of the current implementation.

spec/index.md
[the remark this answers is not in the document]
mraszyk: Concretely,
```
IC0503: Canister bnz7o-iuaaa-aaaaa-qaaaa-cai trapped explicitly: trap
```

spec/index.md
mraszyk: I've added an explicit note:
```
Note that the canister state (except for cycle balance change due to message execution) will not be changed ...
```

spec/index.md
```diff
@@ -72,11 +72,11 @@ To get some consistency in this document, we try to use the following terms with
 
 We avoid the term "client", as it could be the client of the Internet Computer or the client inside the distributed network that makes up the Internet Computer. Instead, we use the term *user* to denote the external entity interacting with the Internet Computer, even if in most cases it will be some code (sometimes called "agent") acting on behalf of a (human) user.
 
-The public entry points of canisters are called *methods*. Methods can be declared to be either *update methods* (state mutation is preserved) or *query methods* (state mutation is discarded, no further calls can be made).
+The public entry points of canisters are called *methods*. Methods can be declared to be either *update methods* (state mutation is preserved, can call update and query methods of arbitrary canisters), *query methods* (state mutation is discarded, no further calls can be made), or *composite query* methods (state mutation is discarded, can call query and composite query methods of canisters on the same subnet).
```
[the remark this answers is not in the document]
mraszyk: I suppose you're referring to this sentence
```
To the extent possible, this document will *not* talk about consensus protocols, nodes, subnets, orthogonal persistence or governance.
```
Given the formulation "to the extent possible", I don't see a reason to change the sentence as I think we're still trying to not talk about subnets more than necessary.

spec/index.md
mraszyk: Good catch! I've rephrased the sentence as follows:
```
Inter-canister calls issued while evaluating an update call can call update and query methods (just like update calls). Inter-canister calls issued while evaluating a query call (to a composite query method) can call query and composite query methods (just like query calls).
```

spec/index.md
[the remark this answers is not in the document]
mraszyk: > if I mark my method as composite_query, I get a guarantee that I won't be charged for it?

That's indeed the case in the current specification (and implementation).

> when we execute a query as an update, we charge cycles for that?

Indeed. See "Message execution" transition where we set (after simplifying the expressions for the special case of a call to a query method, using `query_as_update` helper):
```
balances[M.receiver] = (S.balances[M.receiver] + MAX_CYCLES_PER_MESSAGE) - res.cycles_used
```

> Is this a possibility that we would want to introduce for ordinary queries as well?

Not sure what exactly you mean.

spec/index.md
```diff
@@ -4474,9 +4510,94 @@ S with
...
+
+In query and composite query methods evaluated on the target canister of the query call, a certificate is provided to the canister that is valid, contains a current state tree (or "recent enough"; the specification is currently vague about how old the certificate may be), and reveals the canister's [Certified Data](#system-api-certified-data).
+
+:::note
+
+Composite query methods are EXPERIMENTAL and there might be breaking changes of their behavior in the future. Use at your own risk!
+
+:::
+
+Composite query methods can call query methods and composite query methods up to a maximum depth `MAX_CALL_DEPTH_COMPOSITE_QUERY` of the call graph. The total amount of cycles consumed by executing a (composite) query method and all (transitive) calls it makes must be at most `MAX_CYCLES_PER_COMPOSITE_QUERY`. This limit applies in addition to the limit `MAX_CYCLES_PER_MESSAGE` for executing a single (composite) query method and `MAX_CYCLES_PER_RESPONSE` for executing a single callback of a (composite) query method.
```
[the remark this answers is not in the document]
mraszyk: Good point! Added them to the "Unspecified constants and limits" section.

spec/index.md
```diff
@@ -4474,9 +4510,94 @@ S with
...
 
-During the execution of a query call, a certificate is provided to the canister that is valid, contains a current state tree (or "recent enough"; the specification is currently vague about how old the certificate may be) and reveals the canister's [Certified Data](#system-api-certified-data).
+We define an auxiliary method that handles calls from composite query methods by performing a call graph traversal. It can also be (trivially) invoked for query methods that do not make further calls.
+
+    composite_query_helper(S, Cycles, Depth, Root_canister_id, Caller, Canister_id, Method_name, Arg) =
+      let Mod = S.canisters[Canister_id].module
+      let Cert <- { Cert | verify_cert(Cert) and
+                           lookup(["canister", Canister_id, "certified_data"], Cert) = Found S.certified_data[Canister_id] and
+                           lookup(["time"], Cert) = Found S.system_time // or "recent enough"
```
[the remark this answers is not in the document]
mraszyk: That's a valid point, but should be addressed more generally I believe so I've filed an issue: https://github.com/dfinity/interface-spec/issues/195

spec/index.md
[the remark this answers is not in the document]
oggy-dfin: My problem is that I think we now talk about them less than necessary :) Like I said I think it's fine for this PR, I've added #197 to address this

spec/index.md
```diff
@@ -670,19 +678,25 @@ See [The system state tree](#state-tree) for details on the state tree.
 
 ### Request: Query call {#http-query}
 
-A query call is a fast, but less secure way to call a canister. Only methods that are explicitly marked as "query methods" by the canister can be called this way.
+A query call is a fast, but less secure way to call a canister. Only methods that are explicitly marked as "query methods" and "composite query methods" by the canister can be called this way. In contrast to a query method, a composite query method can make further calls to query and composite query methods of canisters on the same subnet.
```
oggy-dfin, line 681:
```suggestion
A query call is a fast, but less secure way to call a canister. Only methods that are explicitly marked as "query methods" and "composite query methods" by the canister can be called this way. In contrast to a query method, a composite query method can make further calls to query and composite query methods of canisters on the same subnet. The maximum depth of a composite query's call tree may, however, be limited by the implementation.
```

spec/index.md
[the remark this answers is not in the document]
oggy-dfin: I added also a suggestion to also mention this a bit earlier still (IIUC the depth is relatively small, so it's not impossible people will run into it in practice).

spec/index.md
mraszyk: There is one more restriction of this form: the total amount of cycles spent on all the message executions while evaluating a composite query call is bounded. I believe we should either mention that both or none. What would you prefer?

spec/index.md
```diff
@@ -3399,6 +3427,7 @@ S with
       }
     canister_status[CanisterId] = Running
     canister_version[CanisterId] = 0
+    canister_subnet[CanisterId] = Subnet
```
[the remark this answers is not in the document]
mraszyk: Does fd3250e address the concern?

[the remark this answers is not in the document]
oggy-dfin: Yes, thanks

spec/index.md
oggy-dfin: I think at this point we should (at least informally) indicate any restrictions that might be interesting for users. Actually, how about regular query methods? I know that there's also a limit on how long a replica will execute these for, but I don't know if the limit is expressed in terms of cycles, or if it's just a timer? If we want to move to replicated queries, I suppose we will need a limit based on cycles. 

Maybe for the time being we can just be vague and say that the implementation can impose an arbitrary limit on the execution time of both query and composite query methods?

mraszyk: The implementation indeed imposes this restriction:
```
/// The maximum time in seconds a query call is allowed to run.
pub(crate) const MAX_TIME_PER_COMPOSITE_QUERY_CALL: Duration = Duration::from_secs(10);
```

So I've added all the three restrictions to the spec: 1680a26fed8eb9c2d77e8a82d8330611bbbf1c36

oggy-dfin: Thanks!

Dfinity-Bjoern (approved):
Thank you @mraszyk

PR #146, opened 2023-03-09
CRP-1951: forbid cycles and self signed delegations in HTTP requests

gregorydemay:
Strengthens the requirements for an HTTP request to be valid if delegations are provided:
1. Delegations must not contain any cycles or be self-signed.
2. The number of `targets` that can be specified in each delegation is limited to 1000.

The length of the delegation chain is at most 20.


mraszyk:
@Dfinity-Bjoern The changes to the reference implementation have been merged in this [MR](https://github.com/dfinity/ic-hs/pull/154).

PR #147, opened 2023-03-10
Add separate HTTP Gateway Protocol spec and update with response verification v2

nathanosdev:
This PR moves the HTTP Gateway Protocol into its own spec. Since it is technically not part of the core IC protocol, it is better sitting on its own. This also includes the new response verification v2 ([CEL compatible response verification](https://docs.google.com/document/d/1eS9iQE33ehCgSEsB1_42rfjGZnSF6PkBQfm9_EkutQs/edit?usp=sharing)) and a lot of the spec has been reworked in an attempt to (hopefully) improve clarity.

I'm also under the impression that there is a plan to convert the IC spec to markdown, so I used markdown here, but if we prefer to stick with adoc for now I'm happy to convert it.


Dfinity-Bjoern:
Should we maybe leave the did file in the repo so people can import that into their projects if they want?

nathanosdev:
Thanks @Dfinity-Bjoern, I've added the file back. I'm not sure how linking to it from Markdown like this will work in the Portal, but I'll check it once it's updated there to make sure it's working.

I've also added some changes based on Joachim's feedback on the spec.

spec/http-gateway-protocol-spec.md
```diff
@@ -0,0 +1,500 @@
...
+
+4. Check whether the canister is hosted on the IC using a custom domain. There are two options:
+
+   - Check whether there is a TXT record containing a canister ID at the `_canister-id`-subdomain (e.g., to see whether `foo.com` is hosted on the IC, make a DNS lookup for the TXT record of `_canister-id.foo.com`) and use the specified canister ID;
+
+   - Make a `HEAD` request to the hostname. If the response contains an `x-ic-canister-id` header, use the value of this header as the canister ID.
+
+5. Else fail and handle the request as a Web2 request.
+
+If the hostname was of the form `<name>.ic0.app`, it is a _safe_ hostname; if it was of the form `<name>.raw.ic0.app`, it is a _raw_ hostname. Note that other domains may also be used to access canisters, such as `icp0.io`. The same logic concerning \_raw\_\_ domains is also be applied to these alternative domains.
```
keplervital, line 50:
```suggestion
If the hostname was of the form `<name>.ic0.app`, it is a _safe_ hostname; if it was of the form `<name>.raw.ic0.app`, it is a _raw_ hostname. Note that other domains may also be used to access canisters, such as `icp0.io`. The same logic concerning \_raw\_\_ domains can also be applied to these alternative domains.
```

keplervital (approved):

nomeata (changes requested):
Maybe I missed it, but the spec should ideally be very clear about whether the URL path segments are percent-encoded or not. For `/foo bar/index.html`, do we look up `["foo bar","index.html"]` or `["foo%20bar","index.html"]`. (I suspect and recommend the latter).

Maybe refer to https://www.rfc-editor.org/rfc/rfc3986#section-2 and explicitly state that the path segments can only contain characters
```
ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:?#[]@!$&'()*+,;=
```
(Nb: _no_ `/`)

Crucially, `<*>` is not valid. This ought to be checked, to avoid a badly written application allowing a user to, say, register the user name `<*>` and break havoc with the certification.

nomeata (approved):

Dfinity-Bjoern (approved):
Thank you Nathan!

PR #148, opened 2023-03-21
allow multiple request_status paths for a single request ID in read_state

mraszyk:
https://forum.dfinity.org/t/https-api-read-state-status-code-429-can-only-request-up-to-1-paths-for-request-status/19007


Dfinity-Bjoern (approved):
Makes sense, thanks @mraszyk!

PR #149, opened 2023-03-22
Update the interface spec to describe the limit on threshold ECDSA derivation paths

randombit:
BIP32 allows a derivation depth of at most 255, and we enforce this limit.

FOLLOW-1013

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1681,7 +1681,7 @@ This method takes no input and returns 32 pseudo-random bytes to the caller. The
 NOTE: The ECDSA API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
 
 This method returns a https://www.secg.org/sec1-v2.pdf[SEC1] encoded ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
-The `derivation_path` is a vector of variable length byte strings.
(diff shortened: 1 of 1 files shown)


robin-kunzler:
Thanks for addressing this @randombit ! Can't approve but LGTM :-)

randombit:
Ping @Dfinity-Bjoern

spec/index.adoc
```diff
@@ -1681,7 +1681,7 @@ This method takes no input and returns 32 pseudo-random bytes to the caller. The
 NOTE: The ECDSA API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
 
 This method returns a https://www.secg.org/sec1-v2.pdf[SEC1] encoded ECDSA public key for the given canister using the given derivation path. If the `canister_id` is unspecified, it will default to the canister id of the caller.
-The `derivation_path` is a vector of variable length byte strings.
+The `derivation_path` is a vector of variable length byte strings, containing at most 255 byte strings.
```
Dfinity-Bjoern, line 1684: Is 255 a bound on the length of the vector or a bound on the length of the strings in the vector?

randombit: It is the bound on the length of the vector; the length of each individual string is arbitrary. I amended the comment, PTAL.

mraszyk: Thank you!

PR #160, opened 2023-05-11
chore: minor clarification

dfinity-berestovskyy:
As we start every message execution with an empty execution state,
the `response_data` added by the `ic0.msg_reply_data_append` system call
and not committed using the `ic0.msg_reply` gets dropped at the end
of the current message execution.

In other words, it's impossible to append the reply data across await points.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1102,7 +1102,7 @@ Eventually, the canister will want to respond to the original call, either by re
 +
 Appends data it to the (initially empty) data reply.
 +
-NOTE: This can be invoked multiple times to build up the argument with data from various places on the Wasm heap. This way, the canister does not have to first copy all the pieces from various places into one location.
+NOTE: This can be invoked multiple times within the same message execution to build up the argument with data from various places on the Wasm heap. This way, the canister does not have to first copy all the pieces from various places into one location.
 +
 This traps if the current call already has been or does not need to be responded to.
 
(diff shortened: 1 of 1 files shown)


chmllr (approved):

dsarlis (approved):

mraszyk:
Thank you for the clarification! How about adding more of your comments from the PR description to the spec?

dsd-at-dfinity (approved):

spec/index.adoc
```diff
@@ -1102,9 +1102,11 @@ Eventually, the canister will want to respond to the original call, either by re
 +
 Appends data it to the (initially empty) data reply.
 +
-NOTE: This can be invoked multiple times to build up the argument with data from various places on the Wasm heap. This way, the canister does not have to first copy all the pieces from various places into one location.
+NOTE: This can be invoked multiple times within the same message execution to build up the argument with data from various places on the Wasm heap. This way, the canister does not have to first copy all the pieces from various places into one location.
 +
 This traps if the current call already has been or does not need to be responded to.
++
+Any data assembled, but not replied using `ic0.msg_reply`, gets discarded at the end of the current message execution.
```
Dfinity-Bjoern, line 1109:
```suggestion
Any data assembled, but not replied using `ic0.msg_reply`, gets discarded at the end of the current message execution. In particular, the reply buffer gets reset when the canister yields control without calling `ic0.msg_reply`.
```

Dfinity-Bjoern (approved):

spec/index.adoc
mraszyk: @dfinity-berestovskyy Does the additional sentence above make sense to you?

PR #161, opened 2023-05-15
Public Spec: Optional backwards compatible name callbacks

levifeldman:
This is a follow on PR to https://github.com/dfinity/interface-spec/pull/51 to use named exported functions for callbacks.
This PR however puts the name-callbacks functionality into the system-api functions `ic0.call_new_name_callbacks` & `ic0.call_on_cleanup_name` as alternatives for `ic0.call_new` & `ic0.call_on_cleanup`.
Each call can use either `ic0.call_new` or `ic0.call_new_name_callbacks` but not both. Same for `ic0.call_on_cleanup` & `ic0.call_on_cleanup_name`.

This PR is backwards compatible.

### DIFF
--- a/spec/ic0.txt
+++ b/spec/ic0.txt
@@ -29,7 +29,6 @@ ic0.msg_method_name_size : () -> i32;                                       // F
 ic0.msg_method_name_copy : (dst : i32, offset : i32, size : i32) -> ();     // F
 ic0.accept_message : () -> ();                                              // F
 
-
 ic0.call_new :                                                              // U Ry Rt T
   ( callee_src  : i32,
     callee_size : i32,
--- a/spec/ic0.txt
+++ b/spec/ic0.txt
@@ -29,14 +29,16 @@ ic0.msg_method_name_copy : (dst : i32, offset : i32, size : i32) -> ();     // F
 ic0.accept_message : () -> ();                                              // F
 
 ic0.call_new :                                                              // U Ry Rt H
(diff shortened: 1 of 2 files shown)


nomeata:
Cool, thanks!

Technically, I believe the old one is backwards compatible, as Wasm itself can handle importing and exporting functions with same names, as long as the types are different. But pragmatically, hoping that all tooling involved can do that too is probably not a good idea, so 👍

levifeldman:
@nomeata The old one didn't leave room in the spec for canisters to continue registering callbacks with a table index and still be in sync with the spec.

nomeata:
I think my idea was that the replica can support multiple API versions, i.e. one doesn't need “old” function signatures in the spec for them to still work. But keeping both around as officially supported is good, too!

levifeldman:
Thanks for laying the groundwork for this feature @nomeata.

levifeldman:
@Dfinity-Bjoern @manudrijvers @mraszyk @dsarlis What do you guys think?

levifeldman:
@Dfinity-Bjoern @manudrijvers @mraszyk @dsarlis Feedback?

Dfinity-Bjoern:
Closing the PR for now because there is no active work and it is unclear that converting it to markdown is beneficial at this point. The PR can be converted and reopened at any point in time, such as when the engineering teams actually have capacity to work on the implementation.

PR #162, opened 2023-05-24
Add missing response hash pseudoheader

nathanosdev:
The `:ic-cert-status` pseudoheader is missing from the response hash calculation for the response verification spec.


Dfinity-Bjoern (approved):

nathanosdev:
Sorry @Dfinity-Bjoern I didn't notice that you already approved and I just pushed again a suggestion from @robin-kunzler. When you get a chance, could you confirm you're okay with the additional commit?

spec/http-gateway-protocol-spec.md
```diff
@@ -372,7 +375,7 @@ The steps for response verification are as follows:
...
+Canisters can report the versions of response verification that they support using public metadata in the [system state tree](https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-canister-information). This metadata will be read by the HTTP Gateway using a [read_state request](https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-read-state). This metadata is a comma-delimited string of versions under the key "supported_certificate_versions”, for example: "1,2". This is treated as an optional, additional layer of security for canisters supporting multiple versions. If the metadata has not been added (i.e. the lookup of this metadata in the `read_state` response returns `Absent`), then the HTTP Gateway will allow for whatever version the canister has responded with.
```
mraszyk, line 378: > the lookup of this metadata in the `read_state` response returns `Absent`

Note that this is not possible given the current replica implementation that produces responses for which the [lookup](https://ic-interface-spec.netlify.app/#_lookup) of a non-existent path yields `Unknown`. @derlerd-dfinity

mraszyk, line 378: @nathanosdev Note that the read_state request can also fail with an HTTP error if the canister defines the corresponding custom section as "private".

PR #163, opened 2023-05-30
specify replica-signed queries

mraszyk:
This PR adds public keys of nodes in a subnet to the certified state tree and specifies query call response signatures produced by nodes evaluating the query call and returned as part of the query response.


spec/index.adoc
```diff
@@ -629,17 +632,52 @@ In order to make a query call to canister, the user makes a POST request to `/ap
...
+  status: "rejected",
+  reject_code: <reject_code>,
+  reject_message: <reject_message>,
+  error_code: <error_code>,
+  timestamp: <t>,
+  request_id: hash_of_map({request_type: "query", sender: <sender>, nonce: <nonce>, ingress_expiry: <ingress_expiry>, canister_id: <canister_id>, method_name: <method_name>, arg: <arg>})
+})
+....
++
+where `hash_of_map` is the <<hash-of-map, representation-independent hash>> and `<t>` is the timestamp of the node when the signature was produced.
```
nomeata, line 680: How does the client know the timestamp used by the replica? Should this timestamp be added to the response?

Also, is the request_id here the same that's used by the client to sign the request?
If yes, maybe simply state this, instead of (redundantly) repeating it's calculation algorithm?
If no (different fields maybe), better don't use the same name here.

nomeata (commented):
BCOR changes are not in the PR yet. Is this already ready for review, or still a draft?

spec/index.adoc
mraszyk: > How does the client know the timestamp used by the replica? Should this timestamp be added to the response?

The replica's timestamp is part of a node signature record together with the node principal and the actual signature (please see above).

> Also, is the request_id here the same that's used by the client to sign the request?

Yes.

> If yes, maybe simply state this, instead of (redundantly) repeating it's calculation algorithm?

I've replaced `hash_of_map({request_type: "query", sender: <sender>, nonce: <nonce>, ingress_expiry: <ingress_expiry>, canister_id: <canister_id>, method_name: <method_name>, arg: <arg>})` by `hash_of_map(content)`.

mraszyk:
> BCOR changes are not in the PR yet. Is this already ready for review, or still a draft?

Thank you for taking a look! I've added the CBOR changes to the request CDDL specification.

spec/requests.cddl
```diff
@@ -66,13 +66,21 @@ query-content = {
...
   reject_message: text
   ? error_code: text
+  node_signatures: [* node-signature-of-query-response]
 }>
 
+node-signature-of-query-response = {
+  timestamp: timestamp
+  signature: bytes
+  identity: principal
+}
```
nomeata, line 82: This requires clients to separately fetch the public key from the IC, including figuring out the right subnet etc. This is different from update call signatures and from certified data, where the full certificate (including subnet delegation etc.) is included, and can be checked with no further data. Is there a reason to do that differently?

spec/requests.cddl
```diff
@@ -66,13 +66,21 @@ query-content = {
...
   //
   status: "rejected"
   reject_code: unsigned
   reject_message: text
   ? error_code: text
+  node_signatures: [* node-signature-of-query-response]
 }>
 
+node-signature-of-query-response = {
+  timestamp: timestamp
```
nomeata, line 79: Does this timestamp have a particular meaning across different replicas? Is it related to the timestamp in the state tree that this request is based on? Or is there really not anything we can say here?

spec/requests.cddl
```diff
@@ -66,13 +66,21 @@ query-content = {
 query-response = tagged<{
   status: "replied"
   reply: call-reply
+  node_signatures: [* node-signature-of-query-response]
```
nomeata, line 69:
```suggestion
  ? node_signatures: [* node-signature-of-query-response]
```
Maybe? Or will now all query calls be signed? Then
```suggestion
  node_signatures: [+ node-signature-of-query-response]
```
if that's valid CBOR.

spec/index.adoc
```diff
@@ -3910,17 +3948,17 @@ Read response::
 * If `F(Q.Arg, Q.sender, Env) = Trap trap` then
 +
 ....
-{status: rejected; reject_code: CANISTER_ERROR, reject_message: <implementation-specific>, error_code: <implementation-specific>}
+{status: rejected; reject_code: CANISTER_ERROR, reject_message: <implementation-specific>, error_code: <implementation-specific>, node_signatures: <implementation-specific>}
```
nomeata, line 3951: Not really implementation specific, is it? It's specified above what it should be. Maybe just write `[]` to omit from the example.

spec/index.adoc
```diff
@@ -415,6 +415,9 @@ tagged<t> = #6.55799(t) ; the CBOR tag
 +
 NOTE: Because this uses the lexicographic ordering of princpials, and the byte distinguishing the various classes of ids is at the _end_, this range by construction conceptually includes principals of various classes. This specification needs to take care that the fact that principals that are not canisters may appear in these ranges does not cause confusion.
 
+* `/subnet/<subnet_id>/nodes/<node_id>/public_key` (blob)
++
+The public key of a node (a DER-encoded Ed25519 signing key, see https://tools.ietf.org/html/rfc8410[RFC 8410] for reference) with principal `<node_id>` belonging to the subnet with principal `<subnet_id>`.
```
nomeata, line 420: Is that data on every subnet? Or only the root subnet? Or on each subnet, but only that subnet?

spec/requests.cddl
nomeata: I'd include a field `certificate` (like for certified variables) that reveals the corresponding public keys, and that it is responsible for the given canister.

spec/index.adoc
nomeata: Ah, with the CDDL I see it now. Thanks!

spec/requests.cddl
mraszyk: It does not have a particular meaning (yet). It is simply the timestamp when the signature was produced to allow clients check for replayed responses from far in the past. If we support making queries at a specified height (in the future), then we could relate it to the timestamp in the certified state tree.

spec/index.adoc
mraszyk: > Not really implementation specific, is it?

The exact value is implementation-specific because it depends on how the BN chooses the replica(s) for executing the query.

> Maybe just write [] to omit from the example.

This would constraint an implementation to never include any signature into the response.

spec/requests.cddl
mraszyk: Do we really need an optional vector here if the field is omitted? Doesn't CBOR default to an empty vector in that case?

spec/requests.cddl
mraszyk: > This is different from update call signatures

It is still similar in the sense that in both cases you make a read_state request that gives you the delegation and the corresponding path in the subnet's certified state tree.

> Is there a reason to do that differently?

The main reason is to save network bandwidth by not including the certificate in every query response.

> I'd include a field certificate (like for certified variables) that reveals the corresponding public keys, and that it is responsible for the given canister.

This is indeed an alternative option to be discussed in the design review.

spec/requests.cddl
nomeata: I think CBOR distinguishes between field absent, and field present with empty vector as the list. 

The question is what we want: are signatures optional? Or will they be mandatory once this lands?

spec/index.adoc
nomeata: Ok, fine with me

spec/index.adoc
nomeata: But we still need to specify (precisely, ideally) what a client ought to check when validating a signature. How do they get the public key, and know whether its the right one? Should they trust any subnet (as long as it has a valid delegation), only the root subnet, or only the subnet of the current canister, or the latter two.

spec/requests.cddl
nomeata: Sounds like premature optimization at the expense of significant client complexity (caching, extra reques, staleness) to me.

My suggestion is:

* Allow clients to opt into signature in the request, to save work and bandwidth when the client doesn't care.
* Include the delegation and state tree in the response 
* include it in a way that with multiple replicas answering, the boundary node can merge the state trees (the data structure is set up for precisely such use cases!) to save bandwidth. I.e. a single value next to the list of signatures, not one in each.

rumenov: @nomeata in the design discussion few reasons were discussed in favour if not adding the certificate to the response.

1. It is not a premature optimization. [Node providers](https://wiki.internetcomputer.org/wiki/Node_Provider_Network_Setup_Guide) are required to have 300mbps  (37.5 MB/s) of bandwidth. All of this should account for all components that send and receive bytes over the wire - consensus, statesync, ingress, xnet, bitcoin, https outcalls, queries, read state, etc. When you start adding the required bandwidth for all those components once the IC scales just a bit more we start running into problems pretty soon.

In production queries right now are the biggest bandwidth consumer. Adding the certificate will more than double the query response. 

2. Adding the certificate in the response enables clients to do the same thing in two different ways which is not ideal for an API as well.

3. User complexity is pretty bounded given that only agents will have to implement this extra complexity, it won't impact the average user.

nomeata: Fair enough

spec/requests.cddl
derlerd-dfinity: I think if we don't include the certificate in the response (which sounds like a reasonable thing to do as per the discussion [here](https://github.com/dfinity/interface-spec/pull/163/files#r1228516513)) we might be able to afford making the signatures mandatory. This sounds like a nice thing from a security perspective, while it would also enable us to keep the API simple without the need to introduce a second endpoint for signed queries. WDYT?

mraszyk: > we might be able to afford making the signatures mandatory [...] without the need to introduce a second endpoint for signed queries

Even with a single endpoint, we can still keep signatures optional and include them based on the request type (this was @rumenov 's suggestion above).

Hence, I see three options (no certificate in the response in all cases):
1. same endpoint, mandatory signatures
2. same endpoint, signatures based on a new field in the request
3. two endpoints, mandatory signatures for the new endpoint

Could we go for option 1? @rumenov

derlerd-dfinity: > Even with a single endpoint, we can still keep signatures optional and include them based on the request type

Agreed. What I meant to say is that if we want an API thats as simple as possible, I'd rather go for a mandatory signature than for a second endpoint. If we don't mind changing the request type, we can of course have a single endpoint with an optional signature.

rumenov: @mraszyk my preference is option 2 actually. I think David is also advocating for this. Not sure what other think. @nomeata ?

rumenov: I apologize, I wrongly mistyped option 2 initially, sorry about that. 

I am all for **making them mandatory** especially if we think that long term most queries should require signatures.

mraszyk: Based on the above, I've specified the list of node signatures to be non-empty.

spec/index.adoc
mraszyk: Added a precise definition of query response signature validation to the spec.

spec/index.adoc
```diff
@@ -4199,18 +4199,52 @@ Read response::
...
 ....
 * Else if `F(Q.Arg, Q.sender, Env) = Return {response = Reply R; …}` then
 +
 ....
-{status: replied; reply: { arg :  <R> }, node_signatures: <implementation-specific>}
-....
+{status: "replied"; reply: {arg: <R>}, node_signatures: <implementation-specific>}
+....
+
+Given a certificate obtained by requesting the path `/subnet` via a read state request
```
derlerd-dfinity, line 4215: Could we somehow only read the keys we are interested in here as opposed to the entire `/subnet` path?

spec/index.adoc
```diff
@@ -4199,18 +4199,52 @@ Read response::
...
+to `/api/v2/canister/<ECID>/read_state`, the following predicate describes
+when the returned response is correctly signed by the nodes:
+....
+verify_response({status: "replied"; reply: {arg: <R>}, node_signatures: Sigs}, Cert)
+  = verify_cert(Cert) ∧
+    ((Cert.delegation = NoDelegation ∧ SubnetId = RootSubnetId ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert) = Found Ranges) ∨
+     (SubnetId = Cert.delegation.subnet_id ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert.delegation.certificate) = Found Ranges)) ∧
```
derlerd-dfinity, line 4222: My tendency would be to also mandate including a self delegation for the NNS to avoid the special case here.

spec/index.adoc
```diff
@@ -4199,18 +4199,52 @@ Read response::
...
+    ECID ∈ Ranges ∧
+    ∀ {timestamp: T, signature: Sig, identity: NodeId} ∈ Sigs.
+      lookup(["subnet",SubnetId,"nodes",NodeId,"public_key"], Cert) = Found PK ∧
+      verify_sig PK Sig ("\x0Bic-response" · hash_of_map({
+        status: "rejected",
+        reject_code: RejectCode,
+        reject_message: RejectMessage,
+        error_code: ErrorCode,
+        timestamp: T,
+        request_id: hash_of_map(E.content)}))
```
derlerd-dfinity, line 4244: Given that only this part is different in request and response validation, I wonder whether it would make more sense to extract this part into subroutines so that the rest of the validation logic becomes easier to read and understand.

derlerd-dfinity (commented):
Thanks a lot! Left some comments.

spec/requests.cddl
ShuoWangNSL: I think there are three aspects to think about whether to make node signatures mandatory (taking both replica-signed queries and certified queries into account).

1. Some queries are already secured by certified variables, e.g. http asset, rosetta.
2. Some queries will count on the certified query feature .
3. Our long-term perspective on the question if all queries should by default require node signatures.

For 1, it will be a waste of work(signing, network traffic, merging certificate, merging sigs) for these queries.
For 2, NNS dapp has asked for this feature and it will be the main user in the early stage. I don't foresee a surge of demand for certified queries. 
For 3, I am not sure that all requests should by default demand signatures if we considering charging queries in the future.

For the reasons above, I go for option 2. WDYT? @derlerd-dfinity  @mraszyk @rumenov @nomeata

spec/index.adoc
mraszyk: That's tricky because we don't know the subnet ID of the canister based on the query response and thus cannot ask for the specific subtree of `/subnet` for the subnet hosting the queried canister.

spec/requests.cddl
ShuoWangNSL: I got some stats of the certificate size for node keys.
In a network of 40 subnets and in the subnet with 40 nodes:
The certificate for 1 node key is around 1100 bytes. 
The certificate for 14 node keys varies from 1800 bytes to 2400 bytes. 
The certificate for 40 node keys is around 3300 bytes.
In comparison, a delegation is around 500 bytes.

@rumenov 
> In production queries right now are the biggest bandwidth consumer. 
I think there are two aspects regarding the bandwidth consume for certified queries.
1.  Will certified queries account for majority of queries in the future or used in cases where security is a big concern(e.g NNS dapp)?  
This is the same point I bring up regarding whether to make signatures optional.

2. Does fetching node keys through read_state requests really save the bandwidth?
Agents do not know which nodes will respond and thus need to fetch and cache all keys.
Agents do not know the subnet id when making query calls. It is not trivial to tell if agents already have node keys in cache and if a read_state call should be made.

spec/index.adoc
mraszyk: Would you then also mandate an NNS self delegation for certificates passed to canisters in query calls (carrying the canister's certified variables)?

ShuoWangNSL:
I have a strong feeling that the attitudes towards the pros and cons of different options depend on how we see the role of certified queries will play in the long term. It is a topic for the follow-up feature of certified queries but now affects our design.

1. If the certified query is one more way to secure the response in addition to certified variables and is only used when security is a real concern. Then I think including the certificate inside the response is a good way similar to the way we append a 500-KiB delegation to the read_state response .

2. If the certified queries are going to take up the majority of the query traffic in the future, it makes sense to save the 1 to several KiB of certificate size on every query response.

My feeling is that we start with 1 and if we gradually transit to 2 in the future, we don't waster any effort and can still develop in a compatible way.

spec/index.adoc
mraszyk: I've refactored the definition of `verify_response` to remove the redundancy.

mraszyk:
> I have a strong feeling that the attitudes towards the pros and cons of different options depend on how we see the role of certified queries will play in the long term. It is a topic for the follow-up feature of certified queries but now affects our design.
> 
> 1. If the certified query is one more way to secure the response in addition to certified variables and is only used when security is a real concern. Then I think including the certificate inside the response is a good way similar to the way we append a 500-KiB delegation to the read_state response .
> 2. If the certified queries are going to take up the majority of the query traffic in the future, it makes sense to save the 1 to several KiB of certificate size on every query response.
> 
> My feeling is that we start with 1 and if we gradually transit to 2 in the future, we don't waster any effort and can still develop in a compatible way.

I couldn't quite follow the above suggestion. In particular, it doesn't seem backwards-compatible to start omitting the certificate in the query's response when transitioning from 1. to 2. in the future.

spec/index.adoc
nomeata: Valid point, but given that this part is like the existing certification procedure, I'd not change it here, so that the same verification code can be used

derlerd-dfinity: > Valid point, but given that this part is like the existing certification procedure, I'd not change it here, so that the same verification code can be used

Wouldn't the existing verification procedure just transparently work with a self-delegation of the NNS?

> Would you then also mandate an NNS self delegation for certificates passed to canisters in query calls (carrying the canister's certified variables)?

Assuming my point above is correct, we could only require the NNS delegation to be present in this verification algorithm here, but not elsewhere. In summary, I think that if we can make this somehow elegantly work without making the scope of this change too broad it would be nice to have. Otherwise, I'm also happy with this special case here.

spec/requests.cddl
derlerd-dfinity: > 2. Does fetching node keys through read_state requests really save the bandwidth?
> [...]
> It is not trivial to tell if agents already have node keys in cache and if a read_state call should be made.

This is an excellent point. I guess it would be possible to simply invalidate the cache if the certificate is, e.g., older than a couple of minutes (it contains a timestamp). Not claiming this is trivial, but even if it would be, still
- many clients may not implement caching at all and so would do `read_state` requests on every query,
- many clients may only occasionally make queries meaning that that even if they implement caching they will have to re-fetch because the cache is already invalid at the point where they make the next query.

My takeaway from this is that there are use cases where the approach to fetch certificates separately via `read_state` is better and there are use cases where the approach to include certificates in the response is better. Or in other words, bandwidth consumption is probably not the deciding factor regarding whether or not to include the certificate in the reply. 

Taking this into account, I think there is no longer a good reason why we shouldn't allow the client to choose whether or not it wants to include the certificate in the response. This means going for option 2 from the [other thread](https://github.com/dfinity/interface-spec/pull/163#discussion_r1231842381) also including an option to request or not request a certificate seems the most reasonable to me.

spec/requests.cddl
derlerd-dfinity: Your comments on the other thread invalidated my claim that it is likely the most reasonable thing to do to not include the certificate into the response. So I'm also leaning towards Option 2 (also see [here](https://github.com/dfinity/interface-spec/pull/163#discussion_r1232046914)).

spec/index.adoc
nomeata: When you say NNS subnet, you mean root subnet, right? (It happens that the NNS canisters live on the root subnet, but that's more a coincidence than a necessary and could change in the future, and here the root-of-trust aspect is the relevant one.)

I'm still not convinced. The gain is very small; the code already has to be able to check certificates with delegation and certificates without (at least to check the delegations). Forcing the delegation chain to have length 1 doesn't save much here, but incurs relatively pointless additional work and bandwidth (in the form of the self-delegation).

derlerd-dfinity: > When you say NNS subnet, you mean root subnet, right?

Yes. Sorry. 

> I'm still not convinced. [...]

As said above, I don't have a very strong opinion on this. Lets keep what we have now.

ShuoWangNSL:
> it doesn't seem backwards-compatible to start omitting the certificate in the query's response when transitioning from 1. to 2. in the future.

@mraszyk When transitioning to 2, we develop the prefetching and caching node keys in clients. Then the client can specify it they require signature and/or certificate(no sig no cert/ sig with cert, sig without cert). In replica side, the certificate is already an optional field, it can just be set as None.

spec/index.adoc
ShuoWangNSL: That's tricky. I also find that clients do not know the subnet id if they want to prefetch the node keys because they only know about the canister id.

derlerd-dfinity: > I also find that clients do not know the subnet id if they want to prefetch the node keys because they only know about the canister id.

Note that based on what Martin is suggesting they wouldn't need to know the subnet ID. When making a read state query with the canister ID they are interested in then the request will end up on the right subnet. So when reading the entire `/subnet` subtree one would get the right keys even without knowing the subnet ID. One could then look up the subnet ID in the tree that comes back.

derlerd-dfinity:
> > it doesn't seem backwards-compatible to start omitting the certificate in the query's response when transitioning from 1. to 2. in the future.
> 
> @mraszyk When transitioning to 2, we develop the prefetching and caching node keys in clients. Then the client can specify it they require signature and/or certificate(no sig no cert/ sig with cert, sig without cert). In replica side, the certificate is already an optional field, it can just be set as None.

I'm not sure I quite follow either. If the certificate is an optional field in step 1 already, then we'd also need an additional request parameter to let the client specify what it wants, i.e., already implicitly support 2 as well as clients can choose whether they want a certificate or not. This would mean you are suggesting the same thing as @nomeata suggests in his reply to your initial message (and the same as Option 2 on the other thread)?

ShuoWangNSL:
> This would mean you are suggesting the same thing as @nomeata suggests in his reply to your initial message (and the same as Option 2 on the other thread)

That's right. I am suggesting what @nomeata said. I don't think all request should demand node signatures given the fact that a considerable amount of query is already secured by certified variables and some amount of query won't care about certification that much.

In conclusion, 
1. If a considerable amount of query does not require node sigs, we should make sigs optional and let clients specify.
It is the current status and I don't think that's going to change, i.e. switching from certified variables to node sigs.

2. If certified queries are only used in scenarios where security is real concern instead of general cases, it is acceptable to include certificate in the response.
It is the current status, e.g. NNS dapps wants to use certified queries to load balance and proposals, etc. However, we are not sure of the long-term future though. 
At this stage, prefetching all node keys does not necessarily save the bandwidth when certified query is used occasionally.  
If we eventually decide to transition, clients can specify another field or extend the the certification_level field in the request to inform replicas to omit the certificate.

@rumenov WDYT about (1) the long-term role of certified queries in IC (2) making node signatures optional and (3) including certificate in response for now?

spec/requests.cddl
mraszyk: There are two separate decisions to be taken:

A. should we include node signatures in the response?
(1) same endpoint, mandatory signatures
(2) same endpoint, signatures based on a new field in the request
(3) two endpoints, mandatory signatures for the new endpoint

B. should we include (node key) certificate in the response?
(1) never (=> must make a read_state request)
(2) based on an optional integer field included in the request
(3) always (=> no need for a separate read_state request)

I believe we had already converged to A1 and stay with that decision.

Regarding B, we decided B1 and now we're changing our decision to B2, correct?

mraszyk:
> @mraszyk will spec changes for the request type be added as part of this PR as well ?

Yes, I'll include those changes into this PR, too.

mraszyk:
> i.e. having a request parameter where we can either say
> I want nothing
> I want a signature
> I want a signature and a certificate

As I pointed out [here](https://github.com/dfinity/interface-spec/pull/163#discussion_r1232817956), node signatures (present at all? if so, how many?) and (node key) certificate (in query response or via read_state) are two independent decisions to be taken so I wouldn't mix them in a single request field.

Now the three options (nothing, only signature, signature and certificate) suggest you prefer to go for A2 and B2.

spec/requests.cddl
nomeata: > I believe we had already converged to A1 and stay with that decision.

B2 with A1 is an odd choice. If you already introduce configurabilityim the request, then at least I don't see why not allowing the option “no signatures wanted”, and thus A2 (which would be my favorite combination). Maybe this has all been discussed to death internally already, though :-)

derlerd-dfinity: > I believe we had already converged to A1 and stay with that decision.
> [...]
> > B2 with A1 is an odd choice.

I agree. that it is an odd choice. I think we should either make both signatures and certificates requestable or nothing.

In my understanding we had converged on making signatures mandatory under the [assumption](https://github.com/dfinity/interface-spec/pull/163#discussion_r1229169260) that it doesn't make sense to include the certificate, but Shuo's argument [here](https://github.com/dfinity/interface-spec/pull/163#discussion_r1232046914) invalidated this IMHO. 

So I think we should either:
1. Make both, the certificate and the signatures requestable with (an) additional parameter(s) in the request
2. Stick with decision A1 but then also don't include certificates.

With this new input, my tendency is towards 1 and it seems others landed on the same conclusion. If I understood correctly, @rumenov's objections were related to making the API more complicated by allowing the client to select what it wants. @rumenov would you also be fine with 1 given the new inputs linked above?

rumenov:
> i.e. having a request parameter where we can either say

This is my concern, IHMO such a request type looks pretty bad.

Are the signatures useful without the certificate ? In which situation will a user request just the signature without the certificate ?

rumenov:
> I have a strong feeling that the attitudes towards the pros and cons of different options depend on how we see the role of certified queries will play in the long term. It is a topic for the follow-up feature of certified queries but now affects our design.
> 
> 1. If the certified query is one more way to secure the response in addition to certified variables and is only used when security is a real concern. Then I think including the certificate inside the response is a good way similar to the way we append a 500-KiB delegation to the read_state response .
> 2. If the certified queries are going to take up the majority of the query traffic in the future, it makes sense to save the 1 to several KiB of certificate size on every query response.
> 
> My feeling is that we start with 1 and if we gradually transit to 2 in the future, we don't waster any effort and can still develop in a compatible way.

The problem is if we start with 1. In order to transition to 2 then you will have to break many clients that you will have to fix. So this transitioning will never happen.

nomeata:
> again we should just provide the recommended way of doing things and not give user the flexibility to shoot themselves in the foo

Of course! But we can define the recommend way in the agent code rather than the replica code. User ≠ client. If our guiding principle for the public spec was simplicity for a user using the spec directly, a lot of things would look different - consider the complexities of certified variables!

So I don't buy the “user” argument here.

Is there a reason why an agent implementer must be prevented from using the current interface (no signatures)? Because if not, and if it is useful in some cases (certified variables for example), then why prevent it?

(It also seems simpler to add an optional field than introducing a complete new incompatible endpoint in this case, but that's not the main point.)

rumenov:
> (It also seems simpler to add an optional field than introducing a complete new incompatible endpoint in this case, but that's not the main point.)

This is kind of the point. I am just in general against the optional fields both in the request and in the response. The less optional fields we have I believe this will be better for everyone.

Also an additional endpoint will make the whole IC implementation more straight forward with less risk of bugs.

rumenov:
>  But we can define the recommend way in the agent code rather than the replica code. 

The agent API is not part of the spec. Let's be honest, what ever you in some random github repo outside of the spec API people will still not read it. 

> User ≠ client. If our guiding principle for the public spec was simplicity for a user using the spec directly (not through an agent or a CDK), a lot of things would look different - consider the complexities of certified variables!

I guess I meant clients here, i.e. people that implement something on top of the spec.

rumenov:
@nomeata let's take this approach of having the many optional fields as part of the request, 

a similar analogy (at least to me, and maybe taking it to the extreme) is the following. Why classes/structs needs to allow multiple constructors? Why not have one constructor with a bunch of enums and optional fields  ?

another analogy, if you look at most microservices implemented with gRPC then you will not fined a single method in the service with one huge oneoff inside the request protobuf.

derlerd-dfinity:
> > (It also seems simpler to add an optional field than introducing a complete new incompatible endpoint in this case, but that's not the main point.)
> 
> This is kind of the point. I am just in general against the optional fields in the request. The less optional fields we have I believe this will be better for everyone.

Note that we have [established](https://github.com/dfinity/interface-spec/pull/163#discussion_r1232046914) that in some cases it is better to include certificates in the response whereas in some cases it is better to to read them via `read_state` (see also [here](https://github.com/dfinity/interface-spec/pull/163#discussion_r1231848625) for a different argument leading to a similar conclusion). We have also [established](https://github.com/dfinity/interface-spec/pull/163#discussion_r1231842381) that (in contrast to what we initially thought) sometimes we need signatures in the reply and sometimes we don't. This means that if we don't want to allow optional fields we'd already have 3 endpoints, in the future maybe even more.

I think it is hard to claim that 3 different endpoints for essentially the same thing modulo signatures/certificates is simpler in any way, no?

rumenov:
> > > (It also seems simpler to add an optional field than introducing a complete new incompatible endpoint in this case, but that's not the main point.)
> > 
> > 
> > This is kind of the point. I am just in general against the optional fields in the request. The less optional fields we have I believe this will be better for everyone.
> 
> Note that we have [established](https://github.com/dfinity/interface-spec/pull/163#discussion_r1232046914) that in some cases it is better to include certificates in the response whereas in some cases it is better to to read them via `read_state` (see also [here](https://github.com/dfinity/interface-spec/pull/163#discussion_r1231848625) for a different argument leading to a similar conclusion). We have also [established](https://github.com/dfinity/interface-spec/pull/163#discussion_r1231842381) that (in contrast to what we initially thought) sometimes we need signatures in the reply and sometimes we don't. This means that if we don't want to allow optional fields we'd already have 3 endpoints (assuming a certificate without a signature doesn't make sense; otherwise even 4), in the future maybe even more.
> 
> I think it is hard to claim that 3 different endpoints for essentially the same thing modulo signatures/certificates is simpler in any way, no?

You are correct many endpoints is also an overkill. How about a single experimental `certified_query` endpoint for the time being? In this case I will have absolutely no preference on the types. And once the certified query with multiple signatures is scoped we will have better understanding what exactly are the use cases and what will be our long term plan.

derlerd-dfinity:
> You are correct many endpoints is also an overkill. How about a single experimental certified_query endpoint for the time being? In this case I will have absolutely no preference on the types. And once the certified query with multiple signatures is scoped and once you get more mileage with the already implemented functionality you will have better understanding what exactly are the use cases and what will be our long term plan.

Sounds like a pragmatic way forward. So lets do this.

I'm still not entirely convinced because I think that we will quite likely end up with very similar (if not the same) types as envisioned for the single endpoint proposal. However, this may be a feature and not a bug because it would allow us to collapse the two endpoints back into one `query` endpoint (keeping `certified_query` as an alias for `query`) if this turns out to be the most meaningful thing.

mraszyk:
I've updated the replica-signed query spec. Could you please take a look? @derlerd-dfinity @rumenov

spec/index.adoc
```diff
@@ -626,64 +627,97 @@ See <<state-tree>> for details on the state tree.
...
+=== Request: Signed query call
+
+NOTE: The signed query call API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
+A *signed* query call is a query call amended with signatures produced by IC nodes evaluating the query call to certify the query response.
+
+In order to make a signed query call to a canister, the user makes a POST request to `/api/v2/canister/<effective_canister_id>/signed_query`.
+The request body has the same structure as the request body for <<http-query>> with the `content` map having the following *additional* fields:
+
+* (optional) `return_signatures` (`text`): Determines whether and how many signatures are returned in the response; possible values are `"no"`, `"one"`.
```
derlerd-dfinity, line 665: Shall we go for `none`, instead of `no`? Or is there something more suitable, e.g., something similar to an enum, to encode the choices here?

spec/index.adoc
```diff
@@ -626,64 +627,97 @@ See <<state-tree>> for details on the state tree.
...
+* (optional) `return_certificate` (`text`): Determines whether a certificate (as specified in <<certification>>) is returned in the response; possible values are `"no"`, `"yes"`.
+
+The response is a CBOR (see <<cbor>>) with the same fields as for <<http-query>> response and the following *additional* fields:
+
+* (optional) `signatures` (`[+ node-signature]`): a non-empty list of node signatures for the returned query response.
+* (optional) `certificate` (`blob`): a certificate, as specified in <<certification>>.
+
+Unless `return_signatures` is set to `"no"` in the request, the response to a signed query call contains a non-empty list of signatures
```
derlerd-dfinity, line 673: The list would also be empty if `return_signatures` is omitted, right?

spec/index.adoc
```diff
@@ -626,64 +627,97 @@ See <<state-tree>> for details on the state tree.
...
+for the returned response produced by the individual IC nodes that computed the same returned response.
+Every such signature (whose type is denoted as `node-signature`) is a CBOR (see <<cbor>>) map with the following fields:
+
+* `timestamp` (`nat`): the timestamp of the signature.
+* `signature` (`blob`): the actual signature.
+* `identity` (`principal`): the principal of the node producing the signature.
+
+Unless `return_certificate` is set to `"no"` in the request, the response to a signed query call contains a certificate containing
```
derlerd-dfinity, line 681: Same as above.

derlerd-dfinity (commented):
Thanks a lot, @mraszyk. LGTM.

spec/index.adoc
mraszyk: > Shall we go for none, instead of no?

Sure, why not.

> Or is there something more suitable, e.g., something similar to an enum, to encode the choices here?

There's no native support for enum in CBOR. So we need to use integers or text. I went for text as integers could suggest we control how many signatures we precisely get.

spec/index.adoc
mraszyk: fixed

spec/index.adoc
mraszyk: fixed

spec/index.adoc
```diff
@@ -619,29 +627,97 @@ See <<state-tree>> for details on the state tree.
...
-Canister methods that do not change the canister state can be executed more efficiently. This method provides that ability, and returns the canister’s response directly within the HTTP response.
+* `timestamp` (`nat`): the timestamp of the signature.
+* `signature` (`blob`): the actual signature.
+* `identity` (`principal`): the principal of the node producing the signature.
+
+Unless `return_certificate` is null or set to `"no"` in the request, the response to a signed query call contains a certificate containing
+all paths in the state tree with prefixes
+
+* `/time`. Always contained in the certificate.
+* `/subnet/<subnet_id>/canister_ranges`. Contained in the certificate if the certificate does not have a subnet delegation, as specified in <<certification-delegation>>.
```
ShuoWangNSL, line 685: @mraszyk Does this field means we don't include NNS self delegation in certificate but verify the canister ranges directly from the tree?

spec/index.adoc
```diff
@@ -619,29 +627,97 @@ See <<state-tree>> for details on the state tree.
...
+=== Request: Signed query call
+
+NOTE: The signed query call API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
+A *signed* query call is a query call amended with signatures produced by IC nodes evaluating the query call to certify the query response.
+
+In order to make a signed query call to a canister, the user makes a POST request to `/api/v2/canister/<effective_canister_id>/signed_query`.
+The request body has the same structure as the request body for <<http-query>> with the `content` map having the following *additional* fields:
+
+* (optional) `return_signatures` (`text`): Determines whether and how many signatures are returned in the response; possible values are `"none"`, `"one"`.
```
rumenov, line 665: disclaimer: just a question, I will not force my preferences here :) 

Shouldn't we start with what is needed for the first use case - namely the NNS dapp and then extend the request type as new uses emerge instead of starting with request type that has 4 different combinations.

@derlerd-dfinity

derlerd-dfinity: It was my initial suggestion to use an enum, yes. But the discussion on this MR revealed that two separate fields are probably better: for replica-signed queries we'd need the variants:
- 1 Signature without certificate
- 1 Signature with certificate
- (Nothing)

For certified queries (which will be what the NNS dapp use case requires) we need to extend this to
- 1 Signature without certificate
- 1 Signature with certificate
- (Nothing)
- Majority of signatures without certificate
- Majority of signatures with certificate

Based on this I think it is better to let the client choose separately for each field instead of forcing a dependency between these two fields and ending up with a huge enum.

We could discuss whether we want to allow the option for `Nothing` (effectively allowing us to drop `return_signatures` _for now_) given that this can also be obtained via `query`. However, this doesn't seem worth it because
- we already know now that we will have to re-introduce `return_signatures` to encode that we want more than one signatures in the immediate follow up feature on certified queries. 
- it gives us the property that `query` and `certified_query` behave the same if no signatures are requested (or the parameters are omitted). So those two endpoints could still be collapsed into one at a later point. 

Neither of these are very strong arguments, but I also think there are no extremely strong arguments for the alternative. So I'd suggest to simply go ahead with this for now in the interest of progress. If we find that this was a mistake we can always revert given that the API is declared experimental.

nomeata: > So those two endpoints could still be collapsed into one at a later point.

(Or just don't introduce a new endpoint if it is merely a fully compatible extension of an existing one…)

spec/index.adoc
mraszyk: Indeed. See here: https://github.com/dfinity/interface-spec/pull/163#discussion_r1232164262

spec/requests.cddl
```diff
@@ -73,8 +75,39 @@ query-response = tagged<{
   ? error_code: text
 }>
 
-call-reply = {
+; A request as submitted to /api/v2/.../signed_query
+signed-query-request = envelope<signed-query-content>
+signed-query-content = {
+  request_type: "query"
```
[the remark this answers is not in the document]
mraszyk: I think it should be "query" if want to merge the two endpoints in the future with as few changes to the agents as possible.

spec/index.adoc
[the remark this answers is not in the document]
mraszyk: > why the fields are text instead of nat ?

nat could suggest you can specify the exact number which won't be the case in the future when we want to provide multiple signatures (where the exact number will depend on the subnet size and the user won't be able to provide the exact number of signatures to retrieve)

> maybe requested_signatures, requested_certifications is better naming?

sounds good to me; please put "thump up" if you like "requested_signatures", "requested_certifications"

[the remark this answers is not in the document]
mraszyk: renamed the fields

oggy-dfin: Do we envision a use case where we want to request a certificate for a query without requesting a signature? If that's not the case, then two independent fields don't make sense to me, and I'd rather use a single, optional field `signature` or similar, which is a record with a boolean (or optional) `with_certificate` inside.

spec/index.md
```diff
@@ -670,37 +680,107 @@ See [The system state tree](#state-tree) for details on the state tree.
...
+-   contained in the response, i.e., `Cert = R.certificate`, or
+
+-   obtained by requesting the path `/subnet` in a **separate** read state request to `/api/v2/canister/<ECID>/read_state`,
+
+the following predicate describes when the returned response `R` is correctly signed by the nodes:
+
+    verify_response(R, Cert)
+      = verify_cert(Cert) ∧
+        ((Cert.delegation = NoDelegation ∧ SubnetId = RootSubnetId ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert) = Found Ranges) ∨
+         (SubnetId = Cert.delegation.subnet_id ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert.delegation.certificate) = Found Ranges)) ∧
```
oggy-dfin, line 764: In the theory of the spec, we support nested delegations (and our agents actually go through the trouble of supporting them), while here we seem to assume that there's only a single delegation. This could make the check insecure if we ever switched to nested delegations. 

I'd be happy to kill nested delegations though unless someone has a hot use case for them.

spec/index.md
```diff
@@ -670,37 +680,107 @@ See [The system state tree](#state-tree) for details on the state tree.
...
+
+-   (optional) `requested_certificate` (`text`): Determines whether a certificate (as specified in [Certification](#certification)) is returned in the response; possible values are `"no"`, `"yes"`.
+
+The response is a CBOR (see [CBOR](#cbor)) with the same fields as for [Request: Query call](#http-query) response and the following **additional** fields:
+
+-   (optional) `signatures` (`[+ node-signature]`): a non-empty list of node signatures for the returned query response.
+
+-   (optional) `certificate` (`blob`): a certificate, as specified in [Certification](#certification).
+
+Unless `requested_signatures` is null or set to `"none"` in the request, the response to a signed query call contains a non-empty list of signatures for the returned response produced by the individual IC nodes that computed the same returned response. Every such signature (whose type is denoted as `node-signature`) is a CBOR (see [CBOR](#cbor)) map with the following fields:
```
oggy-dfin, line 735: We can still have the boundary nodes return HTTP error responses, no? I see we haven't covered this for queries or read_state yet. Should we do so?

spec/index.adoc
derlerd-dfinity: I can't think of such a use case at the moment. However, I also don't see a good reason to prevent it. So given that with a single field the number of enum variants would grow quite a bit in the future (at latest when support for aggregating multiple signatures is added; see the [message](https://github.com/dfinity/interface-spec/pull/163#discussion_r1238056069) above) I'd tend to stick with the current proposal.

oggy-dfin: I think we're having different things in mind, I meant that we could have:
```
signed-query-content = {
  ...
  ? signature : signature-options
}

signature-options = {
  with_certificate : bool
  ? multiplicity: "single" / "multiple"
}
```
(names left to taste). This doesn't require huge enums.

derlerd-dfinity: I see. No strong preference here. I'd leave it up to @mraszyk. If something like this is easy to model and he sees no other issues then it might be an option.

spec/index.md
mraszyk: Even with nested delegations, the actual subnet ID will be the subnet ID of the last delegation and the canister ranges of every delegation should be included in the canister ranges of its parent delegation (this condition is currently not in the spec, but how about it we just add this condition).

spec/index.md
```diff
@@ -670,37 +680,107 @@ See [The system state tree](#state-tree) for details on the state tree.
...
+
+-   contained in the response, i.e., `Cert = R.certificate`, or
+
+-   obtained by requesting the path `/subnet` in a **separate** read state request to `/api/v2/canister/<ECID>/read_state`,
+
+the following predicate describes when the returned response `R` is correctly signed by the nodes:
+
+    verify_response(R, Cert)
+      = verify_cert(Cert) ∧
+        ((Cert.delegation = NoDelegation ∧ SubnetId = RootSubnetId ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert) = Found Ranges) ∨
```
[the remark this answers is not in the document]
mraszyk: Fixed by this MR: https://github.com/dfinity/interface-spec/pull/193

spec/index.md
mraszyk: Please see the commit d993538.

spec/index.adoc
mraszyk: > Do we envision a use case where we want to request a certificate for a query without requesting a signature?

The certificate alone could be useful if the caller wants to determine the subnet on which the canister is hosted (e.g., because the caller only wants to talk to a subnet with some feature - fiduciary, EU, ...).

oggy-dfin: Isn't this use case of determining the subnet covered by `read_state` requests?

spec/index.md
oggy-dfin: I don't think my comment about this becoming insecure made sense - sorry. But this still seems off if we have nested delegations. If the subnet changes along the chain of delegations, which subnet should we pick? This seems to pick the first one in the chain. Is that the right choice? Should it be the last one in the chain instead? Should it be any subnet in the chain? But then we probably don't want to allow mixing and matching different subnets, otherwise a selection of malicious nodes from each subnet could reach a quorum?

So I'd still propose to just kill nested delegations if we don't have a use for them, and if noone remembers what the envisioned future was where they would make sense. Right now they just seem to be making our lives more complicated.

spec/index.md
oggy-dfin: That makes sense, thanks - but I'm not sure if that's the complete story? For example, what if we try to query a canister that's not in any range in the routing table? I guess we get a 404 back directly from the boundary node? If that's correct, then I'd add something just to be consistent with the update call spec. I don't think we need to resolve this in this PR as you're not introducing the problem, an issue would do.

spec/index.adoc
mraszyk: > Isn't this use case of determining the subnet covered by read_state requests?

You can make a read_state request, but you can also make it for a signed_query. The reason for returning the certificate in a signed_query is that you don't need to make an additional read_state request.

spec/index.md
mraszyk: > If the subnet changes along the chain of delegations, which subnet should we pick?

We should pick the outermost one (as we're doing it now) because the delegation starts with NNS being the innermost.

mraszyk: Regarding why we even have nested delegations: wouldn't it be more complicated to enforce we have only have a single delegation? In the verification algorithm running on a certificate, you'd need a Boolean value expressing whether it's the outer or inner certificate; now you can just define the verification recursively without focusing on the recursion depth.

spec/index.md
mraszyk: > If that's correct, then I'd add something just to be consistent with the update call spec. I don't think we need to resolve this in this PR as you're not introducing the problem, an issue would do.

Agreed. Here's the issue: https://github.com/dfinity/interface-spec/issues/194

spec/index.adoc
oggy-dfin: I'm not convinced by that use case; you'd issue a query to check that the canister is hosted on, say, an EU subnet, but you would be OK not having any evidence that the query was executed correctly. And this would somehow be preferable to issuing a `read_state` request.

So I still think the types I proposed align better with actual use cases, but it seems I'm the only one who thinks that. Given that it's a minor issue, I'll drop it.

mraszyk: > you would be OK not having any evidence that the query was executed correctly.

You wouldn't have this evidence in case of a single signature either.

> And this would somehow be preferable to issuing a read_state request.

Issuing a separate read_state request would increase the latency (and slightly also the bandwidth because you would need to again send all the headers etc.).

spec/index.md
```diff
@@ -670,37 +680,122 @@ See [The system state tree](#state-tree) for details on the state tree.
...
+        (Q.requested_signatures = "one" ∧ R.reject_code ≠ SYS_TRANSIENT => |R.signatures| = 1) ∧
+        (Q.requested_certificate = null ∨ Q.requested_certificate = "no" ∨ R.reject_code = SYS_TRANSIENT => Cert = null) ∧
+        verify_cert(Cert) ∧
+        ((Cert.delegation = NoDelegation ∧ SubnetId = RootSubnetId ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert) = Found Ranges) ∨
+         (SubnetId = Cert.delegation.subnet_id ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert.delegation.certificate) = Found Ranges)) ∧
+        effective_canister_id ∈ Ranges ∧
+        ∀ {timestamp: T, signature: Sig, identity: NodeId} ∈ R.signatures.
+          lookup(["subnet",SubnetId,"nodes",NodeId,"public_key"], Cert) = Found PK ∧
+          if R.status = "replied" then
+            verify_sig PK Sig ("\x0Bic-response" · hash_of_map({
```
robin-kunzler, line 784: @mraszyk a minor thing I think this should be
```suggestion
            verify_signature PK signature ("\x0Bic-response" · hash_of_map({
```
correct? 

Same for line 790

spec/index.md
```diff
@@ -670,37 +680,122 @@ See [The system state tree](#state-tree) for details on the state tree.
...
+          else
+            verify_sig PK Sig ("\x0Bic-response" · hash_of_map({
+              status: "rejected",
+              reject_code: R.reject_code,
+              reject_message: R.reject_message,
+              error_code: R.error_code,
+              timestamp: T,
+              request_id: hash_of_map(Q)}))
+
+where `RootSubnetId` is the a priori known principal of the root subnet. Moreover, all timestamps in `R.signatures`, the certificate `Cert`, and its optional delegation must be "recent enough".
```
robin-kunzler, line 798: @mraszyk Would it make sense to give more precise guidance on how recent the sig timestamps and certificates should be? Mainly asking because it might not be clear to implementers what sensible values might be. We could also defer this to the internal design doc though if you feel that this would be too much detail on this level.

robin-kunzler (commented):
Nice work on the spec @mraszyk , this looks very good. Left two minor comments, see below.

spec/index.md
mraszyk: Thanks for the catch! Fixed.

spec/_attachments/requests.cddl
```diff
@@ -73,8 +75,39 @@ query-response = tagged<{
   ? error_code: text
 }>
 
-call-reply = {
+; A request as submitted to /api/v2/.../signed_query
+signed-query-request = envelope<signed-query-content>
+signed-query-content = {
+  request_type: "query"
```
rumenov, line 81: I believe the request type needs to be signed_query as well

spec/index.md
```diff
@@ -670,37 +680,122 @@ See [The system state tree](#state-tree) for details on the state tree.
...
+
+:::
+
+A **signed** query call is a query call amended with signatures produced by IC nodes evaluating the query call to certify the query response.
+
+In order to make a signed query call to a canister, the user makes a POST request to `/api/v2/canister/<effective_canister_id>/signed_query`. The request body has the same structure as the request body for [Request: Query call](#http-query) with the `content` map having the following **additional** fields:
+
+-   (optional) `requested_signatures` (`text`): Determines whether and how many signatures are returned in the response; possible values are `"none"`, `"one"`.
+
+-   (optional) `requested_certificate` (`text`): Determines whether a certificate (as specified in [Certification](#certification)) is returned in the response; possible values are `"no"`, `"yes"`.
```
rumenov, line 727: "possible values are `"no"`, `"yes"`" why don't you make the field bool then ?

spec/index.md
```diff
@@ -2589,6 +2684,8 @@ These are the synchronous read messages:
         canister_id : CanisterId;
         method_name : Text;
         arg : Blob;
+        requested_certificate : Text;
```
rumenov, line 2687: Should here be another type ?

CanisterSignedQuery ?

spec/index.md
mraszyk: I was thinking of adding more variants in the future, e.g., requesting specific paths to be present in the certificate.

spec/_attachments/requests.cddl
mraszyk: Then it'd be more difficult to merge `/api/v2/.../signed_query` with `/api/v2/.../query` as a URL redirect would not be sufficient (clients would also need to adapt their CBOR payloads).

spec/index.md
mraszyk: I've renamed the `CanisterQuery` type to `CanisterSignedQuery` to not introduce a new type in the formal part of the spec.

spec/index.adoc
mraszyk: This conversation has been resolved offline.

spec/index.md
mraszyk: This conversation has been resolved offline.

spec/_attachments/requests.cddl
mraszyk: This conversation has been resolved offline.

spec/index.md
```diff
@@ -709,23 +717,61 @@ In order to make a query call to canister, the user makes a POST request to `/ap
...
+-   `status` (`text`): `"rejected"`
 
 -   `reject_code` (`nat`): The reject code (see [Reject codes](#reject-codes)).
 
 -   `reject_message` (`text`): a textual diagnostic message.
 
 -   `error_code` (`text`): an optional implementation-specific textual error code (see [Error codes](#error-codes)).
 
-Canister methods that do not change the canister state (except for cycle balance change due to message execution) can be executed more efficiently. This method provides that ability, and returns the canister's response directly within the HTTP response.
+-   (optional) `signatures` (`[+ node-signature]`): a singleton list of node signatures for the returned query response.
```
derlerd-dfinity, line 740: IMHO the formulation "singleton list" is somewhat ambiguous as it is unclear whether it refers to the list or its contents. IIUC we want it to refer to its contents, so maybe we can make this more clear.

spec/index.md
```diff
@@ -2595,7 +2641,7 @@ These are the synchronous read messages:
         sender : UserId;
         paths : List(Path);
       }
-      | CanisterQuery = {
+      | CanisterSignedQuery = {
```
derlerd-dfinity, line 2644: I think this should be `CanisterQuery` now that signatures are just returned as part of the normal query, right? (similarly below)

spec/index.md
```diff
@@ -4608,21 +4654,30 @@ Conditions
...
 
--   Else if `composite_query_helper(S, MAX_CYCLES_PER_QUERY, 0, Q.canister_id, Q.sender, Q.canister_id, Q.method_name, Q.arg) = (Reply R, _)` then
+        {status: "replied"; reply: {arg: Res}, signatures: Sigs}
 
-        {status: "replied"; reply: {arg: R}}
+where the query `Q`, the response `R`, and a certificate `Cert'` that is obtained by requesting the path `/subnet` in a **separate** read state request to `/api/v2/canister/<effective_canister_id>/read_state` satisfy the following:
+
+```html
+
+verify_response(Q, R, Cert') ∧ lookup(["time"], Cert') = Found S.system_time // or "recent enough"
```
derlerd-dfinity, line 4678: `... = S.system_time` seems to be a condition that is impossible to meet in practice, no?

derlerd-dfinity (commented):
Thank you. LGTM.

spec/index.md
mraszyk: That's why we have the remark `// or "recent enough` and the [issue](https://github.com/dfinity/interface-spec/issues/195).

spec/index.md
mraszyk: Done.

spec/index.md
mraszyk: I've rephrased the formulation and also added a note why we have a list at the first place.

spec/index.md
```diff
@@ -4620,14 +4679,23 @@ S.system_time <= Q.ingress_expiry
...
 
-        {status: "rejected"; reject_code: RejectCode; reject_message: RejectMsg; error_code: <implementation-specific>}
+        {status: "rejected"; reject_code: RejectCode; reject_message: RejectMsg; error_code: <implementation-specific>, signatures: Sigs}
+
+-   Else if `composite_query_helper(S, MAX_CYCLES_PER_QUERY, 0, Q.canister_id, Q.sender, Q.canister_id, Q.method_name, Q.arg) = (Reply Res, _)` then
+
+        {status: "replied"; reply: {arg: Res}, signatures: Sigs}
 
--   Else if `composite_query_helper(S, MAX_CYCLES_PER_QUERY, 0, Q.canister_id, Q.sender, Q.canister_id, Q.method_name, Q.arg) = (Reply R, _)` then
+where the query `Q`, the response `R`, and a certificate `Cert'` that is obtained by requesting the path `/subnet` in a **separate** read state request to `/api/v2/canister/<effective_canister_id>/read_state` satisfy the following:
```
ShuoWangNSL, line 4692: > requesting the path `/subnet`

In this way, users don't need to specify the subnet_id, which they may not know in advance.

However, in implementation, the `public_key` and `canister_ranges` for all subnets are populated in the state tree in each subnet. As a result, in addition to the node keys, uses get many extra paths, resulting a larger certificate size.

Suppose we have 36 subnets, and canister ranges only have one entry of (principal, principal). Each principal is 29 bytes and the subnet public key is 48 bytes. In total, that's 36 * (29 + 29 + 48) = 3816 bytes.

I had some early experiments showing that with 40 node keys, the tree size is 2800 bytes. I need to do some more experiments on the tree size with paths of all subnets. I estimate it will be around 3 times larger but not that large to make a significant difference on network transmission.

Still, it would be great if there is a simple way to get rid paths of other subnets in this read_state call.

derlerd-dfinity: Thanks a lot for pointing this out. This is very useful data. We discussed multiple ways of working around this in the future. Options could, e.g., be

- Providing an API that allows clients to obtain the subnet ID that hosts a canister based on its ID. 
- Allowing to request the certificate as part of the response.

For this iteration I'd consider solving this out of scope, though.

ShuoWangNSL: The [verify_responses()](https://github.com/dfinity/interface-spec/pull/163/files#diff-b9a833d6114611a080aee4cd57618feb45aa044bcc11105d44b155b5bbde00ebR759) method above checks:
 `Cert.delegation = NoDelegation ∧ SubnetId = RootSubnetId` or `SubnetId = Cert.delegation.subnet_id`. However, users do not know the subnet id in the first place. 

Actually, if users do not know the subnet id in advance, the condition above serves as the way to figure out the subnet id after having the certificate. @mraszyk

ShuoWangNSL: From the workflow above, there are some awkward part related to subnet id:
a. Before the read_state call, users do not know the subnet_id and thus request the whole `/subnet` path
b. After getting the certificate, users get the subnet_id from the delegation. If no delegation, it implies the subnet is root subnet but there is no hardcoded root subnet id. Hardcoded root subnet id also brings problems in testnet settings.
c.  With the whole tree at path `/subnet`, users need to know the subnet id to look up the node keys. Otherwise, it has to loop on all forks until it find the fork with node keys.

In summary, there are some options before request and after response:
Before request:
1. Providing an API that allows clients to obtain the subnet ID that hosts a canister based on its ID.
2. Allow users requesting node keys with a default subnet id
3. Considering not put node keys under `/subnet/<subnet-id>/node` but directly under **/metadata/node**. cc: @mraszyk We used to argue for `/subnet/<subnet-id>/node` because it conveys the subnet membership. However, it seems the membership information is anyway not straightforward if not knowing the subnet id.
 
After response:
1. Add /metadata/own_subnet_id in the state tree and include it in the certificate. So users has an easy way to know about subnet_id if delegation is not present.

derlerd-dfinity: Thanks @ShuoWangNSL for your explanations. I think we should not overcomplicate this. The way to get the subnet ID you describe above is IMO good enough for the initial iteration (and what we had agreed upon IIRC). We can always improve moving forward.

mraszyk: How about returning the subnet ID as an additional field in the query response?

ShuoWangNSL: Regarding option 3, putting node keys under `/subnet/<subnet-id>/node` gives us the benefit of being able to look at the state tree as one global tree. Therefore, let's keep this one.

The options 1 , 2, and 4 can be considered later. We can use the current way of obtaining subnet_id in our initial iteration.

ShuoWangNSL: > How about returning the subnet ID as an additional field in the query response?

As the read_state is a separate call, I think it is not always feasible to use the information from another call. In practice, I think agents may want parallelize these two calls to reduce latency.

spec/index.md
```diff
@@ -641,7 +645,11 @@ The HTTP response to this request consists of a CBOR (see [CBOR](#cbor)) map wit
 
     If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation's canister id range (see [Delegation](#certification-delegation)).
 
-The returned certificate reveals all values whose path is a suffix of a requested path. It also always reveals `/time`, even if not explicitly requested.
+The returned certificate reveals all values whose path has a requested path as a prefix except for
+
+-   paths with prefix `/subnet/<subnet_id>/nodes`. Only contained in the returned certificate if `<effective_canister_id>` belongs to the canister ranges of the subnet `<subnet_id>`, i.e., if `<effective_canister_id>` belongs to the value at the path `/subnet/<subnet_id>/canister_ranges` in the state tree.
```
ShuoWangNSL, line 650: There are restrictions on paths with prefix `/subnet/<subnet_id>/nodes` but  `/subnet` can always be requested.

This seems to make the restrictions void as `/subnet` contains `/subnet/<subnet_id>/nodes`.

Shall we apply the restriction while reading `/subnet`?

ShuoWangNSL: Never mind. I misunderstand this sentence. It does mean the results of `/subnet` follow this rule.

spec/index.md
mraszyk: > Regarding option 3, putting node keys under /subnet/<subnet-id>/node gives us the benefit of being able to look at the state tree as one global tree. Therefore, let's keep this path.

Agreed!

> In practice, I think agents may want to parallelize these two calls to reduce latency.

Good point! Should we then close this thead?

spec/index.md
```diff
@@ -445,6 +445,10 @@ Because this uses the lexicographic ordering of princpials, and the byte disting
 
 :::
 
+-   `/subnet/<subnet_id>/nodes/<node_id>/public_key` (blob)
```
ShuoWangNSL, line 448: I recall that we use the singular form when drafting design docs and it is consistent with other paths like /subnet and /canister in singular form.  I prefer the singular form here.

The [MR](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/13266) of certifying node keys has been merged. If we want to use the plural form, we will have to make a quick fix before it is rolled out. 
cc: @mraszyk @derlerd-dfinity

mraszyk: Makes totally sense! Thank you for spotting this and proposing the change. I've updated this spec PR to use the singular form.

spec/index.md
mraszyk: In the Interface Specification meeting, we decided to follow up on removing nested delegations separately. Hence, I'm closing this discussion.

Dfinity-Bjoern (approved):
Seems fine now, please make sure the names align between code and spec. (I don't have a strong opinion on which of the names we choose, only that we are consistent.)

spec/index.md
```diff
@@ -721,23 +729,74 @@ In order to make a query call to canister, the user makes a POST request to `/ap
...
+Canister methods that do not change the canister state (except for cycle balance change due to message execution) can be executed more efficiently. This method provides that ability, and returns the canister's response directly within the HTTP response.
+
+If the query call resulted in a reply, the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
 
--   `status` (`text`): `replied`
+-   `status` (`text`): `"replied"`
 
 -   `reply`: a CBOR map with the field `arg` (`blob`) which contains the reply data.
 
+-   (optional) `signatures` (`[+ node-signature]`): a list containing one node signature for the returned query response.
```
mraszyk, line 740: Should we make the signatures non-optional? The replica code MR currently declares the signatures as a non-optional field (same for error code).

derlerd-dfinity: As far as I can tell this might be a good idea. Even if they are non-optional and populated old agents will simply ignore the additional field on deserialization. So there is no need for a rollout as an optional taking into account backward compatibility issues.

ShuoWangNSL: Last time discussing about it, we think the optional field may be useful for boundary nodes if they fail to resemble enough matched query results. In this case, we'd want reject it at BN without node signatures.

However, it seems a plaintext http error is enough instead of a HttpQueryResponse::Rejected with node signatures missing.

spec/index.md
```diff
@@ -721,23 +729,74 @@ In order to make a query call to canister, the user makes a POST request to `/ap
...
+
+-   `signature` (`blob`): the actual signature.
+
+-   `identity` (`principal`): the principal of the node producing the signature.
+
+Given a query (the `content` map from the request body) `Q`, a response `R`, and a certificate `Cert` that is obtained by requesting the path `/subnet` in a **separate** read state request to `/api/v2/canister/<effective_canister_id>/read_state`, the following predicate describes when the returned response `R` is correctly signed:
+
+    verify_response(Q, R, Cert)
+      = verify_cert(Cert) ∧
+        ((Cert.delegation = NoDelegation ∧ SubnetId = RootSubnetId ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert) = Found Ranges) ∨
```
ShuoWangNSL, line 773: The canister ranges of root subnet is present because the whole `/subnet` path is returned.

If only requesting `/subnet/<root_subnet_id>/node`, is `/subnet/<root_subnet_id>/canister_ranges` automatically appended like the `/time`?

I am thinking about if this [MR](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/14035) is already enough for exposing node keys or `/subnet/<root_subnet_id>/canister_ranges` path should be treated specially

spec/index.md
mraszyk: Thank you for your input! I've made the signatures field non-optional.

spec/index.md
mraszyk: > If only requesting /subnet/<root_subnet_id>/node, is /subnet/<root_subnet_id>/canister_ranges automatically appended like the /time?

No, it's not appended automatically.

> I am thinking about if this [MR](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/14035) is already enough for exposing node keys or /subnet/<root_subnet_id>/canister_ranges path should be treated specially

The MR you linked is enough because clients can request both paths `/subnet/<root_subnet_id>/node` and `/subnet/<root_subnet_id>/canister_ranges` in one read_state requests.

spec/index.md
```diff
@@ -721,23 +729,74 @@ In order to make a query call to canister, the user makes a POST request to `/ap
 
 -   `arg` (`blob`): Argument to pass to the canister method.
 
-If the call resulted in a reply, the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
+Canister methods that do not change the canister state (except for cycle balance change due to message execution) can be executed more efficiently. This method provides that ability, and returns the canister's response directly within the HTTP response.
```
THLO, line 732:
```suggestion
Canister methods that do not change the canister state (except for cycle balance changes due to message execution) can be executed more efficiently. This method provides that ability, and returns the canister's response directly within the HTTP response.
```

spec/index.md
```diff
@@ -642,7 +646,11 @@ The HTTP response to this request consists of a CBOR (see [CBOR](#cbor)) map wit
 
     If this `certificate` includes subnet delegations (possibly nested), then the `effective_canister_id` must be included in each delegation's canister id range (see [Delegation](#certification-delegation)).
 
-The returned certificate reveals all values whose path is a suffix of a requested path. It also always reveals `/time`, even if not explicitly requested.
+The returned certificate reveals all values whose path has a requested path as a prefix except for
+
+-   paths with prefix `/subnet/<subnet_id>/node`. Only contained in the returned certificate if `<effective_canister_id>` belongs to the canister ranges of the subnet `<subnet_id>`, i.e., if `<effective_canister_id>` belongs to the value at the path `/subnet/<subnet_id>/canister_ranges` in the state tree.
```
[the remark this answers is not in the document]
mraszyk: which are only

PR #164, opened 2023-05-30
add sender and ingress_expiry to the example request id calculation

mraszyk:
Corresponding replica code:
```
pub(crate) fn representation_indepent_hash_call_or_query(
    request_type: CallOrQuery,
    canister_id: Vec<u8>,
    method_name: &str,
    arg: Vec<u8>,
    ingress_expiry: u64,
    sender: Vec<u8>,
    nonce: Option<&[u8]>,
) -> [u8; 32] {
    use RawHttpRequestVal::*;
    let mut map = btreemap! {
        "request_type".to_string() => match request_type {
            CallOrQuery::Call => String("call".to_string()),
            CallOrQuery::Query => String("query".to_string()),
        },
        "canister_id".to_string() => Bytes(canister_id),
        "method_name".to_string() => String(method_name.to_string()),
        "arg".to_string() => Bytes(arg),
        "ingress_expiry".to_string() => U64(ingress_expiry),
        "sender".to_string() => Bytes(sender),
    };
    if let Some(some_nonce) = nonce {
        map.insert("nonce".to_string(), Bytes(some_nonce.to_vec()));
    }
    hash_of_map(&map)
}
```

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -745,30 +745,36 @@ E.g., request id consisting of bytes `[00, 01, 02, 03, 04, 05, 06, 07, 08, 09, 0
 
 [TIP]
 ====
-Example calculation (where `H` denotes SHA-256 and `·` denotes blob concatenation):
+Example calculation (where `H` denotes SHA-256 and `·` denotes blob concatenation) in which we assume that the optional nonce is not provided and thus omitted:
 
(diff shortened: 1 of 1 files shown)


Dfinity-Bjoern (approved):
Thank you for the correction!

PR #165, opened 2023-05-31
update WebAssembly module requirements

mraszyk:
This MR updates the "WebAssembly module requirements" section of the Interface specification.

- It removes notes about imported memories and tables as they cannot be imported on the replica:
```
unknown import: `env::memory` has not been defined
unknown import: `env::table` has not been defined
```
- It adds conditions on the type of `canister_pre_upgrade` and `canister_post_upgrade`.
- It refines the definition of "the total size of the custom sections".


Dfinity-Bjoern:
I am obviously not sufficiently versed in web assembly. Is it unnecessary to import the memory and function table?

mraszyk:
> I am obviously not sufficiently versed in web assembly. Is it unnecessary to import the memory and function table?

It's not only unnecessary, it's actually forbidden as otherwise the host environment (replica) would need to provide it. @ulan

mraszyk:
Mainnet implementation deviates in the following conditions:
- "It may not export other methods the names of which start with the prefix canister_ besides the methods allowed above." is not enforced.
- `canister_query` corresponds to `canister_query <name>` with empty name.

ulan:
The PR lgtm

I think these `env` imports are pre-Genesis leftovers. Wasm modules that import them are going to fail with
```
Failed to instantiate module: unknown import: `env::memory` has not been defined
```

> "It may not export other methods the names of which start with the prefix canister_ besides the methods allowed above." is not enforced.

We added this requirement to the spec after Genesis. There are already some canisters that violate this requirement, that's   why we couldn't enforce it on the replica side. (Otherwise we would break the canisters).

ulan (approved):

Dfinity-Bjoern:
Discussed in interface spec meeting. If main net does not enforce the rule for `canister_`, then we should also drop the requirement from the spec.

ulan:
> Discussed in interface spec meeting. If main net does not enforce the rule for `canister_`, then we should also drop the requirement from the spec.

Did you discuss the alternative of actually enforcing it on the mainnet? It would involve adding a whitelist for canisters that already break the rule, but enforcing it for all other canister.

mraszyk:
> > Discussed in interface spec meeting. If main net does not enforce the rule for `canister_`, then we should also drop the requirement from the spec.
> 
> Did you discuss the alternative of actually enforcing it on the mainnet? It would involve adding a whitelist for canisters that already break the rule, but enforcing it for all other canister.

How would we identify such canisters?

mraszyk:
Keeping the condition for now that the canister must not export methods whose names start with `canister_*` besides the methods explicitly allowed.

PR #47, opened 2022-05-16
Converting Interface Spec to markdown

aisconnolly:
Hi interface spec folks! 
With the new dev portal, we have been converting documents to markdown. Of course we want the reference documentation to live in their 'home' repos, so I am pushing this md version of the interface spec here. If it's ok with you to work from this file now, as this is what will be pulled into the github.com/dfinity/portal. 
What has been done here? 
- Conversion to md
- UML diagrams rendering
- Section cross-references are maintained 
- Code blocks and tables are tidied
- Changelog is imported at the end (this can otherwise just be linked.. as you prefer)
- Candid files are linked to download - will work on having them embedded in the page. 
You can see this version running here: https://smartcontracts.org/docs/current/references/ic-interface-spec/
Let me know if you've any questions or comments, and thanks for bearing with us as we get the new docs running!

PR #166, opened 2023-05-31
draft conversion to md

crusso:
Derived from motoko adoc conversion script, minus sed pre and postprocessing to prevent inlining of includes and other fixes (like using docusaurus admonitions).

See here for the original conversion.
https://github.com/dfinity/motoko/blob/14219adbe9d568e5d36332c4079c0ba3c4b8aa42/doc/convert.sh


nomeata:
How does this relate to https://github.com/dfinity/interface-spec/pull/47 and https://github.com/dfinity/interface-spec/pull/116? (Just curious)

mraszyk:
All open PRs have been converted to Markdown so there's no need to keep this PR open anymore.

PR #167, opened 2023-06-02
specify that canister version is incremented upon successful message execution

mraszyk:
The replica implementation already increments the canister version upon every successful message execution except for successful message execution of a query method.


spec/index.adoc
```diff
@@ -4075,19 +4076,6 @@ S with
...
-....
-    N0 = S.canister_version[CanisterId]
-    N1 > N0
-....
-State after::
-....
-S with
-    canister_version[CanisterId] = N1
-....
-
```
nomeata, line 4090: I'd suggest to leave it in, unless you really want people to rely on the one-increment-per-message feature. It gives the replica some leeway.

spec/index.adoc
```diff
@@ -3029,6 +3029,7 @@ if
 then
   S with
     canisters[M.receiver].wasm_state = res.new_state;
+    canister_version[M.receiver] = S.canister_version[M.receiver] + 1;
```
mraszyk, line 3032: Should the canister version increase if the method is a query method?

nomeata: Good question, and I am not sure. But probably yes: it's still a update or inter canister call; the fact that some annotation on the canister causes possible state changes to be thrown away is a detail at a different level of abstraction, inside the canister, so to say. So I assume it's more uniform to bump here too.

spec/index.adoc
mraszyk: Reverted the original change.

spec/index.adoc
mraszyk: In the interface spec meeting, we decided that we don't prescribe the canister version to increase after replicated query method execution. Because the canister version can also increase in other cases than those listed explicitly, it is not a breaking change to bump canister version after replicated query method execution in the future.

Dfinity-Bjoern (approved):
Makes sense to get this in line with the implemented protocol.

PR #172, opened 2023-06-07
RUN-636: Add per-call instruction counter

dfinity-berestovskyy:
The current execution instruction counter (performance counter type 0) gets reset at the beginning of each execution, i.e. at the beginning or each call, callback, or a system task. This makes the async code profiling difficult, as the profiling data must be stored across executions and then consolidated.

This new per-call instruction counter (performance counter type 1) is monotonically increasing across nested calls and await points, greatly facilitating the async code profiling and optimization.

### DIFF
--- a/spec/index.adoc
+++ b/spec/index.adoc
@@ -1461,19 +1461,17 @@ Passing zero as an argument to the function deactivates the timer and thus preve
 [#system-api-performance-counter]
 === Performance counter
 
-The canister can query the "performance counter", which is a deterministic monotonically increasing integer approximating the amount of work the canister has done since the beginning of the current execution.
+The canister can query one of the "performance counters", which is a deterministic monotonically increasing integer approximating the amount of work the canister has done. Developers might use this data to profile and optimize the canister performance.
 
 `+ic0.performance_counter : (counter_type : i32) -> i64+`
 
 The argument `type` decides which performance counter to return:
 
-* 0 : instruction counter. The number of WebAssembly instructions the system has determined that the canister has executed.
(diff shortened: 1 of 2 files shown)


mraszyk (commented):
I'd link the type 1 counter to the notion of "call context" so that its specification is well-defined.

spec/index.adoc
```diff
@@ -1461,19 +1461,17 @@ Passing zero as an argument to the function deactivates the timer and thus preve
...
 
 `+ic0.performance_counter : (counter_type : i32) -> i64+`
 
 The argument `type` decides which performance counter to return:
 
-* 0 : instruction counter. The number of WebAssembly instructions the system has determined that the canister has executed.
+* 0 : current execution instruction counter. The number of WebAssembly instructions the canister has executed since the beginning of the current <<rule-message-execution>>.
 
-In the future, we might expose more performance counters.
+* 1 : call context instruction counter. The number of WebAssembly instructions the canister has executed since the current <<call-context-creation>>. The counter monotonically increases across a few <<rule-message-execution,message executions>>, until the corresponding <<call-context-removal,call context is removed>>.
```
mraszyk, line 1472:
```suggestion
* 1 : call context instruction counter. The number of WebAssembly instructions the canister has executed within the call context of the current <<rule-message-execution>> since <<call-context-creation>>. The counter monotonically increases across all <<rule-message-execution,message executions>> in the call context until the corresponding <<call-context-removal,call context is removed>>.
```

mraszyk:
Closing in favor of #185 which is converted from AsciiDoc to MarkDown.

Issue #176, opened 2023-06-20
Description of reinstall vs uninstall+install needs clarification


timohanke:
https://github.com/dfinity/interface-spec/blob/6b7cd0a111e0c54407fcf2987d068d68e5ae958f/spec/index.adoc?plain=1#L1633

This appears to be saying that `install_code` with `mode = reinstall` does not do what `uninstall_code` does. So does it mean that `install_code` with `mode = reinstall` waits for all outstanding responses to come back and lets all non-fulfilled call contexts be fulfilled naturally before the code is reinstalled? I think this requires more explanation.

Also the text "that will forcibly reject all calls awaiting a response" should be "that will forcibly reject all calls that the canister has not yet responded to". If I understand correctly, it is possible for a call to have responded itself already but at the same to still be awaiting a response. In such a case there cannot be a forced reject, or?

nomeata:
> So does it mean that install_code with mode = reinstall waits for all outstanding responses to come back and lets all non-fulfilled call contexts be fulfilled naturally before the code is reinstalled? 

I think it simply swaps out the code, to be used for canisters that keep the information to deal with responses in stable memory (or need no such information, e.g. simply proxying canisters). We don't have any such canisters yet, I believe, as without named callbacks they are even harder to implement.

> that will forcibly reject all calls that the canister has not yet responded to

That's also a good phrasint, to me they mean the same thing (awaiting response = not yet responded to). Maybe both sentences work even better with ”incoming calls”?

timohanke:
> I think it simply swaps out the code, to be used for canisters that keep the information to deal with responses in stable memory (or need no such information, e.g. simply proxying canisters). We don't have any such canisters yet, I believe

Are you sure? I don't think so. But it seems to me now that I have been using canister reinstall all the time without understanding what it actually does.

> to me they mean the same thing (awaiting response = not yet responded to)

I understand them as different. I understand a call can make multiple downstream calls and wait for their responses. And at some point it responds itself to its caller (which fulfils the call context). So the call can be "awaiting responses" which refers to responses from its downstream calls. And the call can "not yet have responded" which refers to not yet have responded to its caller (or not yet have fulfilled its call context). But those two are different things.

nomeata:
[...]

PR #179, opened 2023-06-23
refine note on difference between reinstall and uninstall+install

mraszyk:
Closes https://github.com/dfinity/interface-spec/issues/176


nomeata (approved):
Great!

Dfinity-Bjoern (approved):
Thanks for the clarification!

PR #183, opened 2023-06-26
Initial spec changes for query stats feature

Dfinity-skaestle:
In order to learn about the performance of canister code, developers would like to gain insights into the size of query call payloads and instructions executed by query calls.

The "query stats" features adds the mechanisms to collect per-node query statistics, aggregate them deterministically and save the instruction counts in the replicated state.

Query stats will be made available to canister developers by means of the canister status API.

[Jira](https://dfinity.atlassian.net/browse/IC-1454)
[Design doc](https://docs.google.com/document/d/1awcOUG33wb5KPyjRhsnRgZDeRAO8Gv4xCI9KYYg-1bA/edit)


Dfinity-Bjoern:
Maybe I am missing something but shouldn't we specify some way for the canister to retrieve the stats? Also, please add a bit more context in the PR description above.

Dfinity-skaestle:
> Maybe I am missing something but shouldn't we specify some way for the canister to retrieve the stats? Also, please add a bit more context in the PR description above.

I added a bit of description to the Draft MR incl. explanation that we are planning to make query stats available to developers by means of the (existing) canister status API.

spec/index.md
```diff
@@ -1835,6 +1835,16 @@ Indicates various information about the canister. It contains:
 
 -   The cycle balance of the canister.
 
+-   Statistics regarding the query execution of the canister: a tuple `(num_queries, num_instructions, num_request_payload_bytes, num_response_payload_bytes)` containing
```
Dfinity-skaestle, line 1838: @Dfinity-Bjoern - this is where we are planning to make query stats available to developers.

mraszyk:
@Dfinity-skaestle I suppose we'd also need to count (and later charge) calls to `canister_inspect_message` which is not replicated and handled similarly to a query, e.g., in terms of scheduling.

Dfinity-skaestle:
> @Dfinity-skaestle I suppose we'd also need to count (and later charge) calls to `canister_inspect_message` which is not replicated and handled similarly to a query, e.g., in terms of scheduling.

We discussed this ans decided that while we probably want to do this in the long run, it's probably out of scope for the first feature. We might want to reconsider for follow-up features and when we actually do start charging for queries.

spec/index.md
```diff
@@ -5226,6 +5253,25 @@ verify_response(Q, R, Cert') ∧ lookup(["time"], Cert') = Found S.system_time /
...
+        num_request_payload_bytes = |Q.Arg|
+        num_response_payload_bytes =
+          if composite_query_helper(S, MAX_CYCLES_PER_QUERY, 0, Q.canister_id, Q.sender, Q.canister_id, Q.method_name, Q.arg) = (Reject (RejectCode, RejectMsg), _) then |RejectMsg|
+          else if composite_query_helper(S, MAX_CYCLES_PER_QUERY, 0, Q.canister_id, Q.sender, Q.canister_id, Q.method_name, Q.arg) = (Reply Res, _) then |Res|
+    }
+
+```
+
+where `NumInstructions` is the unspecified number of WebAssembly instructions executed during the evaluation of the query call,
+i.e., the value of `ic0.performance_counter(0)` at the end of the query method execution.
```
mraszyk, line 5273:
```suggestion
where `NumInstructions` is the unspecified number of WebAssembly instructions executed during the evaluation of the query call,
i.e., the sum of values of `ic0.performance_counter(0)` at the end of all query and composite query method executions during the evaluation of the query call.
```
@Dfinity-skaestle Is my above suggestion correct?

mraszyk:
Now that composite query methods executed during the evaluation of a query call can make downstream calls, we should make sure we unambiguously specify to the statistics of which canister the execution of those downstream calls is accounted.

spec/_attachments/ic.did
```diff
@@ -170,6 +170,12 @@ service ic : {
       memory_size: nat;
       cycles: nat;
       idle_cycles_burned_per_day: nat;
+      query_stats: vec record {
```
lwshang, line 173: There are two problems here:
1. The drafted implementation in the mono-repo seems to be inconsistent with the spec here:
    - This spec defines the `query_states` to be a Vector of the Record;
    - The implementation has no outer Vector

2. The extra field `query_stats` should be optional so that it can be backward-compatible:
```
query_stats: opt record {
  ...
}
```

mraszyk: Agreed with 1.

Ad 2. The new field is added to the response and thus old clients should be able to parse a response with the new field into the old type without the field.

spec/index.md
Dfinity-skaestle: For collecting query stats during composite query execution, I think it makes most sense to track each canister individually (instead of grouping in the root canister). E.g. if a users queries canister A which in turn queries canister B, we would record two query calls, one on canister A and one on B, with the respective number of instructions and request / response sizes.

spec/_attachments/ic.did
chenyan-dfinity: Yep, adding a new field in the return position is backward compatible. See Roman's explanation: https://mmapped.blog/posts/20-candid-for-engineers.html#subtyping

lwshang: "Adding a new field to a record in the method result type is backward compatible" is true for normal canister. Because the consumers always upgrade later than the provider.

But for the management canister, the case is different. Let's say, if the API in Rust CDK is updated and a canister using the new CDK is deployed to a network with old management canister API. The canister will fail to decode the return value because it expects a field which is not provided.

We already see some canister projects keep using old version of `dfx`. Such projects won't be able to upgrade their CDK unless they upgrade `dfx` at the same time.

chenyan-dfinity: > The canister will fail to decode the return value because it expects a field which is not provided.
No, that's not possible. The return value from the management canister *always* contains the new field. It's up to the client to decide if they want to decode that new field.

lwshang: > The return value from the management canister always contains the new field.

This will be true for the mainnet. But for those projects using old local replica from old `dfx` in their CI, if they upgrade Rust CDK, the CI will fail. And they will have to upgrade `dfx`.

If I did complete the support in Rust CDK, I wouldn't be able to merge the PR and release today. Because the mainnet doesn't have the feature yet. All the CDK/agent supports must be held to release until the mainnet upgrades.

In other words, how do we treat the case of "new client <-> old server"?
Should we simply ignore it? 
Or if we make the new field `opt`, this case will be handled gracefully. And we completely decouple the server and clients. Server and clients can release at their own paces.

chenyan-dfinity: That all seems to be expected. If they upgrade CDK, they need to upgrade dfx, and CDK changes should be done after it's been rolled out in the main net, or included in the dfx release.

> In other words, how do we treat the case of "new client <-> old server"?
Should we simply ignore it?

I think we can ignore it. It's too restrictive to require both backward and forward compatibility. OTOH, a good way to solve this problem is to make CDK to always fetch the latest did from the network, and generate the binding, instead of hardcoding this binding in a CDK release.

Dfinity-Bjoern (approved):

sa-github-api:
Dear @Dfinity-skaestle,

In order to potentially merge your code in this open-source repository and therefore proceed with your contribution, we need to have your approval on DFINITY's [CLA](https://github.com/dfinity/cla/blob/main/CLA.md).

If you decide to agree with it, please visit this [issue](https://github.com/dfinity/cla/issues/454) and read the instructions there. Once you have signed it, re-trigger the workflow on this PR to see if your code can be merged.

— The DFINITY Foundation

spec/index.md
```diff
@@ -4065,12 +4084,21 @@ S with
             S.freezing_threshold[A.canister_id],
             S.canister_subnet[A.canister_id].subnet_size,
           );
+          query_stats = noise(SUM {{num_calls_total: 1,
+                                    num_instructions_total: single_query_stats.num_instructions,
+                                    request_payload_bytes_total: single_query_stats.request_payload_bytes,
+                                    response_payload_bytes_total: single_query_stats.response_payload_bytes} |
+                                   single_query_stats <- S.query_stats[A.canister_id];
+                                   single_query_stats.timestamp <= S.time[A.canister_id] - T})
```
mraszyk, line 4092: @Sawchord Could you please confirm that the aggregation takes into account all query stats up until a recent point in time (due to query stats algorithm delay), i.e., that the aggregation does not run over a sliding window?

PR #185, opened 2023-06-28
RUN-636: Add call context instruction counter

mraszyk:
The current execution instruction counter (performance counter type 0) gets reset at the beginning of each execution, i.e. at the beginning or each call, callback, or a system task. This makes the async code profiling difficult, as the profiling data must be stored across executions and then consolidated.

This PR introduces a new per-call instruction counter (performance counter type 1) which is monotonically increasing across nested calls and await points, greatly facilitating the async code profiling and optimization.


Dfinity-Bjoern (approved):
Looks good from a spec perspective.

dfinity-berestovskyy:
LGTM, thanks!

PR #187, opened 2023-06-28
add canister_history to initial IC state and remove it when canister is deleted

mraszyk:
This PR fixes some omissions when specifying canister history.


oggy-dfin (approved):
LGTM

PR #191, opened 2023-06-28
feat: Add metrics about subnet usage on the state tree

dsarlis:
### Motivation

The Internet Computer [dashboard](https://dashboard.internetcomputer.org/) exposes various metrics that it retrieves from internal prometheus endpoints, including but not limited to: number of canisters on a subnet, memory used on a subnet, cycles burned, transactions made, etc.

The information shown in the dashboard cannot be verified by 3rd parties as it requires access to internally available only endpoints. As such, people must trust DFINITY who is building the dashboard that they provide correct information. We would like to expose some of these key metrics that show case the usage and capabilities of the IC in a way that is verifiable from external 3rd parties to foster a much more trustworthy ecosystem.

### Proposal

The proposed solution is to expose these metrics through the system state tree. The gist of the proposal is to add a path such as:
```
/subnet/<subnet_id>/metrics
```
which would return a blob of encoded data capturing the metrics that we would like to expose (could also consider a path per metric but might come with downsides, see my inline comment). This path would be exposed through a new `api/v2/subnet/<subnet_id>/read_state` endpoint. Having this endpoint allows us to move all other subtree paths under `/subnet` to be served through this endpoint as well as the `/time` path.


spec/index.md
```diff
@@ -427,6 +427,10 @@ The state tree contains information about the topology of the Internet Computer.
         principal = bytes .size (0..29)
         tagged<t> = #6.55799(t) ; the CBOR tag
 
+-   `/subnet/<subnet_id>/metrics` (blob)
```
dsarlis, line 430: This is not fully fleshed out. We could also imagine an alternative of a separate path per metric. There's probably a trade-off here between having a single struct with all metrics and how difficult it is to verify the certificate for a single blob or having to perform verification for every different path listed.

dsarlis:
> We are discussing in the interface spec meeting: Why not create a new API path `/api/v2/subnet/<principal>/read_state`?

Why would we? :P It's still a read from the system state tree so it seems natural to me to use the existing API path. That said, I don't feel strongly either way. I suppose if we go with a new API path, then the discussion about effective principal id is moot because presumably the `principal` is enough to route the request correctly?

dsarlis:
> > We are discussing in the interface spec meeting: Why not create a new API path `/api/v2/subnet/<principal>/read_state`?
> 
> Why would we? :P It's still a read from the system state tree so it seems natural to me to use the existing API path. That said, I don't feel strongly either way. I suppose if we go with a new API path, then the discussion about effective principal id is moot because presumably the `principal` is enough to route the request correctly?

Actually I think it makes more sense this way given the existing API path is `api/v2/canister/...`, I'll switch it up.

dsarlis:
@Dfinity-Bjoern @mraszyk I'm re-working the PR to incorporate the new API path `/api/v2/subnet/<principal>/read_state`. If we go ahead with this, I believe it makes sense to change the state tree paths used to retrieve other subnet specific information. In particular, I'd suggest we change the following

```
`/subnet/<subnet_id>/public_key`
`/subnet/<subnet_id>/canister_ranges`
`/subnet/<subnet_id>/metrics`
```
to

```
`/public_key`
`/canister_ranges`
`/metrics`
```
when accessed through `/api/v2/subnet/<principal>/read_state` and expect that this would be the new approach. The reason being that it's redundant if we have the api path contain the subnet id already. The old paths can remain available under the old http api path as well for a while until clients migrate and we can eventually remove them and have them only be accessible through the new api path. Wdyt?

mraszyk:
> @Dfinity-Bjoern @mraszyk I'm re-working the PR to incorporate the new API path `/api/v2/subnet/<principal>/read_state`. If we go ahead with this, I believe it makes sense to change the state tree paths used to retrieve other subnet specific information. In particular, I'd suggest we change the following
> 
> ```
> `/subnet/<subnet_id>/public_key`
> `/subnet/<subnet_id>/canister_ranges`
> `/subnet/<subnet_id>/metrics`
> ```
> 
> to just
> 
> ```
> `/public_key`
> `/canister_ranges`
> `/metrics`
> ```
> 
> when accessed through `/api/v2/subnet/<principal>/read_state` and expect that this would be the new approach long term. The reason being that it's redundant if we have the api path contain the subnet id already. The old paths can remain available under the old http api path as well for a while until clients migrate and we can eventually remove them and have them only be accessible through the new api path. Wdyt?

I don't think it'd work unless the subnet signs another certified tree where the leaves are just at the root. Moreover, we should make it possible to browse all subnets (the entire /subnet subtree) without querying the NNS registry canister.

What makes sense (but it's not backward-compatible) is to move the entire /subnet to the new API (with /time being available in both and always included).

CC @derlerd-dfinity

dsarlis:
> I don't think it'd work unless the subnet signs another certified tree where the leaves are just at the root. Moreover, we should make it possible to browse all subnets (the entire /subnet subtree) without querying the NNS registry canister.

Hmm, yeah, good point.

> What makes sense (but it's not backward-compatible) is to move the entire /subnet to the new API (with /time being available in both and always included).

Why can't we serve the entire /subnet through both APIs for a while to allow transition and then stop serving from the old one at some point?

mraszyk:
> Why can't we serve the entire /subnet through both APIs for a while to allow transition and then stop serving from the old one at some point?

Sure, we can do that (and maybe we just do that indefinitely).

spec/index.md
```diff
@@ -4546,22 +4556,52 @@ A record with
...
 
-The predicate `may_read_path` is defined as follows, implementing the access control outlined in [Request: Read state](#http-read-state):
 
-    may_read_path(S, _, ["time"] · _) = True
-    may_read_path(S, _, ["subnet"] · _) = True
-    may_read_path(S, _, ["request_status", Rid] · _) =
+The predicate `may_read_path_for_canister` is defined as follows, implementing the access control outlined in [Request: Read state](#http-read-state):
+
+    may_read_path_for_canister(S, _, ["time"] · _) = False
+    may_read_path_for_canister(S, _, ["subnet"] · _) = False
```
dsarlis, line 4563: I believe we can actually drop these from here and allow implementations to serve these paths from both endpoints for backward compatibility. But I think it's also ok to have them be served by both endpoints. I'll go with whatever is the spec owner preference here.

dsarlis (commented):
@Dfinity-Bjoern @mraszyk I've updated the PR to reflect that we are now proposing a new api endpoint that includes the subnet id in its URI. Let me know what you think of this version and if we need to cover something else (specifically in the prose section). Or perform editorial changes as you see fit if it's faster.

spec/index.md
```diff
@@ -427,6 +427,16 @@ The state tree contains information about the topology of the Internet Computer.
         principal = bytes .size (0..29)
         tagged<t> = #6.55799(t) ; the CBOR tag
 
+-   `/subnet/<subnet_id>/metrics` (blob)
+
+     A list of subnet wide metrics related to this subnet's current usage and/or performance. The metrics are a CBOR map with the following fields:
```
mraszyk, line 432:
```suggestion
     A collection of subnet-wide metrics related to this subnet's current resource usage and/or performance. The metrics are a CBOR map with the following fields:
```
I'm using "collection" instead of "list" as the blob is a CBOR "map".

spec/index.md
```diff
@@ -427,6 +427,16 @@ The state tree contains information about the topology of the Internet Computer.
...
+     
+     - `num_canisters` (`nat`): The number of canisters on this subnet.
+     - `canister_state` (`nat`): The total state taken by canisters on this subnet in bytes.
```
mraszyk, line 435:
```suggestion
     - `canister_state` (`nat`): The total size of the state taken by canisters on this subnet in bytes.
```

spec/index.md
mraszyk: If we define `False` here, then we it shouldn't be permitted to serve those paths (what's the point of this predicate if it was permitted anyway?).

spec/index.md
```diff
@@ -4546,22 +4556,52 @@ A record with
...
+-   `{certificate: C}`
+  
+
+The predicate `may_read_path_for_subnet` is defined as follows, implementing the access control outlined in [Request: Read state](#http-read-state):
+
+    may_read_path_for_subnet(S, _, ["time"] · _) = True
+    may_read_path_for_subnet(S, _, ["subnet"] · _) = True
+    may_read_path_for_subnet(S, _, _) = False
 
 where `UTF8(name)` holds if `name` is encoded in UTF-8.
```
mraszyk, line 4606: not needed here

spec/index.md
```diff
@@ -4546,22 +4556,52 @@ A record with
...
+
+```
+
+Read response  
+A record with
+
+-   `{certificate: C}`
+  
+
+The predicate `may_read_path_for_subnet` is defined as follows, implementing the access control outlined in [Request: Read state](#http-read-state):
```
mraszyk, line 4600: Could you please refine the sentence "The subnet form should be used when the information to be retrieved is subnet specific" to hint at what paths are actually allowed in that case.

spec/index.md
```diff
@@ -4536,7 +4546,7 @@ Conditions
 E.content = ReadState RS
 TS = verify_envelope(E, RS.sender, S.system_time)
 S.system_time <= RS.ingress_expiry
-∀ path ∈ RS.paths. may_read_path(S, R.sender, path)
+∀ path ∈ RS.paths. may_read_path_for_canister(S, R.sender, path)
```
mraszyk, line 4549:
```suggestion
∀ path ∈ RS.paths. may_read_path_for_canister(S, RS.sender, path)
```
You didn't introduce this typo, but why not fix it?

spec/index.md
```diff
@@ -4546,22 +4556,52 @@ A record with
...
+`E`
+
+Conditions  
+
+```html
+
+E.content = ReadState RS
+TS = verify_envelope(E, RS.sender, S.system_time)
+S.system_time <= RS.ingress_expiry
+∀ path ∈ RS.paths. may_read_path_for_subnet(S, R.sender, path)
```
mraszyk, line 4590:
```suggestion
∀ path ∈ RS.paths. may_read_path_for_subnet(S, RS.sender, path)
```
You didn't introduce this typo, but why not fix it?

spec/index.md
```diff
@@ -427,6 +427,16 @@ The state tree contains information about the topology of the Internet Computer.
         principal = bytes .size (0..29)
         tagged<t> = #6.55799(t) ; the CBOR tag
 
+-   `/subnet/<subnet_id>/metrics` (blob)
```
mraszyk, line 430: A danger of this endpoint is that replica-signed queries require read_state requests to `/subnet` and thus its responses would now also serve the metrics (not a problem is we don't have many metrics, but could be a problem in the future).

dsarlis: I am not sure I'm following what's the problem. I suppose for replica-signed queries you don't strictly need the metrics but the metrics are also not a secret, so returning them seems ok security wise.

spec/index.md
```diff
@@ -427,6 +427,16 @@ The state tree contains information about the topology of the Internet Computer.
...
+
+     A list of subnet wide metrics related to this subnet's current usage and/or performance. The metrics are a CBOR map with the following fields:
+     
+     - `num_canisters` (`nat`): The number of canisters on this subnet.
+     - `canister_state` (`nat`): The total state taken by canisters on this subnet in bytes.
+     - `consumed_cycles` (`nat`): The total number of cycles consumed by canisters on this subnet.
```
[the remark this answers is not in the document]
dsarlis: These would be counters counting from when the subnet was created (not sure if that's what you meant by genesis here).

spec/index.md
dsarlis: Ok, maybe I wasn't clear on my intention. I think it's fine from a spec point of view that we drop these from here (and therefore they would be caught in the last predicate below which is the catch all). This would mean that these paths should not be permitted given a new version of the spec released with this change. However, the replica implementation should not immediately drop support for those, otherwise we would have a breaking change (which I would like to avoid).

spec/index.md
dsarlis: Let me know what you think of the current version.

spec/index.md
Dfinity-Bjoern: It is not about security but about data size. If the subnets contain all public keys of the nodes hosting the subnet, though, that that is a lot more data (~ 1KB per subnet) so the metrics at this point don't really add a lot of data on top of this.

dsarlis: I see, thanks for the clarification. Yeah, I don't expect the metrics will ever be anything other than a few counters, so something in the order of hundreds of bytes extra in the worst case.

spec/index.md
mraszyk: Thank you! Pushed a slight reformulation.

spec/index.md
mraszyk: Discussed in the Interface spec meeting: we agreed to keep the `/time` and `/subnet` paths available via the `/api/v2/canister` URL, too (the `/time` is anyway always included and it makes sense to allow `/subnet` for backward-compatibility, too).

spec/index.md
```diff
@@ -439,6 +439,16 @@ The state tree contains information about the topology of the Internet Computer.
...
         tagged<t> = #6.55799(t) ; the CBOR tag
 
+-   `/subnet/<subnet_id>/metrics` (blob)
+
+     A collection of subnet-wide metrics related to this subnet's current resource usage and/or performance. The metrics are a CBOR map with the following fields:
+     
+     - `num_canisters` (`nat`): The number of canisters on this subnet.
+     - `total_canister_state` (`nat`): The total size of the state in bytes taken by canisters on this subnet since this subnet was created.
+     - `total_consumed_cycles` (`nat`): The total number of cycles consumed by canisters on this subnet since this subnet was created.
+     - `num_update_transactions` (`nat`): The total number of transactions processed on this subnet since this subnet was created.
```
alin-at-dfinity, line 449: I would consider whether it makes sense to follow [Prometheus naming best practices](https://prometheus.io/docs/practices/naming/#metric-names) (or similar) when naming these metrics. Among other things, it's generally a good idea to include the measurement unit in the name (e.g. `total_canister_state_bytes` rather than a confusing `total_canister_state`). And it is good to identify counters, e,g, by giving them a `total` suffix (e.g. `total_consumed_cycles_total` or ``update_transactions_total`).

dsarlis: I saw this late and the change has already been merged on the replica side. Given that these are not prometheus metrics I don't think we're bound to follow the best practices it establishes. We explain what the metrics represent so that should be ok. If spec owners feel strongly, I'm happy to change (although the hassle of changing some names given we're talking about certified state is likely not worth it imo).

mraszyk: I think Alin's suggestion to make the metric names more self-explanatory makes sense so I'd be up for it if it's still possible.

dsarlis: Are the following changes ok with everyone?

`total_canister_state` -> `total_canister_state_bytes`
`num_update_transactions` -> `update_transactions_total`

The suggestion of `total_consumed_cycles_total` seems confusing, at least to me (there's two instances of "total" in the name).

alin-at-dfinity: I don't mind if we don't follow Prometheus naming practices specifically. The `total` can go at the end or at the start (or be replaced with something else, e.g. `all_time`). I would very much prefer it for the naming to be consistent, though. I.e. don't use `total` in both gauges and counters; and don't put it at the beginning of one metric name and the end of another.

One possible naming scheme (although, again, I'm not insisting it be followed to the letter) could be:
```
canisters
canister_state_bytes
consumed_cycles_total
update_transactions_total
```

alin-at-dfinity: One technical aspect of this (which David pointed out) is that since the metrics are encoded as a blob; and the canonical state uses packed CBOR encoding; the names can be changed at any time, without breaking compatibility.

E.g. `{"num_canisters": 5, "total_canister_state_bytes": 100}` would be encoded as `{0: 5, 1: 100}`, which is the same as what `{"canisters": 5, "canister_state_bytes": 100}` would encode to (as long as we preserved field ordering).

spec/index.md
```diff
@@ -439,6 +439,16 @@ The state tree contains information about the topology of the Internet Computer.
...
         tagged<t> = #6.55799(t) ; the CBOR tag
 
+-   `/subnet/<subnet_id>/metrics` (blob)
+
+     A collection of subnet-wide metrics related to this subnet's current resource usage and/or performance. The metrics are a CBOR map with the following fields:
+     
+     - `num_canisters` (`nat`): The number of canisters on this subnet.
+     - `total_canister_state` (`nat`): The total size of the state in bytes taken by canisters on this subnet since this subnet was created.
+     - `total_consumed_cycles` (`map`): The total number of cycles consumed by canisters on this subnet since this subnet was created. It's a map of two values, a low part of type `nat` and a high part of type `opt nat`.
+     - `num_update_transactions` (`nat`): The total number of transactions processed on this subnet since this subnet was created.
```
alin-at-dfinity, line 449: Not sure how transactions are counted, but consumed cycles are summed across all canisters (to which the cycles burned by deleted canisters are added). Currently, this is indeed "cycles consumed [...] since this subnet was created". But if we consider subnet splitting/canister migration, this no longer holds.

Maybe it makes sense to reword this (or maybe not). That being said, I don't have a suggestion for how to make this both generic and concise. Maybe "the sum of all-time burned cycles by all current and deleted canisters".

spec/index.md
dsarlis: Consumed cycles is a gauge and update transactions is a counter, so your suggestion is inconsistent with itself if we use `total` in both. If we are ok with that, then I am ok with everything else except `canisters`, I'd prefer to keep `num_canisters` as it's more descriptive imo.

spec/index.md
dsarlis: @mraszyk any thoughts on this? I'm ok with "the sum of all-time burned cycles by all current and deleted canisters".

Re update_transactions: we keep accumulating them in a counter at the end of every round execution. So, it would include indeed all transactions since the subnet was created (even ones that were for canisters that were migrated away).

spec/index.md
alin-at-dfinity: Consumed cycles is a gauge only because it's not monotonically increasing (we subtract what are essentially reserved cycles and then return part of them later on). But leaving aside that awkwardness, it is essentially a counter, in that the absolute value is not particularly meaningful; what is meaningful about it by how much it changes over a day or an hour.

So it's all up in the air whether we want to call it a counter or a gauge. I would go with counter, since we are at very low risk of someone treating it strictly as a Prometheus counter and e.g. applying Prometheus' definition of `rate()` to it. People are just going to take the value now minus the value one hour ago. They may be surprised to see a negative difference, but that's about it.

dsarlis: Alright, I buy this argument. I'll update to:

```
num_canisters
canister_state_bytes
consumed_cycles_total
update_transactions_total
```

spec/index.md
mraszyk: I'm also fine with the formulation "the sum of all-time burned cycles by all current and deleted canisters".

spec/index.md
dsarlis: Updated.

spec/index.md
dsarlis: Ok, changed it to the proposed formulation.

alin-at-dfinity (approved):
LGTM, thank you.

Dfinity-Bjoern (approved):

spec/index.md
```diff
@@ -4670,33 +4680,66 @@ A record with
...
-    may_read_path(S, _, ["request_status", Rid, "error_code"]) =
+
+The predicate `may_read_path_for_canister` is defined as follows, implementing the access control outlined in [Request: Read state](#http-read-state):
+
+    may_read_path_for_canister(S, _, ["time"]) = True
+    may_read_path_for_canister(S, _, ["subnet"]) = True
+    may_read_path_for_canister(S, _, ["subnet", sid]) = True
+    may_read_path_for_canister(S, _, ["subnet", sid, "public_key"]) = True
+    may_read_path_for_canister(S, _, ["subnet", sid, "canister_ranges"]) = True
+    may_read_path_for_canister(S, _, ["subnet", sid, "metrics"]) = True
```
dsarlis, line 4691: @mraszyk I would like to get a confirmation on what we are expecting to do with allowing these under the canister endpoint. In the implementation, I would definitely still allow them for backwards compatibility (and allow end users to transition to the new endpoint) but I think we could also already remove them from the spec already.

The metrics one seems safe to only allow in the new endpoint given that it's a new path on the tree, so at the very least this one can be removed from here.

On a separate note: I've seen in the code that we also allow some paths of the form `["subnet", sid, "node", node_id])`, is this some other PR still in progress, so it doesn't appear here?

mraszyk: > I would like to get a confirmation on what we are expecting to do with allowing these under the canister endpoint. In the implementation, I would definitely still allow them for backwards compatibility (and allow end users to transition to the new endpoint) but I think we could also already remove them from the spec as we are providing them through the new subnet endpoint.

I would keep them in the spec as long as we keep them in the actual implementation. If we want to deprecate them (by forbidding the paths in the HTTP handler), then I'd announce this intention as a note in the spec (at the moment we're just saying that the subnet form "should" be used and we can strengthen this statement in a further note).

> The metrics path seems safe to only allow in the new endpoint given that it's a new path on the tree, so at the very least this one can be removed from here.

Makes sense. Done

> On a separate note: I've seen in the code that we also allow some paths of the form ["subnet", sid, "node", node_id]), is this some other PR still in progress, so it doesn't appear here?

Indeed. Those paths are specified in this [PR](https://github.com/dfinity/interface-spec/pull/163) that could not be merged yet because the replica changes are still on hold due to performance benchmarks.

dsarlis: > I would keep them in the spec as long as we keep them in the actual implementation. If we want to deprecate them (by forbidding the paths in the HTTP handler), then I'd announce this intention as a note in the spec (at the moment we're just saying that the subnet form "should" be used and we can strengthen this statement in a further note).

I'm fine with this, can you please edit that part in a way that it makes sense to you?

mraszyk: I've added a deprecation announcement and a note. Please close this thread if the formulation makes sense to you.

PR #193, opened 2023-06-29
fix definition of lookup

mraszyk:
This PR fixes the definition of `lookup_path` and replaces `lookup` by `lookup_path` incl. fixing arguments where appropriate. The issue has been reported by @oggy-dfin .

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -354,7 +354,7 @@ This section makes forward references to other concepts in this document, in par
 
     -   The `certificate` must be a valid certificate as described in [Certification](#certification), with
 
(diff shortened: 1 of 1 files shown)


nomeata (approved):

oggy-dfin (approved):
Thanks! `check_delegation` still doesn't really type, but it's wrong anyways since it doesn't cover the canister ranges. But we can merge this and fix that later.

mraszyk:
> check_delegation still doesn't really type

right, also because verify_cert can potentially fail and this error condition is not reflected in the return type

PR #198, opened 2023-07-08
redefine lookup_path if path extends past a leaf in the tree

mraszyk:
Suppose a tree has a path `/a/b` ending in a leaf with the value "secret". According to the current definition, `lookup_path` for the path `a/b/c` returns `Absent` and a pruned tree (witness for the lookup) must contain the leaf value "secret" although the value at that path was not requested. This PR redefines `lookup_path` (more precisely, a helper function `find_label` used by `lookup_path`) to return `Unknown` in that case so that a pruned tree (witness for the lookup) can actually have the leaf value "secret" pruned and the lookup in the pruned tree returns `Unknown` just like the lookup in the original tree (also returning `Unknown` after the change from this PR).

Note. The issue motivating this PR could be solved in a better way if it was possible to only prune the leaf value while still indicating that the tree has a leaf at a certain place: then we could keep the current definition of `lookup_path` yielding `Absent` (which is more precise than `Unknown`) without returning the actual leaf value "secret" (which is desirable as the value at this path was not requested). However, this would be a _breaking_ change (as it would require to change the tree datatype) and thus difficult to roll out in practice.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -2165,7 +2165,7 @@ find_label(l, _ · Labeled l1 t · _)                | l == l1     = Found t
 find_label(l, _ · Labeled l1 _ · Labeled l2 _ · _) | l1 < l < l2 = Absent
 find_label(l,                    Labeled l2 _ · _) |      l < l2 = Absent
 find_label(l, _ · Labeled l1 _ )                   | l1 < l      = Absent
-find_label(l, [Leaf _])                                          = Absent
+find_label(l, [Leaf _])                                          = Error
 find_label(l, [])                                                = Absent
 find_label(l, _)                                                 = Unknown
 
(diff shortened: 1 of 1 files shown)


nomeata:
Sounds reasonable. In my book, such a request is simply ”wrongly typed” and means someone has a wrong schema in mind, so as per garbage-in-garbage-out, either implementation is fine.

(If I could go back I'd probably just allow both values and subtrees at a certain path, more like urls and less like file systems; it would make a few things easier. But too late now, and not too bad.)

mraszyk:
This [PR](https://github.com/dfinity/interface-spec/pull/216) forbids requesting paths that are extensions of paths containing values in the state tree. Hence, I'm closing this PR as the definition of `lookup` for such paths does not affect externally visible behavior.

PR #201, opened 2023-07-11
update contexts in which ic0.msg_caller_size and ic0.msg_caller_copy can be called

mraszyk:
It does not make sense to prevent retrieving the caller in response (reply and reject) and cleanup callbacks so this PR makes this possible. In particular, calling `ic_cdk::caller()` after an await point should not trap in an update method. Moreover, this PR makes `ic_principal` (the management canister ID) the caller of system tasks and their callbacks.


dsarlis:
Reading this I'm left with a question: in case a call is made from a heartbeat/timer (i.e. a system task), there's no caller and indeed for context `T` we don't allow to call `ic_cdk::caller()`. However, if we're in a callback triggered by a system task, at least reading the spec, I might get the impression that the caller can be retrieved when in reality there isn't any caller. Can we clarify this point somehow?

mraszyk:
> I see we have now specific callback contexts for composite queries so perhaps the easiest is to introduce specific callback contexts for system tasks?

That would indeed be an option, but I see an even easier way around the issue:
- allow calling `ic0.msg_caller_size` and `ic0.msg_caller_copy` in system tasks `T` as well as response callbacks `Ry` and `Rt` and cleanup callback `C`;
- define the caller as the management canister (`aaaaa-aa`), i.e., the system, for system tasks, both in their entry point (context `T`), response callbacks (contexts `Ry` and `Rt`), and cleanup callbacks (context `C`).

It seems to me that this would might also simplify the replica code change as no optional caller type are necessary, although some additional test(s) might be necessary.

@dsarlis What do you think about my suggestion?

Dfinity-Bjoern:
I have a slight preference for using `aaaaa-aa`, because of the following: You could be using the same function as a callback from both `U` and `T` contexts, and if calling a system API in one of the two cases traps and in the other it doesn't, it's practically not usable. There's of course another way of resolving this: making the call context explicit to the canister through an additional API.

dsarlis:
> I have a slight preference for using `aaaaa-aa`, because of the following: You could be using the same function as a callback from both `U` and `T` contexts, and if calling a system API in one of the two cases traps and in the other it doesn't, it's practically not usable.

I'm ok having `aaaaa-aa` as the caller, the use case you brought up is a good one. We can make the change.

> There's of course another way of resolving this: making the call context explicit to the canister through an additional API.

The call context maintained by the system is relatively complex data structure, I'm not sure I'd go down that path because it's gonna require some thinking how we can encode the different aspects. The caller seems to be the most useful part for canisters anyway.

Dfinity-Bjoern (approved):
Thank you @mraszyk! Let us coordinate with the execution team but otherwise this looks good!

PR #202, opened 2023-07-11
install_code checks freezing_limit

### LINKED ISSUE #199: Code installation should check the freezing threshold
Currently code installation succeeds as long as the used cycles does not exceeds the balance of the canister. This is dangerous because the canister may get uninstalled if the controller does an expensive upgrade or if the upgrade increases the memory usage of the canister.

Code installation should check the freezing threshold like regular update messages do.

Details:
- https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-management-canister-code-installation
- `Cycles_used ≤ S.balances[A.canister_id]`

This issue is related to Issue #184

mraszyk:
Closes #199

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1836,6 +1836,8 @@ The `wasm_module` field specifies the canister module to be installed. The syste
 
 The optional `sender_canister_version` parameter can contain the caller's canister version. If provided, its value must be equal to `ic0.canister_version`.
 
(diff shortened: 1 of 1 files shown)


mraszyk:
Note. The replica code change has already been [merged](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/13444).

PR #206, opened 2023-07-26
drop formal model in Isabelle/HOL

mraszyk:
We decided to hold off further development of the formal model of the Interface specification in Isabelle/HOL so this PR removes the formal model from the repository (to make sure we don't have outdated models in the repository).


nomeata:
Will you put it in a separate public repo? Not sure what for, but it would be nice not to lose it.

mraszyk:
> Will you put it in a separate public repo? Not sure what for, but it would be nice not to lose it.

We'll have it in the git history of this repository.

Dfinity-Bjoern (approved):
Sad to see this go 😢 but practically it is the right move.

PR #207, opened 2023-07-26
rephrase description of create_canister method

mraszyk:
The existing description of the `create_canister` method caused some confusion so this PR rephrases the description.


LaraAS (approved):
LGTM, thanks for following up so quickly!

Dfinity-Bjoern (approved):

PR #210, opened 2023-07-29
update note about ingress history pruning

mraszyk:


mraszyk:
Closing this PR as requests are indeed kept around (in the state "Done") also after pruning the ingress history.

PR #211, opened 2023-07-31
add note on supported certificate versions in HTTP Gateway

mraszyk:
This PR clarifies under which conditions the HTTP Gateway can allow for arbitrary certification version in the canister's response.


Dfinity-Bjoern (approved):
Thank you @mraszyk

PR #215, opened 2023-08-09
specify IC management method for retrieving metrics

mraszyk:
This PR introduces a new IC management canister method called `node_metrics_history` for retrieving history of metrics for all nodes on a given subnet.

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -198,7 +198,7 @@ service ic : {
     requested_timestamp_nanos: nat64;
   }) -> (record {
     timestamp_nanos : nat64;
-    metrics : vec metrics;
+    node_metrics : vec node_metrics;
   });
 
   // provisional interfaces for the pre-ledger world
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -199,9 +199,9 @@ service ic : {
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
   // metrics interface
-  node_metrics : (record {
+  node_metrics_history : (record {
     subnet_id : principal;
-    requested_timestamp_nanos: nat64;
+    start_at_timestamp_nanos: nat64;
   }) -> (vec record {
     timestamp_nanos : nat64;
     node_metrics : vec node_metrics;
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -193,7 +193,7 @@ service ic : {
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
   // metrics interface
-  metrics : (record {
+  node_metrics : (record {
     subnet_id : principal;
     requested_timestamp_nanos: nat64;
   }) -> (record {
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -114,8 +114,8 @@ type millisatoshi_per_byte = nat64;
 type metrics = variant {
   node_metrics : record {
     node_id : principal;
-    num_blocks : nat64;
-    num_block_failures : nat64;
+    num_blocks_total : nat64;
+    num_block_failures_total : nat64;
   }
 };
 
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -194,7 +194,7 @@ service ic : {
   node_metrics : (record {
     subnet_id : principal;
     requested_timestamp_nanos: nat64;
-  }) -> (record {
+  }) -> (vec record {
     timestamp_nanos : nat64;
     node_metrics : vec node_metrics;
   });
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -114,9 +114,8 @@ type millisatoshi_per_byte = nat64;
 type metrics = variant {
   node_metrics : record {
     node_id : principal;
-    start_timestamp_nanos : nat64;
-    end_timestamp_nanos : nat64;
-    num_proposed_blocks : nat;
+    num_blocks : nat64;
+    num_block_failures : nat64;
   }
 };
 
@@ -197,6 +196,7 @@ service ic : {
   metrics : (record {
     subnet_id : principal;
   }) -> (record {
+    timestamp_nanos : nat64;
     metrics : vec metrics;
   });
 
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -111,12 +111,10 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
-type node_metrics = variant {
-  node_metrics : record {
+type node_metrics = record {
     node_id : principal;
     num_blocks_total : nat64;
     num_block_failures_total : nat64;
-  }
 };
 
 service ic : {
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -195,6 +195,7 @@ service ic : {
   // metrics interface
   metrics : (record {
     subnet_id : principal;
+    requested_timestamp_nanos: nat64;
   }) -> (record {
     timestamp_nanos : nat64;
     metrics : vec metrics;
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -111,7 +111,7 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
-type metrics = variant {
+type node_metrics = variant {
   node_metrics : record {
     node_id : principal;
     num_blocks_total : nat64;
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -111,6 +111,15 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type metrics = variant {
+  node_metrics : record {
+    node_id : principal;
+    start_timestamp_nanos : nat64;
+    end_timestamp_nanos : nat64;
+    num_proposed_blocks : nat;
+  }
+};
+
 service ic : {
   create_canister : (record {
     settings : opt canister_settings;
@@ -183,6 +192,13 @@ service ic : {
   bitcoin_get_utxos: (get_utxos_request) -> (get_utxos_response);
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
+
+  // metrics interface
+  metrics : (record {
+    subnet_id : principal;
+  }) -> (record {
+    metrics : vec metrics;
+  });
 
   // provisional interfaces for the pre-ledger world
   provisional_create_canister_with_cycles : (record {
--- a/spec/index.md
+++ b/spec/index.md
@@ -2034,7 +2034,7 @@ The node metrics management canister API is considered EXPERIMENTAL. Canister de
 
 :::
 
-Given a subnet ID as input, this method returns a collection of metrics for all nodes on the given subnet (field `node_metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. This timestamp is the earliest timestamp after the requested timestamp (field `requested_timestamp_nanos`) for which node metrics are available.
+Given a subnet ID as input, this method returns a collection of metrics for all nodes on the given subnet (field `node_metrics`) and timestamps provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. The returned timestamps are all timestamps after (and including) the requested timestamp (field `requested_timestamp_nanos`) for which node metrics are available.
 
 A single metric entry is a record with the following fields:
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -4163,7 +4163,7 @@ S.messages = Older_messages · CallMessage M · Younger_messages
 (M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
 M.callee = ic_principal
 M.method_name = 'metrics'
-M.arg = candid()
+M.arg = candid(A)
 R = <implementation-specific>
 
 ```
--- a/spec/index.md
+++ b/spec/index.md
@@ -2046,6 +2046,24 @@ Any user can top-up any canister this way.
 
 This method is only available in local development instances.
 
+### IC method `node_metrics` {#ic-node-metrics}
+
+:::note
+
+The node metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
+:::
+
+Given a subnet ID as input, this method returns a collection of metrics for all nodes on the given subnet (field `node_metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. This timestamp is the earliest timestamp after the requested timestamp (field `requested_timestamp_nanos`) for which node metrics are available.
+
+A single metric entry is a record with the following fields:
+
+- `node_id` (`principal`): the principal characterizing a node;
+
+- `num_blocks_total` (`nat64`): the number of blocks proposed by this node;
+
+- `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
+
 ## The IC Bitcoin API {#ic-bitcoin-api}
 
 The Bitcoin functionality is exposed via the management canister. Information about Bitcoin can be found in the [Bitcoin developer guides](https://developer.bitcoin.org/devguide/). Invoking the functions of the Bitcoin API will cost cycles. We refer the reader to the [Bitcoin documentation](https://internetcomputer.org/docs/current/developer-docs/integrations/bitcoin/bitcoin-how-it-works) for further relevant information and the [IC pricing page](https://internetcomputer.org/docs/current/developer-docs/gas-cost) for information on pricing for the Bitcoin mainnet and testnet.
@@ -2113,24 +2131,6 @@ The transaction fees in the Bitcoin network change dynamically based on the numb
 This function returns fee percentiles, measured in millisatoshi/vbyte (1000 millisatoshi = 1 satoshi), over the last 10,000 transactions in the specified network, i.e., over the transactions in the last approximately 4-10 blocks.
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
-
-### IC method `node_metrics` {#ic-node-metrics}
-
-:::note
-
-The node metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
-
-:::
-
-Given a subnet ID as input, this method returns a collection of metrics for all nodes on the given subnet (field `node_metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. This timestamp is the earliest timestamp after the requested timestamp (field `requested_timestamp_nanos`) for which node metrics are available.
-
-A single metric entry is a record with the following fields:
-
-- `node_id` (`principal`): the principal characterizing a node;
-
-- `num_blocks_total` (`nat64`): the number of blocks proposed by this node;
-
-- `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
 
 ## Certification {#certification}
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -2026,6 +2026,24 @@ If you do not specify the `max_response_bytes` parameter, the maximum of a `2MB`
 
 :::
 
+### IC method `node_metrics` {#ic-node-metrics}
+
+:::note
+
+The node metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
+:::
+
+Given a subnet ID as input, this method returns a collection of metrics for all nodes on the given subnet (field `node_metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. This timestamp is the earliest timestamp after the requested timestamp (field `requested_timestamp_nanos`) for which node metrics are available.
+
+A single metric entry is a record with the following fields:
+
+- `node_id` (`principal`): the principal characterizing a node;
+
+- `num_blocks_total` (`nat64`): the number of blocks proposed by this node;
+
+- `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
+
 ### IC method `provisional_create_canister_with_cycles` {#ic-provisional_create_canister_with_cycles}
 
 As a provisional method on development instances, the `provisional_create_canister_with_cycles` method is provided. It behaves as `create_canister`, but initializes the canister's balance with `amount` fresh cycles (using `DEFAULT_PROVISIONAL_CYCLES_BALANCE` if `amount = null`). If `specified_id` is provided, the canister is created under this id. Note that canister creation using `create_canister` or `provisional_create_canister_with_cycles` with `specified_id = null` can fail after calling `provisional_create_canister_with_cycles` with provided `specified_id`. In that case, canister creation should be retried.
@@ -2045,24 +2063,6 @@ Cycles added to this call via `ic0.call_cycles_add128` are returned to the calle
 Any user can top-up any canister this way.
 
 This method is only available in local development instances.
-
-### IC method `node_metrics` {#ic-node-metrics}
-
-:::note
-
-The node metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
-
-:::
-
-Given a subnet ID as input, this method returns a collection of metrics for all nodes on the given subnet (field `node_metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. This timestamp is the earliest timestamp after the requested timestamp (field `requested_timestamp_nanos`) for which node metrics are available.
-
-A single metric entry is a record with the following fields:
-
-- `node_id` (`principal`): the principal characterizing a node;
-
-- `num_blocks_total` (`nat64`): the number of blocks proposed by this node;
-
-- `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
 
 ## The IC Bitcoin API {#ic-bitcoin-api}
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -2128,9 +2128,9 @@ A single metric entry is of an enumeration type with a single variant (in the fu
 
 - `node_metrics`: provides metrics for a node characterized by its principal (field `node_id`) and the actual metrics values:
 
-  - `num_blocks` (`nat64`): the number of blocks proposed by this node;
+  - `num_blocks_total` (`nat64`): the number of blocks proposed by this node;
 
-  - `num_block_failures` (`nat64`): the number of failed block proposals by this node.
+  - `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
 
 ## Certification {#certification}
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -4152,7 +4152,7 @@ S with
 
 #### IC Management Canister: Metrics
 
-The management canister can returns metrics for a given subnet. The definition of the metrics values
+The management canister returns metrics for a given subnet. The definition of the metrics values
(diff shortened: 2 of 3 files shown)


Dfinity-Bjoern:
I wondered why we would not just put that information into the state tree, but I assume the reason is that we want it to be readable from a canister. We have also been debating in the past whether we should just make the state tree readable for canisters (which it isn't right now), which we could follow up on. So I guess what I was trying to say is: Wouldn't it be more generic to put this information into the subnet state tree and make the entire state tree readable for a canister?

spec/index.md
```diff
@@ -4140,6 +4150,38 @@ S with
 
 ```
 
+#### IC Management Canister: Metrics
+
+The management canister can returns metrics for a given subnet. The definition of the metrics values
```
mraszyk, line 4155:
```suggestion
The management canister returns metrics for a given subnet. The definition of the metrics values
```

mraszyk:
> Wouldn't it be more generic to put this information into the subnet state tree and make the entire state tree readable for a canister?

We discussed this in the Interface spec meeting and agreed to not pursue this now.

spec/index.md
```diff
@@ -2114,6 +2114,16 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `metrics` {#ic-metrics}
+
+Given a subnet ID as input, this method returns a collection of metrics for the given subnet.
+
+A single metric entry is of an enumeration type with a single variant (in the future, more variants might be added):
+
+- `node_metrics`: provides metrics for a node characterized by its principal (field `node_id`), for a period of time characterized by start and end timestamps in nanoseconds since 1970-01-01 (fields `start_timestamp_nanos` and `end_timestamp_nanos`), and the actual metrics values:
```
alin-at-dfinity, line 2123: Having a `start_timestamp` and `end_timestamp` attached to a bunch of gauges is not ideal.

On the one hand, it limits the resolution of the metrics. Although in this case that limitation comes mostly from the checkpointing interval or query epoch. But if we were to change the implementation of either of those in the future and be able to update said metric more often (e.g. every round), we'd either have to retain the artificially reduced resolution or have `start_timestamp` and `end_timestamp` one second apart.

Which leads to the second issue: if anyone only cares about the cumulative amount over a longer period (e.g. how many queries were executed by the subnet over the past day) they will be forced to query the metrics at least every checkpoint interval (or possibly more often) and deduplicate and maybe even interpolate to get a reasonable value.

A third issue is that as you add more and more metrics (and they are collected with different resolutions: e.g. number of replicated calls is updated every round; node metrics every checkpoint interval; query metrics every "query epoch") you'll end up with different metrics covering different and even non-overlapping time ranges (e.g. node metrics would be from 10 minutes ago; query metrics from over half an hour ago).

The way Prometheus works around this (and we also have such subnet metrics ourselves) is to use a monotonically increasing accumulator (i.e. a [counter](https://prometheus.io/docs/tutorials/understanding_metric_types/)). We could label it as e.g. "total queries executed by subnet", but the absolute value is usually not that relevant (because it depends on when the subnet was created or when the metric started being recorded). OTOH, if someone collects at least a couple of samples of this metric at different times, they can trivially compute the difference in observations between the first and the second "scrape". One can scrape the metrics endpoint as often or as rarely as they want and they will still get meaningful data out of it.

I.e. I would strongly suggest using a counter instead of gauges over arbitrary and possibly disjoint time ranges.

derlerd-dfinity: I think that a simple counter would not be enough based on how I understand the requirements. The goal of this feature is to be able to have a way to query the entire metrics for (at least) the previous month with a resolution of one day. To achieve this with a single counter would not work.

Maybe the requirements can somehow be adapted to make a single counter sufficient? A single counter would arguably be much simpler in many respects.

spec/_attachments/ic.did
```diff
@@ -184,6 +192,14 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // metrics interface
+  metrics : (record {
+    subnet_id : principal;
+  }) -> (record {
+    timestamp_nanos : nat64;
```
mraszyk, line 199: I've now drafted the API to return a single timestamp in the response that should characterize all returned metrics across all nodes. I believe this makes sense given that the metrics are gonna be fetched from the replicated state and the replicated state has a unique time associated with it (see function `ReplicatedState::time`).

alin-at-dfinity: It's not clear from the type or the description whether the API will return one `metrics` entry for every subnet node, multiple entries per subnet node; or even multiple entries for a single node. I'm guessing it's the former (even though that conflicts with what David was saying in the other comment thread), but it's never spelled out.

yvonneanne: @mraszyk, I agree that one timestamp is sufficient for this. 

@alin-at-dfinity: we haven't fully converged yet on what should be returned (we discussed several options and their pros and cons on Friday with Martin, David Derler, Dimitris and Sasa).
Once we know, we should definitely make it explicit.

For simpler auditability, multiple entries per subnet subnet node would be easiest, I'll try to expand on that in the other open thread.

spec/index.md
yvonneanne: Thanks for your comments!
I agree that a counter with a timestamp of its last change would be much simpler.
 In the future (after IC-1591), I'd envision automated NP remuneration based on such a timestamped counter, executed by a NNS-controlled canister. In the short term however, (i.e., the [feature IC-1591 we're working on right now](https://docs.google.com/presentation/d/1KXIZtYtn658L2mDKrgmL502jeBo1QbHoJRYYAg7Pu68/edit#slide=id.p)), we want to allow users to have access to historical data on node performance, ideally without building additional canister logic.
 
In a discussion last week (with Sasa, Martin, David and Dimitris), we discussed different options (listed in the slide set linked above), with and without storing historical data and making it available via this interface. 
 
 
Option 1: 

* store num_blocks and num_block_failure counters per node in subnet, 
* management canister metrics method returns list of all nodes’ current counter values together with timestamp
* build a canister storing in historical values by making  periodic calls to metrics method, stores them and makes them available via canisters methods 

+ Very simple implementation on the replica side
- Development and maintenance of new canister logic: currently there is no team responsible for such a canister, knowhow and ownership would need to be built up
- Auditability requires analyzing the canister that stores the data and its changes


Option 2:
* store num_blocks and num_block_failure counters per day (using data from first block of new calendar day) per node in subnet for the last 60 days 
* management canister metrics method returns list of all nodes’ counter values for a given day

+ Very simple auditability for NPs and IC community
+ No new canisters to be developed and maintained
- Slightly more complexity to prepare stored data and return the metrics for the desired timestamp
- More data needs to be stored (could be reduced to keeping 2 old counter values + timestamps, if 60 is too much)

In order to allow NPs and the IC community to see how the nodes perform soon and more easily, I currently favour Option 2 (less DFINITY effort, easier for users). Note that in any case this API change will be marked experimental and is expected to change.

alin-at-dfinity: Agreed. The start timestamp plus end timestamp made it look like these were gauges ("this many blocks between start timestamp and end timestamp"), which may or may not have been the intent.

Retaining the last N values of a counter at specific timestamps (at intervals of 1 day or not), while not ideal, is certainly a lot cleaner and opens he possibility of later going for a single sample at the "current timestamp" without even changing the API.

spec/index.md
```diff
@@ -2114,6 +2114,24 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
 
+### IC method `metrics` {#ic-metrics}
+
+:::note
+
+The metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
+:::
+
+Given a subnet ID as input, this method returns a collection of metrics for the given subnet (field `metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics are sampled.
```
yvonneanne, line 2125:
```suggestion
Given a subnet ID as input, this method returns a collection of metrics for the given subnet (field `metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were starting to be sampled.
```
see discussion in comments on design slides https://docs.google.com/presentation/d/1KXIZtYtn658L2mDKrgmL502jeBo1QbHoJRYYAg7Pu68/edit?disco=AAAA4HT7no0.

It is simpler to record the metrics and also accessing a certain days' metrics

spec/_attachments/ic.did
```diff
@@ -184,6 +192,14 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // metrics interface
+  metrics : (record {
+    subnet_id : principal;
```
yvonneanne, line 197:
```suggestion
    subnet_id : principal;
    minimal_timestamp_nanos: nat64;
```
see discussion on design slides https://docs.google.com/presentation/d/1KXIZtYtn658L2mDKrgmL502jeBo1QbHoJRYYAg7Pu68/edit?disco=AAAA4HT7no4

To ensure the size of the response is small enough to justify charging the method invocation cost only we want to return only the records with the smallest timestamp_nanos > minimal_timestamp_nanos

spec/index.md
```diff
@@ -2114,6 +2114,24 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+A single metric entry is of an enumeration type with a single variant (in the future, more variants might be added):
+
+- `node_metrics`: provides metrics for a node characterized by its principal (field `node_id`) and the actual metrics values:
+
+  - `num_blocks_total` (`nat64`): the number of blocks proposed by this node;
+
+  - `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
+
```
yvonneanne, line 2134: When calling `metrics(subnet_id, minimal_timestamp_nanos)` the response returns the smallest timestamp > 'minimal_timestamp_nanos' for which metrics are stored in replicated state, together with a `node_metrics` entry for each node of the subnet.

yvonneanne, line 2134:
```suggestion
When calling `metrics(subnet_id, minimal_timestamp_nanos)` the response returns the smallest timestamp > 'minimal_timestamp_nanos' for which metrics are stored in replicated state, together with a `node_metrics` entry for each node of the subnet.
```

spec/index.md
```diff
@@ -2114,6 +2114,24 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `metrics` {#ic-metrics}
+
+:::note
+
+The metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
```
yvonneanne, line 2122:
```suggestion

Calling this API charges the method invocation cost to the source canister. Therefore the size of the response must be kept small. 
```

spec/index.md
sasa-tomic: `>=` instead of `>`?

spec/_attachments/ic.did
```diff
@@ -184,6 +192,14 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // metrics interface
+  metrics : (record {
```
sasa-tomic, line 196: Should we make this more specific, since there may be canister (or query etc) metrics in the future?
Or do you feel there is no need for this?
```suggestion
  subnet_metrics : (record {
```

spec/_attachments/ic.did
sasa-tomic, line 199: based on the other discussion:
```suggestion
    start_timestamp_nanos : nat64;
```

spec/index.md
mraszyk: > were starting to be

this sounds confusing to me; given the comments on the slides, I believe we just sample the metrics at the first block of a calendar day, but this block's timestamp is the timestamp *at* which the metrics are sampled. How this timestamp is established (first/last block of a calendar day, ...) could be stated as a note on the implementation details, but does not seem to change the semantics of the values.

spec/_attachments/ic.did
```diff
@@ -111,6 +111,14 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type metrics = variant {
```
sasa-tomic, line 114:
```suggestion
type node_metrics = variant {
```

spec/_attachments/ic.did
```diff
@@ -184,6 +192,14 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // metrics interface
+  metrics : (record {
+    subnet_id : principal;
+  }) -> (record {
+    timestamp_nanos : nat64;
+    metrics : vec metrics;
```
sasa-tomic, line 200:
```suggestion
    subnet_metrics : vec node_metrics;
```

mraszyk: Shouldn't this be called node_metrics then?

spec/index.md
yvonneanne: Fair point, I probably misinterpreted @derlerd-dfinity's comment on the slide. It is much more clear if it denotes when the counter is sampled. I'll close this thread.

spec/_attachments/ic.did
yvonneanne: I'd leave it as is, see https://github.com/dfinity/interface-spec/pull/215/files#r1318372688

spec/index.md
yvonneanne: yes, that makes more sense

spec/index.md
```diff
@@ -4141,6 +4159,44 @@ S with
...
+S with
+    messages = Older_messages · Younger_messages ·
+      ResponseMessage {
+        origin = M.origin
+        response = Reply (candid(R))
+        refunded_cycles = M.transferred_cycles
+      }
+
+```
+
```
yvonneanne, line 4199: @mraszyk, I'm not familiar with this notation. Do we need to update this part of the PR to reflect the discussions from last week and the description on the design slides, or are we good?

spec/index.md
mraszyk: I believe this remark applies in general and should be clear from the page on cycle costs. Hence, let's close this thread.

spec/_attachments/ic.did
mraszyk: Indeed.

spec/_attachments/ic.did
mraszyk: Let's go for node_metrics everywhere. Originally, I've drafted the type `metrics` as a variant type so that other kinds of metrics can be added. But they'll likely have a different interface, e.g., you'd also specify a requested canister ID, and thus a separate method seems more appropriate (with corresponding types).

spec/_attachments/ic.did
mraszyk: Agreed with the name. Not sure if we then still need an enumeration (variant) type though.

spec/index.md
mraszyk: The actual response is left under-specified
```
R = <implementation-specific>
```
in the formal semantics and thus no changes are required.

spec/index.md
mraszyk: I'd have the following suggestion: let's add the sentence
```
This timestamp is the earliest timestamp after the requested timestamp (field `requested_timestamp_nanos`) for which the metrics are available.
```
after the existing sentence
```
Given a subnet ID as input, this method returns a collection of metrics for the given subnet (field `metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics are sampled.
```

spec/index.md
mraszyk: We ended up with a compromise between Option 1 and Option 2, so I'm closing this thread.

spec/_attachments/ic.did
yvonneanne: I like your suggestions!

spec/index.md
yvonneanne: Sounds great, thanks for the suggestion!

spec/index.md
yvonneanne, line 2125:
```suggestion
Given a subnet ID as input, this method returns a collection of metrics for the given subnet (field `metrics`) and a timestamp provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics are sampled.
This timestamp is the earliest timestamp after the requested timestamp (field `requested_timestamp_nanos`) for which the metrics are available.
```

spec/_attachments/ic.did
yvonneanne:
```suggestion
  node_metrics : (record {
```

spec/_attachments/ic.did
```diff
@@ -186,6 +192,15 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // metrics interface
+  node_metrics : (record {
+    subnet_id : principal;
+    requested_timestamp_nanos: nat64;
```
mraszyk, line 198: `start_at_timestamp_nanos` might be more descriptive now that we return all available metrics since this timestamp.

Dfinity-Bjoern: Should it maybe also `node_metrics_history` or something similar? If we return a time series, we should somehow make clear that you would receive a time series.

mraszyk: Sure, done.

spec/index.md
```diff
@@ -2095,6 +2095,24 @@ If you do not specify the `max_response_bytes` parameter, the maximum of a `2MB`
...
 
+### IC method `node_metrics_history` {#ic-node-metrics-history}
+
+:::note
+
+The node metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
+:::
+
+Given a subnet ID as input, this method returns collections of metrics for all nodes on the given subnet (field `node_metrics`) at timestamps provided in nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. The returned timestamps are all timestamps after (and including) the provided timestamp (field `start_at_timestamp_nanos`) for which node metrics are available. The maximum number of returned timestamps is 60 and no two returned timestamps belong to the same UTC day.
```
alin-at-dfinity, line 2106: This requirement that "no two returned timestamps belong to the same UTC day" makes the actual timestamps quite unwieldy.

I.e. ideally one would expect one sample per day, recorded at midnight UTC. But that would mean that we could never return the "current" metrics (because there would be a sample at "midnight today" and the spec does not allow for another sample for "now"). And I would argue that the ability to query the current stats (at arbitrary intervals) is at least as important as the ability to query historical stats.

As is, the implementation is forced to record samples with timestamps just before midnight UTC. Whether that's e.g. always one second before midnight; or (as currently implemented) the actual timestamp of the last observation before midnight (which introduces arbitrary jitter).

sasa-tomic: I hear you @alin-at-dfinity , and I agree with you - if I understood correctly your concern.
Would you mind going ahead and proposing some changes that would improve the interface? Would the change be as simple as removing this limitation `no two returned timestamps belong to the same UTC day` and thus allowing two entries for the current day: one for period since 0:00 UTC and another from now-delta?

derlerd-dfinity: Currently the implementation returns an iterator over all snapshots, including the current metrics as the last element. I guess it would be easy to change if we want it differently, we just need to decide.

I also wonder whether we should say something about counter resets: if a node doesn't report metrics for one day before a snapshot the implementation will not record its counter in this snapshot; similar to how Prometheus handles this. It seems to be important to mention on the interface level so that users end up with the right values when computing differences between several days.

yvonneanne:
Resets: Afaiu, no resets happen, the counters can only go up, or am I
missing something?

Last element: To keep the interface cleaner I'd follow Alin's suggestion to
have a separate endpoint for this to avoid any confusion and ambiguity.

spec/index.md
derlerd-dfinity: Following up on @yvonneanne's comment here to stick to one thread.

> Resets: Afaiu, no resets happen, the counters can only go up, or am I
> missing something?

Not sure I understand. If a node is gone for some time (the implementation currently uses 1 day) then it is garbage collected to bound the size of the recorded metrics. When it re-appears its counter will start from 0. So for the raw metrics in the replicated state it is not always guaranteed that they only go up. We could also offer an API function which hides this detail from the user, but we'd currently prefer not to because it keeps the replica implementation simple and it is consistent with the way Prometheus handles such things. 

Or are you saying that it will never happen that a node disappears for a day?

> Last element: To keep the interface cleaner I'd follow Alin's suggestion to
> have a separate endpoint for this to avoid any confusion and ambiguity.

No strong opinion. Once there is consensus please ping us so that we can provide an appropriate interface.

alin-at-dfinity: @yvonneanne
> Resets: Afaiu, no resets happen, the counters can only go up, or am I missing something?

Resets will happen (as currently implemented) if a node is removed from the subnet and later added back.

We probably shouldn't hold on to stats for nodes have long since been removed from the subnet. So we have 2 options:
* we drop node stats as soon as they stop being updated (i.e. the day after the node has been removed); or
* we keep node stats around for 60 days (or however long the sliding window is) and (as soon as the stats are identical across all snapshots) go back and remove the node from all snapshots.

The latter has the downside that we're changing history (the stats from one week ago may be different depending on when they were queried), so we went with the first option (which also simplifies the implementation).

However, if the consensus is to add a separate Management Canister method for current metrics (as opposed to all snapshots after some timestamp); then without changing the implementation we could artificially populate removed nodes' stats  in the response to `node_metrics_history`.Or we may return the raw stats, something that may result in counter resets. Or at least in some node's stats being present in the first snapshot, but not in the last.

Regardless of which approach we choose, it may be worth specifying it explicitly.

derlerd-dfinity: > Regardless of which approach we choose, it may be worth specifying it explicitly.

+1

yvonneanne: > Following up on @yvonneanne's comment here to stick to one thread.
Thanks, David (I used the email reply button, clearly this was a mistake)

> Not sure I understand. If a node is gone for some time (the implementation currently uses 1 day) then it is garbage collected to bound the size of the recorded metrics. When it re-appears its counter will start from 0. So for the raw metrics in the replicated state it is not always guaranteed that they only go up. We could also offer an API function which hides this detail from the user, but we'd currently prefer not to because it keeps the replica implementation simple and it is consistent with the way Prometheus handles such things.
> 
> Or are you saying that it will never happen that a node disappears for a day?

haha, I guess it's me who doesn't understand how things work :) 
let me ask some questions.

Is it correct that we should have a record for a node for sure if the node is rank 0 at least once per interval, because in this case either `num_blocks_total` or `num_block_failures_total` will be increased in the replicated state based on information passed by consensus? 
This should happen regardless of whether the node in question is actively participating in the system or not.

If the node belongs to the subnet according to the registry versions being used in the subnet for more than 1h, then probability to never have been blockmaker at least one is < (1-1/13)^3600 = 7.184846e-126 in an app subnet and 1.6155048e-20 for the NNS (if I try to do it for 23h hours google tells me it's 0 :)).

If a node belongs to the subnet according to the registry, but is offline for a day, then `num_block_failures_total` will be increasing with very high probability.

> 
> > Last element: To keep the interface cleaner I'd follow Alin's suggestion to
> > have a separate endpoint for this to avoid any confusion and ambiguity.
> 
> No strong opinion. Once there is consensus please ping us so that we can provide an appropriate interface.

derlerd-dfinity: Yes, I think this is correct. The case described above would only happen if a node is removed and then added back after some time. Admittedly also an unlikely case but we should still handle it properly.

yvonneanne: Thanks for your explanations, this helped a lot!

In order to do proper accounting, I expect information about all nodes that have been in the subnet for some time in the last 60 days is necessary, isn't it @sasa-tomic?

Therefore I would suggest the following:
1. separate endpoint for current metrics (to keep the interface clean and avoid confusion)
2. report node stats for all nodes that have been part of the subnet in the past 60 days

imho doing resets when node rejoin the subnet makes sense, but I think it's more important that @sasa and his team are able to consume the data in a suitable form, so I defer this question to him :)

derlerd-dfinity: Makes sense. 

> report node stats for all nodes that have been part of the subnet in the past 60 days

Not sure I understand this. We'd do this in any option. Could you clarify?

yvonneanne: Thanks for the clarification! For me this would be good enough (but clearly needs to be described in the spec), but let's see what @sasa-tomic says.

sasa-tomic: From my view, if we do as @yvonneanne suggested:
1. separate endpoint for current metrics (to keep the interface clean and avoid confusion)
2. report node stats for all nodes that have been part of the subnet in the past 60 days

we should be good to go. We only need to provide consistency within a 30 day window, where this window is somewhere in the last 60 days, in order to distribute rewards accordingly, and allow the community to verify data, again in this window. After the window is gone, it wouldn't be my biggest problem anymore.

The open question for me are the gaps with missing nodes in the stats within the 60 day window, as @derlerd-dfinity wrote above. I'll sync with him directly to see if there is something easy we can do about this.

derlerd-dfinity: Had a discussion with @sasa-tomic. We decided that it is the most pragmatic thing to expose the metrics including counter resets. Mainly because the planned client will treat nodes which are missing for an entire day as misbehaving anyways. So one can filter out those first; what remains are only metrics which don't have counter resets. 

Later, if there is demand, we can always add a second API which provides the metrics adjusted for counter resets. 


So for the spec this concretely means the following changes:

- Introduce a separate endpoint for the current metrics.
- Add a note that users of these API have to account for counter resets in the historic metrics, i.e., that if a node doesn't report metrics for a day it will not appear in the metrics for this day and its counters will start from 0 when it re-appears after that.

@sasa-tomic please correct/extend the above in case I misunderstood/missed anything.

sasa-tomic: Yes, that sounds good to me, thanks for writing down the summary, @derlerd-dfinity. Also, as we discussed in person, we don't need to have a definite and final version from the first go. We can make a couple of iterations to come up with a design acceptable to all stakeholders. The key intention here is to 1) appropriately reward well behaving nodes, and 2) penalize misbehaving nodes, both without adding a lot of complexity and strain on our side. So let's go with the cleanest design for now, and we can later discuss where we need to make some changes.

yvonneanne: Thanks for everyone's comments and especially the summary. What do you think of the following additions to ic.did and this file respectively? 

```
current_node_metrics: (subnet_id : principal;) -> (record {
    timestamp_nanos : nat64;
    node_metrics : vec node_metrics;
  });
```

```
... the number of failed block proposals by this node.

Note that the counters for both the number of blocks proposed and failed block proposals for a node are reset to zero whenever a node is removed and re-added to a subnet's membership.
```

yvonneanne: Actually, do we even need the `current_node_metrics()` endpoint at this time? who will use it in the near future?

sasa-tomic: I don't see the immediate need for `current_node_metrics()` but maybe someone else does?

derlerd-dfinity: I'm also fine with not providing `current_node_metrics` this at all. Regarding the note, I'd state this a bit more abstractly. Maybe something along these lines:

```suggestion
Given a subnet ID as input, this method returns a time series of node metrics (field `node_metrics`). The timestamps are represented as nanoseconds since 1970-01-01 (field `timestamp_nanos`) at which the metrics were sampled. The returned timestamps are all timestamps after (and including) the provided timestamp (field `start_at_timestamp_nanos`) for which node metrics are available. The maximum number of returned timestamps is 60 and no two returned timestamps belong to the same UTC day.

Note that a sample will only include metrics for nodes whose metrics changed compared to the previous sample. This means that if a node disappears in one sample and later reappears its metrics will restart from 0 and consumers of this API need to adjust for these resets when aggregating over multiple samples.
```

spec/index.md
```diff
@@ -4468,6 +4488,44 @@ S with
...
 
+#### IC Management Canister: Node Metrics
+
+:::note
+
+The node metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
+
+:::
+
+The management canister returns metrics for all nodes on a given subnet. The definition of the metrics values
```
mraszyk, line 4499:
```suggestion
The management canister returns metrics for nodes on a given subnet. The definition of the metrics values
```

PR #217, opened 2023-08-10
feat: add query endpoints to the Bitcoin API

ielashi:
There are a number of applications, particularly frontends, that require some signal as to whether or not some event on the Bitcoin chain has happened. For example, a frontend can allow you to make a Bitcoin transaction, and it wants to auto-refresh and display a signal once the transaction has been mined.

Currently, frontends cannot query the Bitcoin API because it requires cycles. They'd either need to proxy their requests via their own canister, or rely on a block explorer. The former is expensive and slow, while the latter is unideal given that the IC provides its own Bitcoin API.

This commit adds query endpoints to the Bitcoin API. They are query methods enforcing that they can only be called in non-replicated mode. The query endpoints allow frontends to query information about the Bitcoin state. These queries are fast and free for callers, knowing that they cannot fully trust the information being returned and should not be used for anything sensitive.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -2084,7 +2084,18 @@ The recommended workflow is to issue a request with the desired number of confir
 ### IC method `bitcoin_get_utxos_query` {#ic-bitcoin_get_utxos_query}
 
 This method is identical to [`bitcoin_get_utxos`](#ic-bitcoin_get_utxos), but exposed as a query.
-Note that the response of a query comes from a single replica, and is therefore not appropriate for security-sensitive applications.
+
+:::note
+
+This query is only accessible in non-replicated mode. Calls in replicated mode are rejected.
+
+:::
+
+:::warning
+
+The response of a query comes from a single replica, and is therefore not appropriate for security-sensitive applications.
+
+:::
 
 ### IC method `bitcoin_get_balance` {#ic-bitcoin_get_balance}
 
@@ -2099,7 +2110,18 @@ Given an address and the optional `min_confirmations` parameter, `bitcoin_get_ba
 ### IC method `bitcoin_get_balance_query` {#ic-bitcoin_get_balance_query}
 
 This method is identical to [`bitcoin_get_balance`](#ic-bitcoin_get_balance), but exposed as a query.
-Note that the response of a query comes from a single replica, and is therefore not appropriate for security-sensitive applications.
+
+:::note
+
+This query is only accessible in non-replicated mode. Calls in replicated mode are rejected.
+
+:::
+
+:::warning
+
+The response of a query comes from a single replica, and is therefore not appropriate for security-sensitive applications.
+
+:::
 
 ### IC method `bitcoin_send_transaction` {#ic-bitcoin_send_transaction}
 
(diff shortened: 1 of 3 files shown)


spec/index.md
```diff
@@ -2081,6 +2081,10 @@ A `get_utxos_request` without the optional `filter` results in a request that co
 
 The recommended workflow is to issue a request with the desired number of confirmations. If the `next_page` field in the response is not empty, there are more UTXOs than in the returned vector. In that case, the `page` field should be set to the `next_page` bytes in the subsequent request to obtain the next batch of UTXOs.
 
+### IC method `bitcoin_get_utxos_untrusted` {#ic-bitcoin_get_utxos_untrusted}
+
+This method is identical to [`bitcoin_get_utxos`](#ic-bitcoin_get_utxos). The only difference is that it is exposed as an uncertified query call, and its results can therefore not be trusted.
```
manudrijvers, line 2086: I would prefer to perhaps call the functions `bitcoin_get_utxos_query` or so, and tone this text down a little bit to say something like "as it's a query, the answer comes from a single replica and is therefore not appropriate for security-sensitive applications" or so.

manudrijvers (approved):
minor comment, but I am clearly in favor of adding these query methods: this is crucial for building nice front-ends for bitcoin stuff on the IC, and internally we want to use this for the NNS FE.

spec/index.md
ielashi: Done.

dragoljub-duric (approved):

Dfinity-Bjoern:
Seems fine by me, we will discuss this in next week's interface spec meeting.

Dfinity-Bjoern (approved):

ielashi:
@manudrijvers @Dfinity-Bjoern As we were implementing this, the consensus between the engineers in [the PR](https://github.com/dfinity/bitcoin-canister/pull/250) is that it may be better to avoid marking these query as composite and instead make them regular queries. The main arguments presented were:

* Composite queries incur a performance penalty as they need to keep track of the state delta (not done by regular queries).
* Composite queries may support replicated execution in the future.

In terms of the implementation, we can enforce these endpoints to only be accessible in replicated-mode with [this trick](https://dfinity.slack.com/archives/C01D7R95YJE/p1691418227396519?thread_ts=1691417215.202529&cid=C01D7R95YJE).

WDYT? If you agree, then we can update the specification accordingly.

mraszyk:
> WDYT? If you agree, then we can update the specification accordingly.

It makes sense to me. If we go for marking the methods as queries, I'd also point out that these query methods will reject replicated calls (to avoid confusion why we have both an update and a query method for the same functionality).

dragoljub-duric:
@Dfinity-Bjoern @mraszyk We have implementation merged on the Bitcoin canister, and we plan to publish the release next week or the week after it. Should this PR be merged now, or wait for the release?

mraszyk:
> @Dfinity-Bjoern @mraszyk We have implementation merged on the Bitcoin canister, and we plan to publish the release next week or the week after it. Should this PR be merged now, or wait for the release?

If the code has been merged according to the current spec PR and is expected to be released soon, then we can merge this PR.

ielashi:
Ack, merged!

PR #219, opened 2023-08-14
(IC-1243) Install large wasm 

bogwar:
This PR introduces the API changes needed to support installing large Wasms


spec/index.md
```diff
@@ -1806,6 +1806,17 @@ Not including a setting in the `settings` record means not changing that field.
 
 The optional `sender_canister_version` parameter can contain the caller's canister version. If provided, its value must be equal to `ic0.canister_version`.
 
+
+### IC method `upload_chunk` {#ic-upload_chunk}
```
mraszyk, line 1810: We should limit the maximum number of chunks that can be stored (it shouldn't be necessary to have more than 5 chunks stored at any time) as otherwise we could make the management canister API vulnerable from the security perspective.

adambratschikaye: Agree, but I think the limit will be more than 5. We currently allow Wasms to be 30MB uncompressed, so if we want to allow them to installed using chunked uploading without being compressed and choose 512KB for the chunk size it'll be more like ~100 chunks as the minimum.

But user's might also want to store the code for multiple canisters in a single canister's Wasm store, so I think the limit should be closer to 1GB.

bogwar: We'll start with some conservative choice, use 50MB for chunked storage and the slowly increase it.

spec/index.md
```diff
@@ -1830,15 +1834,14 @@ The optional `sender_canister_version` parameter can contain the caller's canist
 
 
 ### IC method `upload_chunk` {#ic-upload_chunk}
-Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister to upload such chunks. The method returns the hash of the chunk that was stored. 
+Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister to upload such chunks. The method returns the hash of the chunk that was stored. The size of each chunk should be up to 1MB. The size of the chunk store is bounded: currently it can hold up to 40chunks.
```
mraszyk, line 1837:
```suggestion
Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister to upload such chunks. The method returns the hash of the chunk that was stored. The size of each chunk should be up to 1MB. The size of the chunk store is bounded: currently it can hold up to 40 chunks. 
```

spec/index.md
```diff
@@ -1867,6 +1870,9 @@ The `wasm_module` field specifies the canister module to be installed. The syste
 
 -   If the `wasm_module` starts with byte sequence `[0x1f, 0x8b, 0x08]`, the system parses `wasm_module` as a gzip-compressed WebAssembly binary.
 
+-   If the `wasm_module` starts with the byte sequence `0xd9 0xd9 0xf7 `, the system parses `wasm_module` as a self describing CBOR encoding of a map that specifies: an optional canister identifier `storage_canister` and a list of hash values, `[h0,h1,...,hk]`. In this case, the system looks up in the chunk store of `storage_canister` (or that of the target canister if this parameter is not provided) blobs corresponding to `h1,...,hk`, concatenates them to obtain a blob of bytes and checks that `h0` is the hash of the resulting blob. The caller must be the controller of the `storage_canister`.  If the lookup succeeds, then the system interprets the blob as a vanilla wasm module (or a gzipped one) per the rules above.
```
mraszyk, line 1873:
```suggestion
-   If the `wasm_module` starts with the byte sequence `[0xd9, 0xd9, 0xf7]`, the system parses `wasm_module` as a self describing CBOR encoding of a map that specifies: an optional canister identifier `storage_canister` and a list of hash values, `[h0,h1,...,hk]`. In this case, the system looks up in the chunk store of `storage_canister` (or that of the target canister if this parameter is not provided) blobs corresponding to `h1,...,hk`, concatenates them to obtain a blob of bytes and checks that `h0` is the hash of the resulting blob. The caller must be the controller of the `storage_canister`.  If the lookup succeeds, then the system interprets the blob as a vanilla wasm module (or a gzipped one) per the rules above.
```

mraszyk, line 1873: I suppose the storage canister must be on the same subnet.

spec/index.md
```diff
@@ -3616,6 +3626,68 @@ S with
 
 ```
 
+
+#### IC Management Canister: Upload Chunk
+
+The controller of a canister, or the canister itself can upload chunks to the chunk store of that canister.
```
mraszyk, line 3632:
```suggestion
A controller of a canister, or the canister itself can upload chunks to the chunk store of that canister.
```
Since there might be multiple controllers.

spec/index.md
bogwar: Good point; i'll add that as an explanation

adambratschikaye (approved):
Great, thanks!

spec/index.md
```diff
@@ -1828,6 +1836,16 @@ Not including a setting in the `settings` record means not changing that field.
 
 The optional `sender_canister_version` parameter can contain the caller's canister version. If provided, its value must be equal to `ic0.canister_version`.
 
+
+### IC method `upload_chunk` {#ic-upload_chunk}
+Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister to upload such chunks. The method returns the hash of the chunk that was stored. The size of each chunk should be up to 1MB. The size of the chunk store is bounded: currently it can hold up to `CHUNK_STORE_SIZE` chunks. 
+
+### IC method `delete_chunks` {#ic-delete_chunks}
+Canister controllers can delete chunks stored in the canister's chunk storage. The caller provides as input a list of hashes; the effect of the call is that the IC deletes from the canister's chunk storage the chunks that correspond to hashes in the provided list.
```
mraszyk, line 1844:
```suggestion
Canister controllers can delete chunks stored in the canister's chunk storage. The caller provides as input a list of hashes; the effect of the call is that the IC deletes from the canister's chunk storage the chunks whose hashes are contained in the provided list.   
```

spec/index.md
```diff
@@ -1828,6 +1836,16 @@ Not including a setting in the `settings` record means not changing that field.
 
 The optional `sender_canister_version` parameter can contain the caller's canister version. If provided, its value must be equal to `ic0.canister_version`.
 
+
+### IC method `upload_chunk` {#ic-upload_chunk}
+Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister to upload such chunks. The method returns the hash of the chunk that was stored. The size of each chunk should be up to 1MB. The size of the chunk store is bounded: currently it can hold up to `CHUNK_STORE_SIZE` chunks.
```
mraszyk, line 1841:
```suggestion
Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister to upload such chunks. The method returns the hash of the chunk that was stored. The size of each chunk should be up to 1MiB. The size of the chunk store is bounded: currently it can hold up to `CHUNK_STORE_SIZE` chunks. 
```
I suppose the limit is half of a maximum message size which is 1MiB.

mraszyk, line 1841:
```suggestion
Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister to upload such chunks. The method returns the hash of the chunk that was stored. The size of each chunk must be at most 1MB. The maximum number of chunks in the chunk store is `CHUNK_STORE_SIZE` chunks. 
```

mraszyk, line 1841:
```suggestion
Canisters have associated some storage space (hence forth chunk storage) where they can hold chunks of Wasm modules that are too lage to fit in a single message. This method allows the controllers of a canister (and the canister itself) to upload such chunks. The method returns the hash of the chunk that was stored. The size of each chunk should be up to 1MB. The size of the chunk store is bounded: currently it can hold up to `CHUNK_STORE_SIZE` chunks. 
```

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
+
+Conditions
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.method_name = 'upload_chunk'
+M.arg = candid(A)
+chunk_store_size = |{x | chunk_store[A.canister_id][x] not null}|
```
mraszyk, line 3646:
```suggestion
chunk_store_size = |{x | S.chunk_store[A.canister_id][x] not null}|
```

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
+
+
+```
+
+State after
+
+```html
+
+S with
+    chunk_store[A.canister_id](hash') = chunk_store[A.canister_id] if hash'≠ hash
```
mraszyk, line 3659: I'd drop this line due to the implicit assumption that everything stays the same that we don't explicitly override in the "with" clause.

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
+    chunk_store[A.canister_id](hash) = A.chunk
```
mraszyk, line 3660: Do we also require that no chunk with that hash is already in the store? If not, then we should adjust the condition on the maximum number of chunks in the store (we don't need a strict inequality in this case).

spec/index.md
mraszyk: also applied to the other methods below

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
+The controller of a canister, or the canister itself can delete chunks from the canister chunk store. 
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.method_name = 'delete_chunks'
+M.arg = candid(A)
+chunk_store'[hash] = chunk_store[A.canister_id][hash] if hash ∉ A.hash_list
+chunk_store'[hash] = null if hash ∈ A.hash_list
```
mraszyk, line 3708: I'd drop these lines and undefine `hash` in the "State after" section analogously to the canister deletion transition (for instance).

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
+M.caller ∈ S.controllers[A.canister_id] ∪ {A.canister_id}
+
+```
+
+State after
+
+```html
+
+S with
+    chunk_store[A.canister_id] = chunk_store'
```
mraszyk, line 3718:
```suggestion
    chunk_store[A.canister_id](hash) = (deleted)
```

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
```
mraszyk, line 3747:
```suggestion
```

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
+
+The controller of a canister, or the canister itself can list the hashes of the chunks stored in the chunk store.
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.method_name = 'stored_chunks'
+M.arg = candid(A)
+stored_chunks = {hash | chunk_store[A.canister_id][hash] ≠ null}
```
mraszyk, line 3737: I'd inline this expression as
```
dom(S.chunk_store[A.canister_id])
```
in the "State after" part below.

spec/index.md
```diff
@@ -1022,7 +1030,7 @@ Applications can work around these problems. For the first problem, the query re
...
 
-A canister module is a [WebAssembly module](https://webassembly.github.io/spec/core/index.html) that is either in binary format (typically `.wasm`) or gzip-compressed (typically `.wasm.gz`). If the module starts with byte sequence `[0x1f, 0x8b, 0x08]`, then the system decompresses the contents as a gzip stream according to [RFC-1952](https://datatracker.ietf.org/doc/html/rfc1952.html) and then parses the output as a WebAssembly binary.
+A canister module is a [WebAssembly module](https://webassembly.github.io/spec/core/index.html) that is either in binary format (typically `.wasm`), gzip-compressed (typically `.wasm.gz`), or specified through a list of hashes.  If the module starts with byte sequence `[0x1f, 0x8b, 0x08]`, then the system decompresses the contents as a gzip stream according to [RFC-1952](https://datatracker.ietf.org/doc/html/rfc1952.html) and then parses the output as a WebAssembly binary. If the module starts with `[0xd9, 0xd9, 0xf7]` then the system interprets the content as a list of hashes.  In this case, the actual Wasm module is obtained by concatenating the individual bitstrings that correspond to these hashes.
```
mraszyk, line 1033:
```suggestion
A canister module is a [WebAssembly module](https://webassembly.github.io/spec/core/index.html) that is either in binary format (typically `.wasm`), gzip-compressed (typically `.wasm.gz`), or specified through a list of hashes. If the module starts with byte sequence `[0x1f, 0x8b, 0x08]`, then the system decompresses the contents as a gzip stream according to [RFC-1952](https://datatracker.ietf.org/doc/html/rfc1952.html) and then parses the output as a WebAssembly binary. If the module starts with the byte sequence `[0xd9, 0xd9, 0xf7]` (self-describe CBOR tag), then the system interprets the content as a CBOR (see [CBOR](#cbor)) value according to the following [CDDL](_attachments/chunked-wasms.cddl) (see [CDDL](#cddl)).  In this case, the actual Wasm module is obtained by concatenating the individual chunks (from the chunk store of the installed canister or from the chunk store of a specified canister on the same subnet and controlled by the caller) that correspond to these hashes. 
```

spec/index.md
```diff
@@ -1231,17 +1239,17 @@ The following sections describe various System API functions, also referred to a
...
     ic0.stable64_write : (offset : i64, src : i64, size : i64) -> ();           // * s
     ic0.stable64_read : (dst : i64, offset : i64, size : i64) -> ();            // * s
-
+    
     ic0.certified_data_set : (src: i32, size: i32) -> ();                       // I G U Ry Rt T
     ic0.data_certificate_present : () -> i32;                                   // *
     ic0.data_certificate_size : () -> i32;                                      // *
     ic0.data_certificate_copy : (dst: i32, offset: i32, size: i32) -> ();       // *
-
+
```
mraszyk, line 1247: Could we please get rid of those white spaces throughout this PR?

spec/_attachments/chunked-wasms.cddl
```diff
@@ -0,0 +1,15 @@
+
+; The specification of a large Wasm; consists of a location (canister principal) and a list of hashes.
```
mraszyk, line 2:
```suggestion
; The specification of a large Wasm; consists of an optional location (canister principal) and a list of hashes.
```

spec/index.md
```diff
@@ -1856,6 +1874,9 @@ The `wasm_module` field specifies the canister module to be installed. The syste
 
 -   If the `wasm_module` starts with byte sequence `[0x1f, 0x8b, 0x08]`, the system parses `wasm_module` as a gzip-compressed WebAssembly binary.
 
+-   If the `wasm_module` starts with the byte sequence `[0xd9, 0xd9, 0xf7]`, the system parses `wasm_module` as a self describing CBOR encoding of a map that specifies: an optional canister identifier `storage_canister` and a list of hash values, `[h0,h1,...,hk]` with `k <= MAX_CHUNKS_IN_LARGE_WASM`. In this case, the system looks up in the chunk store of `storage_canister` (or that of the target canister if this parameter is not provided) blobs corresponding to `h1,...,hk`, concatenates them to obtain a blob of bytes and checks that `h0` is the hash of the resulting blob. The caller must be a controller of the `storage_canister`.  If the lookup succeeds, then the system interprets the blob as a vanilla wasm module (or a gzipped one) per the rules above.
```
mraszyk, line 1877: In the CDDL, the canister is referred to as "location".

mraszyk, line 1877:
```suggestion
-   If the `wasm_module` starts with the byte sequence `[0xd9, 0xd9, 0xf7]`, the system parses `wasm_module` as a self describing CBOR encoding of a map that specifies: an optional canister identifier `storage_canister` and a list of hash values, `[h0,h1,...,hk]` with `k <= MAX_CHUNKS_IN_LARGE_WASM`. In this case, the system looks up in the chunk store of `storage_canister` (or that of the target canister if this parameter is not provided) blobs corresponding to `h1,...,hk`, concatenates them to obtain a blob of bytes and checks that `h0` is the hash of the resulting blob. The caller must be a controller of the `storage_canister` or the caller must be the `storage_canister`.  If the lookup succeeds, then the system interprets the blob as a vanilla wasm module (or a gzipped one) per the rules above.
```
I think this also makes sense.

spec/_attachments/chunked-wasms.cddl
mraszyk: referred to as `storage_canister` in the main text (please sync)

spec/index.md
bogwar: done

spec/index.md
bogwar: Good point

spec/index.md
```diff
@@ -3605,6 +3630,132 @@ S with
...
+Conditions
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.method_name = 'upload_chunk'
+M.arg = candid(A)
+chunk_store_size = |{x | chunk_store[A.canister_id][x] not null}|
+chunk_store_size < CHUNK_STORE_SIZE and (M.caller ∈ S.controllers[A.canister_id] ∪ {A.canister_id})
```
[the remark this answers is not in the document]
bogwar: done

spec/index.md
bogwar: done

spec/index.md
```diff
@@ -3605,6 +3630,125 @@ S with
...
+
+Conditions
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.method_name = 'upload_chunk'
+M.arg = candid(A)
+|dom(S.chunk_store[A.canister_id]) ∪ {SHA-256(A.chunk}| <= CHUNK_STORE_SI
```
mraszyk, line 3646:
```suggestion
|dom(S.chunk_store[A.canister_id]) ∪ {SHA-256(A.chunk)}| <= CHUNK_STORE_SIZE
```

spec/index.md
mraszyk: Thank you!

spec/_attachments/ic.did
```diff
@@ -121,6 +123,13 @@ service ic : {
     settings : canister_settings;
     sender_canister_version : opt nat64;
   }) -> ();
+  upload_chunk : (record {
+    canister_id : principal;
+    chunk : blob;
+  }) -> (chunk_hash);
+  stored_chunks: (record {canister_id : canister_id}) -> (vec chunk_hash);
+  delete_chunks: (record {canister_id : canister_id; hash_list : vec chunk_hash}) -> ();
+  clear_chunk_store: (record {canister_id: canister_id}) -> ();
```
Dfinity-Bjoern, line 132: Why have this as an additional method if you can achieve the same via `delete_chunks`?

spec/index.md
```diff
@@ -1022,7 +1030,7 @@ Applications can work around these problems. For the first problem, the query re
...
+A canister module is a [WebAssembly module](https://webassembly.github.io/spec/core/index.html) that is either in binary format (typically `.wasm`), gzip-compressed (typically `.wasm.gz`), or specified through a list of hashes. If the module starts with byte sequence `[0x1f, 0x8b, 0x08]`, then the system decompresses the contents as a gzip stream according to [RFC-1952](https://datatracker.ietf.org/doc/html/rfc1952.html) and then parses the output as a WebAssembly binary. If the module starts with the byte sequence `[0xd9, 0xd9, 0xf7]` (self-describe CBOR tag), then the system interprets the content as a CBOR (see [CBOR](#cbor)) value according to the following [CDDL](_attachments/chunked-wasms.cddl) (see [CDDL](#cddl)). In this case, the actual Wasm module is obtained by concatenating the individual chunks (from the chunk store of the installed canister or from the chunk store of a specified storage canister) that correspond to these hashes (skipping the first hash which is the hash of the overall concatenated Wasm module).
```
Dfinity-Bjoern, line 1033: What is the rationale for using CBOR here?

Dfinity-Bjoern:
Overall looks good, let's discuss on Tuesday!

spec/_attachments/ic.did
bogwar: clear_chunk_store is the poor man's garbage collection and this is when the storage used is reduced to 0. 
to avoid moving stuff around, delete_chunks doesn't really shrink the space occupied in storage (it only marks space as available), so canisters need to pay for the largest amount of storage used since the last clear_chunk_store. 
One could argue that the same can be achieved by deleting all chunks in memory (and we could detect that) but it it's probably easier to explain the difference in charging if there are also two different methods

mraszyk: > to avoid moving stuff around, delete_chunks doesn't really shrink the space occupied in storage (it only marks space as available), so canisters need to pay for the largest amount of storage used since the last clear_chunk_store.

I find that very confusing given that you can reactivate the deleted chunk using its hash.

spec/index.md
mraszyk: We could use Candid that should satisfy all the requirements.

spec/_attachments/ic.did
Dfinity-Bjoern: If we would continue charging for the storage anyway, then why would one delete them in the first place?

adambratschikaye: > I find that very confusing given that you can reactivate the deleted chunk using its hash.

What do you mean by "reactivate"? Once the chunk is deleted, you can no longer retrieve its contents at all. To do so you'd need to re-upload the same chunk again.

> If we would continue charging for the storage anyway, then why would one delete them in the first place?

This would be useful for cases where multiple Wasms are being stored and you want to replace one without deleting them all. E.g. if we want to keep the last three versions of our canister then we could delete the oldest one before inserting a new one and the memory usage would stay roughly constant.

> but I think "mark_chunk_as_unused" is more descriptive of the actual behavior.

I'm not so sure about that - the chunk **is** being deleted in the sense that no one can access it anymore. It's just that the underlying space isn't being freed. This is the same behavior you'd see when a canister "deletes" data from a structure in stable memory. However I do agree that we should emphasize that the only way stop paying for storage is via the `clear_chunk_store` method. Maybe that one should have a name which makes in clearer that it's returning the space to the subnet?

spec/index.md
adambratschikaye: I think one other consideration is that third-party apps may want to decode the contents to get the hash (e.g. an app that's evaluating SNS proposals to upgrade). Maybe we want to avoid limiting those apps to using languages with Candid support?

adambratschikaye: Are there any definite advantages to using Candid?

spec/_attachments/ic.did
mraszyk: If we want to keep the currently intended behavior, then I'd point out that `delete_chunks` does not reduce the canister's memory usage while `clear_chunk_store`. It'd be easier to specify this precisely once this other [PR](https://github.com/dfinity/interface-spec/pull/221) is merged in and the canister's memory usage is introduced into the specification.

crusso:
Sorry, I was just looking over this and noticed that my comments were merely pending, so I must have failed to submit my review when I made the comments weeks ago.

spec/index.md
mraszyk: We could introduce a new management canister method, e.g., `install_code_chunked`, and specify the chunks in the Candid payload to the management canister.

spec/index.md
mraszyk: Implemented as a manual commit.

spec/index.md
```diff
@@ -1929,6 +1949,18 @@ The optional `sender_canister_version` parameter can contain the caller's canist
...
 This method traps if the canister's cycle balance decreases below the canister's freezing limit after executing the method.
 
+### IC method `install_chunked_code` {#ic-install_chunked_code}
+
+This method installs code that had previously been uploaded in chunks.
+
+The `mode`, `arg`, and `sender_canister_version` parameters are as for `install_code`.
+The `target_canister` specifies the canister where the code should be installed.
+The optional `storage_canister` specifies the canister in whose chunk storage the chunks are stored (this parameter defaults to `target_canister` if not specified).
+For the call to succeed, the caller must be a controller of the `storage_canister` or the caller must be the `storage_canister` and
```
mraszyk, line 1959: The issue with ... or ... and ... is that it's not clear whether we mean
- (... or ...) and ...
- ... or (... and ...)

spec/index.md
```diff
@@ -3028,7 +3027,7 @@ Requests that have expired are dropped here.
...
+The (unspecified) function `idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size)` determines the idle resource consumption rate in cycles per day of a canister given its current compute and memory allocation, memory usage, and subnet size. The function `freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size)` determines the freezing limit in cycles of a canister given its current compute and memory allocation, freezing threshold in seconds, memory usage, chunk store usage & and subnet size. The value `freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size)` is derived from `idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size)` and `freezing_threshold` as follows:
```
mraszyk, line 3030: Doesn't the "chunk store usage" count towards the canister's memory usage?

bogwar: Good point --

Dfinity-Bjoern (approved):
This is matured now.

PR #220, opened 2023-08-16
point out that stable memory is cleared during reinstall

mraszyk:
This PR points out explicitly that stable memory is cleared during reinstall. Currently, this is only apparent when reading the formal part of the specification (Abstract transitions).


Dfinity-Bjoern (approved):
Thanks for the clarification, @mraszyk

Issue #184, opened 2023-06-27
Check for the freezing threshold on increase of compute/memory allocation


ulan:
Currently the spec underspecifies the behaviour in the case when the canister increases its memory or compute allocation such that the cycle balance of the canister drops below its freezing threshold.

This case is a potential footgun for developers because it can lead to unexpected uninstalls of the canister.

Example:
- The developer sets the freezing threshold to 1 year.
- The canister has enough cycles to last 1 year with its _current_ memory and compute allocation
- The developer increases the memory or compute allocation of the canister (intentionally or unintentionally due to a bug). If the increase is large enough, then the canister becomes frozen and its runway could be much smaller than 1 year (e.g. few hours).

The underspecification is motivated by:
```
The compute_allocation and memory_allocation settings are ignored in this abstract model of the Internet Computer, as it does not address questions of performance or scheduling.
```

However, this example demonstrates that memory and compute allocations also affect the lifetime of the canister.

ulan:
The relevant links in the spec:
* https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-management-canister-canister-creation
* https://internetcomputer.org/docs/current/references/ic-interface-spec/#ic-management-canister-changing-settings

PR #221, opened 2023-08-21
Check freezing limit more thoroughly

mraszyk:
This PR makes the following changes:

- model compute and memory allocation and canister's memory usage in the formal semantics
- model subnet size in the formal semantics
- inspect_message is not invoked on frozen canisters (similarly to query calls)
- freezing limit is checked when creating and installing canisters and changing their settings

Closes https://github.com/dfinity/interface-spec/issues/184


spec/index.md
```diff
@@ -3409,6 +3486,42 @@ if A.settings.controllers is not null:
 else:
   New_controllers = [M.caller]
 
+if New_compute_allocation > 0 or New_memory_allocation > 0:
```
mraszyk, line 3489: In principle, we could always check, but the implementation only checks in those cases.

Dfinity-Bjoern (approved):

PR #222, opened 2023-08-22
use asset canisters for PR preview deployments

mraszyk:
This PR uses assets canisters for PR preview deployments to serve as a replacement for Netlify deployments.


Dfinity-Bjoern (approved):
Fantastic @mraszyk – thank you very much for this quick migration!

PR #228, opened 2023-09-24
query call ingress_expiry is ignored in query call signatures

mraszyk:
To facilitate BN query caching of signed query call responses, this PR makes the query call signatures ignore the query call ingress_expiry which is expected to vary for every query call.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -780,7 +780,14 @@ Given a query (the `content` map from the request body) `Q`, a response `R`, and
               status: "replied",
               reply: R.reply,
               timestamp: T,
-              request_id: hash_of_map(Q)}))
+              request_id: hash_of_map({
+                request_type: "query",
+                sender: Q.sender,
+                nonce: Q.nonce,
+                canister_id: Q.canister_id,
+                method_name: Q.method_name,
+                arg: Q.arg
+              })}))
           else
             verify_signature PK Sig ("\x0Bic-response" · hash_of_map({
               status: "rejected",
@@ -788,9 +795,16 @@ Given a query (the `content` map from the request body) `Q`, a response `R`, and
               reject_message: R.reject_message,
               error_code: R.error_code,
               timestamp: T,
-              request_id: hash_of_map(Q)}))
+              request_id: hash_of_map({
+                request_type: "query",
+                sender: Q.sender,
+                nonce: Q.nonce,
+                canister_id: Q.canister_id,
+                method_name: Q.method_name,
+                arg: Q.arg
+              })}))
 
-where `RootSubnetId` is the a priori known principal of the root subnet. Moreover, all timestamps in `R.signatures`, the certificate `Cert`, and its optional delegation must be "recent enough".
+where `RootSubnetId` is the a priori known principal of the root subnet. Moreover, all timestamps in `R.signatures`, the certificate `Cert`, and its optional delegation must be "recent enough". The optional nonce is ignored in the `hash_of_map` computation if no nonce is provided in the query call request.
 
 :::note
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -780,7 +780,7 @@ Given a query (the `content` map from the request body) `Q`, a response `R`, and
               status: "replied",
               reply: R.reply,
               timestamp: T,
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -780,17 +780,31 @@ Given a query (the `content` map from the request body) `Q`, a response `R`, and
...
               timestamp: T,
-              request_id: hash_of_map(Q)}))
+              request_id: hash_of_map({
+                request_type: "query",
+                sender: Q.sender,
+                nonce: Q.nonce,
+                canister_id: Q.canister_id,
+                method_name: Q.method_name,
+                arg: Q.arg
+              })}))
```
derlerd-dfinity, line 790: Would it make sense to use some notation which makes it clear that all fields of `Q` except `ingress_expiry` are used?

mraszyk: I've given it a try. Could you please take a look?

derlerd-dfinity: Thanks a lot. Looks good to me. Maybe we could consider using the `\` (i.e., set minus operator) instead of `-` but I have no strong opinion here as the explanation makes it clear what we mean anyways.

mraszyk: Sure, I haven't seen either `\` or `-` in the current spec so changed to `\`.

spec/index.md
```diff
@@ -780,17 +780,19 @@ Given a query (the `content` map from the request body) `Q`, a response `R`, and
...
               reject_code: R.reject_code,
               reject_message: R.reject_message,
               error_code: R.error_code,
               timestamp: T,
-              request_id: hash_of_map(Q)}))
+              request_id: hash_of_map(Q \ {ingress_expiry: Q.ingress_expiry})
+            }))
 
-where `RootSubnetId` is the a priori known principal of the root subnet. Moreover, all timestamps in `R.signatures`, the certificate `Cert`, and its optional delegation must be "recent enough".
+where `RootSubnetId` is the a priori known principal of the root subnet and `hash_of_map(Q - {ingress_expiry: Q.ingress_expiry})` denotes the [representation-independent hash](#hash-of-map) of the query's content with its `ingress_expiry` field ignored. Moreover, all timestamps in `R.signatures`, the certificate `Cert`, and its optional delegation must be "recent enough".
```
derlerd-dfinity, line 795:
```suggestion
where `RootSubnetId` is the a priori known principal of the root subnet and `hash_of_map(Q \ {ingress_expiry: Q.ingress_expiry})` denotes the [representation-independent hash](#hash-of-map) of the query's content with its `ingress_expiry` field ignored. Moreover, all timestamps in `R.signatures`, the certificate `Cert`, and its optional delegation must be "recent enough".
```

spec/index.md
```diff
@@ -780,17 +780,31 @@ Given a query (the `content` map from the request body) `Q`, a response `R`, and
               status: "replied",
               reply: R.reply,
               timestamp: T,
-              request_id: hash_of_map(Q)}))
+              request_id: hash_of_map({
```
ShuoWangNSL: We already have a section of request id in the [specification](https://internetcomputer.org/docs/current/references/ic-interface-spec/#request-id).

> When signing requests or querying the status of a request (see [Request status](https://internetcomputer.org/docs/current/references/ic-interface-spec/#state-tree-request-status)) in the state tree, the user identifies the request using a request id, which is the [representation-independent hash](https://internetcomputer.org/docs/current/references/ic-interface-spec/#hash-of-map) of the content map of the original request. A request id must have length of 32 bytes.

When the user sign the query request,  it still use the original request id(with ingress expiry). Shall we distinguish these two things in the specification to avoid confusion? For example, give this field a different name in the query response? What about `query_id`?

ShuoWangNSL: Sorry that I just find I commented on an outdated line.

It is clear that ingress_expiry is explicitly ignored. Now I am fine with keeping request_id or changing to another name.

mraszyk:
We could also make ingress_expiry deterministic by rounding it to a minute in the agents. Then BN query caching should not suffer either.

spec/index.md
ShuoWangNSL: Edit: Resolve it as the comment is outdated.

PR #229, opened 2023-09-24
add new transition rejecting stop_canister calls

mraszyk:
Given that every canister on the IC mainnet is also charged for its message memory usage which is a scarce resource, it is prudent that calls are not long-running. Hence, this PR introduces the possibility to reject `stop_canister` calls at any time.

It also updates the reject code for pending `stop_canister` calls after calling `start_canister` from `CANISTER_REJECT` (4) to `CANISTER_ERROR` (5). This makes sense because `CANISTER_REJECT` (4) is characterized as an "explicit reject by the canister".

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -4505,7 +4505,7 @@ Conditions
 
 ```html
 
-S.canister_status[CanisterId] = Stopping (Older_origins · O · Younger_origins)
+S.canister_status[CanisterId] = Stopping (Older_origins · (O, C) · Younger_origins)
 
 ```
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -2789,7 +2789,7 @@ Finally, we can describe the state of the IC as a record having the following fi
     }
     CanStatus
       = Running
-      | Stopping (List (CallOrigin, Nat))
+      | Stopping (List (CallOrigin, Nat, Nat))
       | Stopped
     ChangeOrigin
       = FromUser {
@@ -2928,7 +2928,7 @@ The following is an incomplete list of invariants that should hold for the abstr
         ∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ∈ dom(S.call_contexts)
         ∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ∈ dom(S.call_contexts)
         ∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ∈ dom(S.call_contexts)
-        ∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ∈ dom(S.call_contexts)
+        ∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _, _) ∈ Origins. O.calling_context ∈ dom(S.call_contexts)
 
 ### State transitions
 
@@ -3523,7 +3523,7 @@ S.call_contexts[Ctxt_id].needs_to_respond = true
 ∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
 ∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
 ∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ≠ Ctxt_id
-∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ≠ Ctxt_id
+∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _, _) ∈ Origins. O.calling_context ≠ Ctxt_id
 
 ```
 
@@ -3556,7 +3556,7 @@ S.call_contexts[Ctxt_id].needs_to_respond = false
 ∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
 ∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
 ∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ≠ Ctxt_id
-∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ≠ Ctxt_id
+∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _, _) ∈ Origins. O.calling_context ≠ Ctxt_id
 
 ```
 
@@ -4172,13 +4172,13 @@ The controllers of a canister can stop a canister. Stopping a canister goes thro
 
 We encode this behavior via three (types of) transitions:
 
-1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the status the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped).
+1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the IC state the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped). Every such `stop_canister` call has a timeout of 5 minutes and the corresponding `stop_canister` call is rejected if it is not responded within the timeout (the canister stays stopping in this case).
 
 2.  Next, when the canister has no open call contexts (so, in particular, all outstanding responses to the canister have been processed), the status of the canister is set to `Stopped`.
 
 3.  Finally, each pending `stop_canister` call (which are encoded in the status) is responded to, to indicate that the canister is stopped.
 
-    Conditions  
+Conditions  
 
 ```html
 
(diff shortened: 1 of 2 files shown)


Dfinity-Bjoern:
Timing out the `stop_canister` in principle seems reasonable, but I wonder about two things:
- First, we write that the target canister stays `stopping` in this case. But that seems weird: as the initiating canister, I receive a reject, but my call did actually have the effect of making the canister `stopping` and thus reject any further calls. Wouldn't it me more consistent to return to `running` state if we reject the call?
- I do not understand the switch from `CANISTER_REJECT` to `CANISTER_ERROR`. Aren't those calls rejected *by the management canister*, which means the `CANISTER_REJECT` makes sense since it is the response of the management canister?

spec/index.md
```diff
@@ -4172,13 +4172,13 @@ The controllers of a canister can stop a canister. Stopping a canister goes thro
 
 We encode this behavior via three (types of) transitions:
 
-1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the status the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped).
+1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the IC state the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped). Every such `stop_canister` call has a timeout of 5 minutes and the corresponding `stop_canister` call is rejected if it is not responded within the timeout (the canister stays stopping in this case).
```
bogwar, line 4175:
```suggestion
1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the IC state the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped). Every such `stop_canister` call has a timeout of 5 minutes and the corresponding `stop_canister` call is rejected if it is not responded to within the timeout (the canister stays stopping in this case).
```

spec/index.md
```diff
@@ -4198,12 +4198,10 @@ State after
 
 S with
     messages = Older_messages · Younger_messages
-    canister_status[A.canister_id] = Stopping [(M.origin, M.transferred_cycles)]
+    canister_status[A.canister_id] = Stopping [(M.origin, M.transferred_cycles, S.time[A.canister_id])]
```
bogwar, line 4201: Do we need to be this precise and hardwire in the spec that expiration takes 5 mins? How about leaving it underspecified & adding a non-deterministic stopping time out transition?

bogwar:
Related: if one stopping request times out, what happens to the remaining ones? The current specified behaviour is that they time out one by one whenever they timeout.

mraszyk:
> First, we write that the target canister stays stopping in this case. But that seems weird: as the initiating canister, I receive a reject, but my call did actually have the effect of making the canister stopping and thus reject any further calls. Wouldn't it me more consistent to return to running state if we reject the call?

If we keep the canister Running, then it might receive further calls making it impossible to ever stop it.

> I do not understand the switch from CANISTER_REJECT to CANISTER_ERROR. Aren't those calls rejected by the management canister, which means the CANISTER_REJECT makes sense since it is the response of the management canister?

The implementation was inconsistent in this regard:
- ingress calls to stop_canister were rejected with error code CanisterStoppingCancelled corresponding to CanisterError reject code (5),
- inter-canister calls to stop_canister were rejected with reject code CanisterReject (4),

and it made sense to me to go for CANISTER_ERROR.

spec/index.md
mraszyk: Sure we could define a constant for that, but I'd still define the value of the constant (otherwise it's unclear what you should expect as a developer).

spec/index.md
```diff
@@ -4290,6 +4284,34 @@ S with
...
+S.canister_status[CanisterId] = Stopping Origins
+
+```
+
+State after  
+
+```html
+
+S with
+    canister_status[CanisterId] = Stopping [ (O, C, T) | (O, C, T) ∈ Origins, S.time[CanisterId] <= T + 300s ]
```
oggy-dfin, line 4302: I'd get rid of the magic numbers here.

oggy-dfin:
>     * First, we write that the target canister stays `stopping` in this case. But that seems weird: as the initiating canister, I receive a reject, but my call did actually have the effect of making the canister `stopping` and thus reject any further calls. Wouldn't it me more consistent to return to `running` state if we reject the call?

I think it'd make sense to return it to running, but only if this is the only active call to `stop_canister`

Dfinity-Bjoern (approved):
We discussed that it probably makes sense to:
- keep the canister `stopping` (my worry about inconsistency was unjustified, rejected calls can change state also elsewhere)
- return a `CANISTER_REJECT` instead of a `CANISTER_ERROR`, since the operation actually changed the state.

spec/index.md
```diff
@@ -4196,13 +4196,13 @@ The controllers of a canister can stop a canister. Stopping a canister goes thro
 
 We encode this behavior via three (types of) transitions:
 
-1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the status the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped).
+1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the IC state the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped). Note that every such `stop_canister` call can be rejected by the system at any time (the canister stays stopping in this case), e.g., if the `stop_canister` call could not be responded for a long time.
```
crusso, line 4199:
```suggestion
1.  First, any `stop_canister` call sets the state of the canister to `Stopping`; we record in the IC state the origin (and cycles) of all `stop_canister` calls which arrive at the canister while it is stopping (or stopped). Note that every such `stop_canister` call can be rejected by the system at any time (the canister stays stopping in this case), e.g., if the `stop_canister` call could not be responded to for a long time.
```

dsarlis:
> We discussed that it probably makes sense to:
> 
> * keep the canister `stopping` (my worry about inconsistency was unjustified, rejected calls can change state also elsewhere)
> * return a `CANISTER_REJECT` instead of a `CANISTER_ERROR`, since the operation actually changed the state.

I am actually using a `SysTransient` in the replica draft. It felt reasonable given that the timeout is potentially a transient error (maybe if you retry the responses you were waiting on have come back by the time you retry to stop). Canister reject is used when the canister explicitly rejects a call whereas here it's more like the system decides that this request is no longer relevant after the timeout. Also compare to ingress messages timing out which is also surfaced as a `SysTransient`.

mraszyk:
> > We discussed that it probably makes sense to:
> > 
> > * keep the canister `stopping` (my worry about inconsistency was unjustified, rejected calls can change state also elsewhere)
> > * return a `CANISTER_REJECT` instead of a `CANISTER_ERROR`, since the operation actually changed the state.
> 
> I am actually using a `SysTransient` in the replica draft. It felt reasonable given that the timeout is potentially a transient error (maybe if you retry stopping the responses you were waiting on have come back and you can actually stop). Canister reject is used when the canister explicitly rejects a call whereas here it's more like the system decides that this request is no longer relevant after the timeout. Compare to ingress messages timing out which is also surfaced as a `SysTransient`.

If an ingress message times out, then it has no affect on the canister state whereas here the canister status will still be changed from running to stopping. So it feels more natural to return CANISTER_REJECT.

dsarlis:
> > > We discussed that it probably makes sense to:
> > > 
> > > * keep the canister `stopping` (my worry about inconsistency was unjustified, rejected calls can change state also elsewhere)
> > > * return a `CANISTER_REJECT` instead of a `CANISTER_ERROR`, since the operation actually changed the state.
> > 
> > 
> > I am actually using a `SysTransient` in the replica draft. It felt reasonable given that the timeout is potentially a transient error (maybe if you retry stopping the responses you were waiting on have come back and you can actually stop). Canister reject is used when the canister explicitly rejects a call whereas here it's more like the system decides that this request is no longer relevant after the timeout. Compare to ingress messages timing out which is also surfaced as a `SysTransient`.
> 
> If an ingress message times out, then it has no affect on the canister state whereas here the canister status will still be changed from running to stopping. So it feels more natural to return CANISTER_REJECT.

Wait, maybe I did not understand. The canister changes from running to stopping when a stop request comes in. Then if the stop request times out, you propose that this is a canister reject because it changes its status? But I thought that we keep it in stopping in that case...

mraszyk:
> The canister changes from running to stopping when a stop request comes in. Then if the stop request times out, you propose that this is a canister reject because it changes its status? But I thought that we keep it in stopping in that case...

Yes, we keep it stopping and that's the change: it was running when the sender made the call and when the call responses (with a reject), then the status has changed (to stopping).

mraszyk:
> return a CANISTER_REJECT instead of a CANISTER_ERROR, since the operation actually changed the state.

We also return `CANISTER_ERROR` in case a call context is starved (has no more outstanding calls and did not produce a reply) and then there could be changes to the canister's state, too. So it actually makes sense to return `CANISTER_ERROR` here, too. Then the management canister would only return `CANISTER_REJECT` for calls to canisters that have been uninstalled while the call was pending and we wouldn't introduce a new condition under which such a reject code is produced.

mraszyk:
I'll bring up the question reject codes in the next Interface spec meeting. I believe it should be easy to update them in your MR if necessary.

dsarlis:
@mraszyk Thank you. Yes, it is easy to change if needed -- I mostly want to ensure we use the most intuitive option.

spec/index.md
```diff
@@ -4517,6 +4511,31 @@ S with
...
+State after
+
+```html
+
+S with
+    canister_status[CanisterId] = Stopping (Older_origins · Younger_origins)
+    messages = S.Messages ·
+      ResponseMessage {
+        origin = O
+        response = Reject (SYS_TRANSIENT, 'Canister could not be stopped yet')
```
ielashi, line 4533: In order to match the implementation:

```suggestion
        response = Reject (SYS_TRANSIENT, 'Stop canister request timed out')
```

spec/index.md
```diff
@@ -4505,6 +4499,31 @@ S with
 
 ```
 
+Pending `stop_canister` calls may be rejected by the system at any time (the canister stays stopping in this case):
+
+Conditions
+
+```html
+
+S.canister_status[CanisterId] = Stopping (Older_origins · O · Younger_origins)
```
mraszyk, line 4508:
```suggestion
S.canister_status[CanisterId] = Stopping (Older_origins · (O, C) · Younger_origins)
```

PR #230, opened 2023-09-27
feat: Add a system api call to burn cycles

dragoljub-duric:
This PR adds a new system API call to burn cycles. This is useful in the case of the Swiss subnet where we want to create the cycles burning canister on the system subnet which should explicitly burn cycles each day using GlobalTimer, and which will represent prepayment for usage of the whole subnet.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1262,6 +1262,7 @@ The following sections describe various System API functions, also referred to a
     ic0.msg_cycles_accept : (max_amount : i64) -> (amount : i64);               // U Rt Ry
     ic0.msg_cycles_accept128 : (max_amount_high : i64, max_amount_low: i64, dst : i32)
                            -> ();                                               // U Rt Ry
+    ic0.burn_cycles : (amount_to_burn : i64) -> (amount : i64);               // U T Rt Ry
 
     ic0.canister_self_size : () -> i32;                                         // *
     ic0.canister_self_copy : (dst : i32, offset : i32, size : i32) -> ();       // *
@@ -1614,6 +1615,20 @@ Example: To accept all cycles provided in a call, invoke `ic0.msg_cycles_accept(
     This call also copies the amount of cycles that were actually moved starting at the location `dst` in the canister memory.
 
     This does not trap.
+
+-   `ic0.burn_cycles : (amount_to_burn : i64) → (amount : i64)`
+
+    This burns cycles from the camoster. It burns as many cycles as possible, up to these constraints:
+
+    It burns no more cycles than `amount_to_burn`.
+
+    It moves no more cycles than available according to `balance` - `freezing_threshold`, and
+
+    It can be called multiple times, each time possibly burning more cycles from the balance.
+
+    The return value indicates how many cycles were actually burned.
+
+    This system call does not trap.
 
 -   `ic0.call_cycles_add : (amount : i64) → ()`
 
@@ -5672,6 +5687,12 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       es.cycles_accepted := es.cycles_accepted + amount
       es.balance := es.balance + amount
       copy_cycles_to_canister<es>(dst, amount.to_little_endian_bytes())
+
+    ic0.burn_cycles<es>(amount_to_burn : i64) : i64 =
+      if es.context ∉ {U, T, Rt, Ry} then Trap {amount = 0;}
+      let amount = min(amount_to_burn, es.balance - es.params.sysenv.freezing_threshold)
+      es.balance := es.balance - amount
+      return amount
 
     ic0.canister_self_size<es>() : i32 =
       if es.context = s then Trap {cycles_used = es.cycles_used;}


spec/index.md
```diff
@@ -1615,6 +1617,20 @@ Example: To accept all cycles provided in a call, invoke `ic0.msg_cycles_accept(
 
     This does not trap.
 
+-   `ic0.burn_cycles : (amount_to_burn : i64) → (amount : i64)`
+
+    This burns cycles from the canister. It burns as many cycles as possible, up to these constraints:
+
+    It burns no more cycles than `amount_to_burn`.
+
+    It moves no more cycles than available according to `balance` - `freezing_threshold`, and
```
mraszyk, line 1626:
```suggestion
    It burns no more cycles than available according to `balance` - `freezing_threshold`.
```

mraszyk, line 1626: `freezing_threshold` is a value in time units how long the canister should remain installed: we can't deduct time duration from cycles; instead, we need to compute the so-called freezing limit in cycles. I'd reformulate the condition to saying that the canister doesn't become frozen after this system API call.

spec/index.md
```diff
@@ -1263,6 +1263,8 @@ The following sections describe various System API functions, also referred to a
     ic0.msg_cycles_accept128 : (max_amount_high : i64, max_amount_low: i64, dst : i32)
                            -> ();                                               // U Rt Ry
 
+    ic0.burn_cycles : (amount_to_burn : i64) -> (amount : i64);               // U T Rt Ry
```
mraszyk, line 1266: Do you also plan a 128-bit version?

spec/index.md
```diff
@@ -5673,6 +5689,12 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       es.balance := es.balance + amount
       copy_cycles_to_canister<es>(dst, amount.to_little_endian_bytes())
 
+    ic0.burn_cycles<es>(amount_to_burn : i64) : i64 =
+      if es.context ∉ {U, T, Rt, Ry} then Trap {amount = 0;}
+      let amount = min(amount_to_burn, es.balance - es.params.sysenv.freezing_threshold)
```
mraszyk, line 5694:
```suggestion
      let amount = min(amount_to_burn, es.balance - freezing_limit(
        es.params.sysenv.compute_allocation,
        es.params.sysenv.memory_allocation,
        es.params.sysenv.freezing_threshold,
        memory_usage_wasm_state(es.wasm_state) + es.params.sysenv.memory_usage_raw_module + es.params.sysenv.memory_usage_canister_history,
        es.params.sysenv.subnet_size,
      ))
```

spec/index.md
```diff
@@ -5673,6 +5689,12 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       es.balance := es.balance + amount
       copy_cycles_to_canister<es>(dst, amount.to_little_endian_bytes())
 
+    ic0.burn_cycles<es>(amount_to_burn : i64) : i64 =
+      if es.context ∉ {U, T, Rt, Ry} then Trap {amount = 0;}
```
mraszyk, line 5693:
```suggestion
      if es.context ∉ {U, T, Rt, Ry} then Trap {cycles_used = es.cycles_used;}
```
In case of a Trap we can't return any value.

spec/index.md
dsarlis: nit: We typically put the verb at the end of a system call name, so maybe this should be `ic0.cycles_burn` instead?

Dfinity-Bjoern: I guess we *only* want the 128-bit version, no? I think we consider the 64-bit versions as deprecated.

spec/index.md
dragoljub-duric: I am not sure that formulation is correct, as far as I understand canister may become frozen if it has `freezing_limith` cycles left. But it will not be deleted, since it will use those resources to pay for memory and other fixed costs.

spec/index.md
dragoljub-duric: My reasoning was that no one will need more the max value of 64-bit cycles to burn in one go. But it is small change so I can change it to 128 if everyone agrees.

dragoljub-duric: It is changed.

mraszyk: Maybe `cycles_burn128` to be consistent with the remaining function names in the 128-bit API.

spec/index.md
```diff
@@ -1615,6 +1617,20 @@ Example: To accept all cycles provided in a call, invoke `ic0.msg_cycles_accept(
 
     This does not trap.
 
+-   `ic0.cycles_burn : (amount_high : i64, amount_low : i64, dst : i32) -> ()`
+
+    This burns cycles from the canister. It burns as many cycles as possible, up to these constraints:
+
+    It burns no more cycles than the amount obtained by combining `amount_high` and `amount_low`. Cycles are represented by 128-bit values.
+
+    It burns no more cycles than available according to `balance` - `freezing_threshold`.
```
mraszyk, line 1626:
```suggestion
    It burns no more cycles than `balance` - `freezing_limit`, where `freezing_limit` is the amount of idle cycles burned by the canister during its `freezing_threshold`.
```

spec/index.md
mraszyk: Made a concrete edit suggestions => closing this thread.

spec/index.md
dragoljub-duric: If you feel strongly about that I can change it, but I prefer it this way because we only have one `cycles_burn` API.

mraszyk: I don't have a strong opinion. How about @Dfinity-Bjoern and @dsarlis ?

dsarlis: I can see the arguments of both sides -- consistency with other apis but also in this case we have only the 128-bit api.

Given Björn's comment that we can consider the 64-bit versions deprecated and that one day we might only have 128-bit apis but I don't expect we'll ever go back and change their names, we might choose `cycles_burn128` for the sake of consistency.

But I'm also not feeling strongly about this, I think it's ultimately up to you guys as the spec owners.

spec/index.md
```diff
@@ -1615,6 +1617,20 @@ Example: To accept all cycles provided in a call, invoke `ic0.msg_cycles_accept(
 
     This does not trap.
 
+-   `ic0.cycles_burn : (amount_high : i64, amount_low : i64, dst : i32) -> ()`
+
+    This burns cycles from the canister. It burns as many cycles as possible, up to these constraints:
+
+    It burns no more cycles than the amount obtained by combining `amount_high` and `amount_low`. Cycles are represented by 128-bit values.
```
oggy-dfin, line 1624: What does "combining" mean here? When I saw the names of the parameters I thought these amounts were upper and lower bounds, but here it says "no more cycles than", are we adding them up?

spec/index.md
Dfinity-Bjoern: I would suggest consistency and use `cycles_burn128`.

spec/index.md
```diff
@@ -1263,6 +1263,8 @@ The following sections describe various System API functions, also referred to a
     ic0.msg_cycles_accept128 : (max_amount_high : i64, max_amount_low: i64, dst : i32)
                            -> ();                                               // U Rt Ry
 
+    ic0.cycles_burn128 : (amount_high : i64, amount_low : i64, dst : i32) -> ();               // U T Rt Ry
```
ielashi, line 1266: Is there a reason why this method cannot be invoked everywhere? i.e.

```suggestion
    ic0.cycles_burn128 : (amount_high : i64, amount_low : i64, dst : i32) -> ();               // * s
```

mraszyk, line 1266:
```suggestion
    ic0.cycles_burn128 : (amount_high : i64, amount_low : i64, dst : i32) -> ();               // I G U Ry Rt C T
```

spec/index.md
```diff
@@ -5673,6 +5689,19 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       es.balance := es.balance + amount
       copy_cycles_to_canister<es>(dst, amount.to_little_endian_bytes())
 
+    ic0.cycles_burn128<es>(amount_high : i64, amount_low : i64, dst : i32) =
+      if es.context ∉ {U, T, Rt, Ry} then Trap {cycles_used = es.cycles_used;}
```
mraszyk, line 5693:
```suggestion
      if es.context ∉ {I G U Ry Rt C T} then Trap {cycles_used = es.cycles_used;}
```

dragoljub-duric:
@mraszyk  [MR](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/14987) is now merged. Can you please approve this PR?

spec/index.md
```diff
@@ -1637,6 +1639,20 @@ Example: To accept all cycles provided in a call, invoke `ic0.msg_cycles_accept(
...
+
+    It burns no more cycles than the amount obtained by combining `amount_high` and `amount_low`. Cycles are represented by 128-bit values.
+
+    It burns no more cycles than `balance` - `freezing_limit`, where `freezing_limit` is the amount of idle cycles burned by the canister during its `freezing_threshold`.
+
+    It can be called multiple times, each time possibly burning more cycles from the balance.
+
+    This call also copies the amount of cycles that were actually burned starting at the location `dst` in the canister memory.
+
+    This system call does not trap.
```
crusso, line 1654: What if dst is too close to end of memory. Does it truncate the write or trap?

We should specify that it writes to addresses [dst, dst+16), in little-endian, I guess

crusso:
Do we actually need this new functionality? It seems to be that just creating a canister (with some cycles) and then deleting that canister will have the same effect. This can all be done with existing functionality in the management canister.

Here's an example of different methods, including a best one.

https://github.com/crusso/burn

https://github.com/crusso/burn/blob/main/src/burn/main.mo#L46

I haven't tried this code in a while so it's possible it no longer works. @mraszyk, will this still work?

It might also not work in the Motoko playground due to sandboxing of canisters: the playground intercepts management canister calls and cycle transfers to prevent third parties stealing cycles from the playground. (@chenyan-dfinity can advise).

dsarlis:
> Do we actually need this new functionality? It seems to be that just creating a canister (with some cycles) and then deleting that canister will have the same effect. This can all be done with existing functionality in the management canister.
> 
> Here's an example of different methods, including a best one.
> 
> https://github.com/crusso/burn
> 
> https://github.com/crusso/burn/blob/main/src/burn/main.mo#L46
> 
> I haven't tried this code in a while so it's possible it no longer works. @mraszyk, will this still work?
> 
> It might also not work in the Motoko playground due to sandboxing of canisters: the playground intercepts management canister calls and cycle transfers to prevent third parties stealing cycles from the playground. (@chenyan-dfinity can advise).

This seems like a big hack to me. Besides the conceptual point of abusing canister creation/deletion to get cycles appear as burned, I see some other problems as well:

1. The fact that cycles are gone when you delete a canister now is something that could change in the future. IIRC, there were discussions to return the cycles to the controller (or one of the controllers) or even any principal that was designated by a controller (assuming it's a canister of course). I don't recall the details of why we did not go with this decision but it seems risky to indirectly depend on this behaviour.
2. I hardly believe we want to encourage this behaviour of creating canisters with the sole purpose of killing them immediately after. Canister ids cannot be reused when a canister is deleted so this could turn into a huge problem if people abuse creating canisters that way and we start running out of canister ids.

An api that directly allows you to burn cycles seems more appropriate and something that has a predictable behaviour in the long run. It's something we have also discussed in the past in other contexts so it's not something entirely new or out of the question.

crusso:
> > Do we actually need this new functionality? It seems to be that just creating a canister (with some cycles) and then deleting that canister will have the same effect. This can all be done with existing functionality in the management canister.
> > Here's an example of different methods, including a best one.
> > https://github.com/crusso/burn
> > https://github.com/crusso/burn/blob/main/src/burn/main.mo#L46
> > I haven't tried this code in a while so it's possible it no longer works. @mraszyk, will this still work?
> > It might also not work in the Motoko playground due to sandboxing of canisters: the playground intercepts management canister calls and cycle transfers to prevent third parties stealing cycles from the playground. (@chenyan-dfinity can advise).
> 
> This seems like a big hack to me. Besides the conceptual point of abusing canister creation/deletion to get cycles appear as burned, I see some other problems as well:
> 
> 1. The fact that cycles are gone when you delete a canister now is something that could change in the future. IIRC, there were discussions to return the cycles to the controller (or one of the controllers) or even any principal that was designated by a controller (assuming it's a canister of course). I don't recall the details of why we did not go with this decision but it seems risky to indirectly depend on this behaviour.
> 2. I hardly believe we want to encourage this behaviour of creating canisters with the sole purpose of killing them immediately after. Canister ids cannot be reused when a canister is deleted so this could turn into a huge problem if people abuse creating canisters that way and we start running out of canister ids.
> 
> An api that directly allows you to burn cycles seems more appropriate and something that has a predictable behaviour in the long run. It's something we have also discussed in the past in other contexts so it's not something entirely new or out of the question.

Ok, I'll buy those arguments.

Dfinity-Bjoern (approved):

PR #231, opened 2023-09-27
System API method in_replicated_execution

bogwar:
This PR adds a system API method to determine if the canister is running in replicated or non replicated mode.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1308,6 +1308,7 @@ The following sections describe various System API functions, also referred to a
     ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U Ry Rt C T
     ic0.performance_counter : (counter_type : i32) -> (counter : i64);          // * s
     ic0.is_controller: (src: i32, size: i32) -> ( result: i32);                 // * s
+    ic0.in_replicated_execution: () -> (result: i32);                           // * 
 
     ic0.debug_print : (src : i32, size : i32) -> ();                            // * s
     ic0.trap : (src : i32, size : i32) -> ();                                   // * s
@@ -1756,6 +1757,14 @@ In the future, we might expose more performance counters.
 The system resets the counter at the beginning of each [Entry points](#entry-points) invocation.
 
 The main purpose of this counter is to facilitate in-canister performance profiling.
+
+### Replicated execution check {#system-api-replicated-execution-check}
+
+The canister can check whether it is currently running in replicated or non replicated execution. 
+
+`ic0.in_replicated_execution : () -> (result: i32)`
+
+Returns 1 if the canister is being run in replicated mode and 0 otherwise. 
 
 ### Controller check {#system-api-controller-check}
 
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -5921,6 +5930,12 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       else
         Trap {cycles_used = es.cycles_used;}
 
+    ic0.in_replicated_execution<es>() : i32 =
+      if es.context = s then Trap {cycles_used = es.cycles_used;}
+      if es.params.sysenv.certificate = NoCertificate
```
mraszyk, line 5935: Do we want to specify the return value based on the certificate or use `es.context` instead?

spec/index.md
```diff
@@ -1308,6 +1308,7 @@ The following sections describe various System API functions, also referred to a
     ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U Ry Rt C T
     ic0.performance_counter : (counter_type : i32) -> (counter : i64);          // * s
     ic0.is_controller: (src: i32, size: i32) -> ( result: i32);                 // * s
+    ic0.in_replicated_execution: () -> (result: i32);                           // *
```
mraszyk, line 1311:
```suggestion
    ic0.is_replicated_execution: () -> (result: i32);                           // * 
```
just a suggestion

dsarlis (commented):
Generally lgtm.

One minor comment I also mentioned to Bogdan offline is whether replicated execution is an established term in the spec. I know that most of us understand what it means but I am not sure if that would be clear to someone who's reading the spec.

spec/index.md
bogwar: Yeah -- I had is_in_replicated_execution to start with. The way I picked the current proposal is the canister asking "am I in replicated execution". 
Otherwise it should be "is_execution_replicated" :). 
let's not bikeshed too much -- of these options is one that you favor more?

spec/index.md
bogwar: Good question. I thought doing it this way is future proof since we won't change how certified state is created/handled, but using es would be the clean way to do it (with the caveat that we may forget to update it if/when we introduce some new execution modes). wdyt?

mraszyk: Either way sounds good to me if you think it is future proof.

mraszyk:
> whether replicated execution is an established term in the spec

there are already a few occurrence, but all seem to be added recently

spec/index.md
mraszyk: Makes sense. Let's stick with the current name.

bogwar:
> Generally lgtm.
> 
> One minor comment I also mentioned to Bogdan offline is whether replicated execution is an established term in the spec. I know that most of us understand what it means but I am not sure if that would be clear to someone who's reading the spec.

This was a very good point; I checked and we didn't actually introduce these terms properly so I've added a small clarification in the nomenclature section.

PR #233, opened 2023-10-05
ingress messages to canisters that are not running cannot be received

mraszyk:
This MR adds a check that a canister receiving an ingress message is Running before the ingress message is marked as `Received`. This is because update calls to Stopping/Stopped canisters are rejected anyway and thus it doesn't make sense to accept such an ingress message.

This MR also cleans up the conditions on ingress messages to provisional API (controller check is dropped for the provisional API).


Dfinity-Bjoern (approved):
Thank you for fixing these issues!

PR #234, opened 2023-10-06
update status endpoint spec

mraszyk:
This MR updates the specification of the `/api/v2/status` endpoint:
- drop `ic_api_version` field that does not reflect the actual specification version implemented by the replica;
- drop `impl_source` field that is not populated by the replica implementation;
- rename `impl_revision` into `impl_version` (dropping the current `impl_version`);
- do not specify `root_key` to be only present in development instances.

The last three changes are catching up with the actual behavior of the IC mainnet.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1024,6 +1024,8 @@ Additionally, the Internet Computer provides an API endpoint to obtain various s
 
 For this endpoint, the user performs a GET request, and receives a CBOR (see [CBOR](#cbor)) value with the following fields. The IC may include additional implementation-specific fields.
 
+-   `ic_api_version` (string, mandatory): Identifies the interface version supported, i.e. the version of the present document that the Internet Computer aims to support, e.g. `0.8.1`. The implementation may also return `unversioned` to indicate that it does *not* comply to a particular version, e.g. in between releases.
+
 -   `impl_version` (string, optional): The precise git revision of the Internet Computer Protocol implementation
 
 -   `root_key` (blob, optional): The public key (a DER-encoded BLS key) of the root key of this instance of the Internet Computer Protocol. This *must* be present in short-lived development instances, to allow the agent to fetch the public key. For the Internet Computer, agents must have an independent trustworthy source for this data, and must not be tempted to fetch it from this insecure location.
--- a/spec/index.md
+++ b/spec/index.md
@@ -1024,8 +1024,6 @@ Additionally, the Internet Computer provides an API endpoint to obtain various s
 
 For this endpoint, the user performs a GET request, and receives a CBOR (see [CBOR](#cbor)) value with the following fields. The IC may include additional implementation-specific fields.
 
--   `ic_api_version` (string, mandatory): Identifies the interface version supported, i.e. the version of the present document that the Internet Computer aims to support, e.g. `0.8.1`. The implementation may also return `unversioned` to indicate that it does *not* comply to a particular version, e.g. in between releases.
-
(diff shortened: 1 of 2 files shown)


Dfinity-Bjoern:
As discussed in the previous meeting, moving the version from the status endpoint to the state tree seems correct. I wonder whether we should make the root key accessible by spec or maybe remove it in prod instance. Retrieving the root key from some (untrusted) node seems a flaw and a footgun.

mraszyk:
> I wonder whether we should make the root key accessible by spec or maybe remove it in prod instance. Retrieving the root key from some (untrusted) node seems a flaw and a footgun.

Done.

spec/index.md
```diff
@@ -1015,13 +1015,9 @@ Additionally, the Internet Computer provides an API endpoint to obtain various s
...
-
 -   `impl_source` (string, optional): Identifies the implementation of the Internet Computer Protocol, by convention with the canonical location of the source code (e.g. `https://github.com/dfinity/ic`).
 
--   `impl_version` (string, optional): If the user is talking to a released version of an Internet Computer Protocol implementation, this is the version number. For non-released versions, output of `git describe` like `0.1.13-13-g2414721` would also be very suitable.
-
--   `impl_revision` (string, optional): The precise git revision of the Internet Computer Protocol implementation
+-   `impl_version` (string, optional): The precise git revision of the Internet Computer Protocol implementation
 
 -   `root_key` (blob, only in development instances): The public key (a DER-encoded BLS key) of the root key of this development instance of the Internet Computer Protocol. This *must* be present in short-lived development instances, to allow the agent to fetch the public key. For the Internet Computer, agents must have an independent trustworthy source for this data, and must not be tempted to fetch it from this insecure location.
```
Dfinity-Bjoern, line 1022:
```suggestion
-   `root_key` (blob): The public key (a DER-encoded BLS key) of the root key of this development instance of the Internet Computer Protocol. This *must* be present in short-lived development instances, to allow the agent to fetch the public key. For the Internet Computer, agents must have an independent trustworthy source for this data, and must not be tempted to fetch it from this insecure location.
```

spec/index.md
```diff
@@ -1015,15 +1015,11 @@ Additionally, the Internet Computer provides an API endpoint to obtain various s
...
-
--   `impl_revision` (string, optional): The precise git revision of the Internet Computer Protocol implementation
+-   `impl_version` (string, optional): The precise git revision of the Internet Computer Protocol implementation
 
--   `root_key` (blob, only in development instances): The public key (a DER-encoded BLS key) of the root key of this development instance of the Internet Computer Protocol. This *must* be present in short-lived development instances, to allow the agent to fetch the public key. For the Internet Computer, agents must have an independent trustworthy source for this data, and must not be tempted to fetch it from this insecure location.
+-   `root_key` (blob): The public key (a DER-encoded BLS key) of the root key of this instance of the Internet Computer Protocol. This *must* be present in short-lived development instances, to allow the agent to fetch the public key. For the Internet Computer, agents must have an independent trustworthy source for this data, and must not be tempted to fetch it from this insecure location.
```
mraszyk, line 1022:
```suggestion
-   `root_key` (blob, mandatory): The public key (a DER-encoded BLS key) of the root key of this instance of the Internet Computer Protocol. This *must* be present in short-lived development instances, to allow the agent to fetch the public key. For the Internet Computer, agents must have an independent trustworthy source for this data, and must not be tempted to fetch it from this insecure location.
```

Dfinity-Bjoern (approved):
It really hurts, but it's probably the right thing to do for now. Ouch!

PR #235, opened 2023-10-06
store ic_api_version of each subnet in the certified state tree

mraszyk:
This MR stores the supported interface version of every subnet in the path `/subnet/<subnet_id>/ic_api_version` of the certified state tree and updates the condition on read_state requests to retrieve this path.


Dfinity-Bjoern (commented):
Thanks, Martin! This looks good to me, let's get feedback from engineering teams!

PR #237, opened 2023-10-11
Increase max number of WASM globals

mraszyk:
The current limit is not enough for C++ and Python languages as discussed on the forum: https://forum.dfinity.org/t/wasm-module-defined-globals-which-exceeds-the-maximum-number-allowed-200/


Dfinity-Bjoern (approved):
If the runtime team considers this reasonable, then let's go.

PR #238, opened 2023-10-13
add clarification for whitespace in the ic-certificateexpression header

nathanosdev:
This PR removes some whitespace in the EBNF specification of the IC-CertificateExpression header. Our current implementation is whitespace agnostic, but the specification should be clear that there is no whitespace allowed in the value of this header. I've also added a note that implementers can also _optionally_ allow for whitespace.

Some other minor changes are a result of Prettier formatting.

Thanks @frederikrothenberger for bringing this up: https://dfinity.slack.com/archives/CN4SJRNG4/p1696930887498429

This change could actually break an implementation of this spec, if there was such an implementation out there that took the EBNF literally and did not account for any whitespace flexibility. We're not currently aware of any implementations aside from our own, which is agnostic of whitespace: https://github.com/dfinity/response-verification

### DIFF
--- a/spec/http-gateway-protocol-spec.md
+++ b/spec/http-gateway-protocol-spec.md
@@ -274,14 +274,18 @@ REQUEST-CERTIFICATION = 'RequestCertification{certified_request_headers:', STRIN
 
 RESPONSE-CERTIFICATION = 'ResponseCertification{', ('response_header_exclusions:' | 'certified_response_headers:'), RESPONSE-HEADER-LIST, '}'
 
-CERTIFICATION = 'Certification{', ('no_request_certification: Empty{}' | 'request_certification:', REQUEST-CERTIFICATION), ',response_certification:', RESPONSE-CERTIFICATION, '}'
+CERTIFICATION = 'Certification{', ('no_request_certification:Empty{}' | 'request_certification:', REQUEST-CERTIFICATION), ',response_certification:', RESPONSE-CERTIFICATION, '}'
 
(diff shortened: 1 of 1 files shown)


spec/http-gateway-protocol-spec.md
```diff
@@ -274,15 +274,19 @@ REQUEST-CERTIFICATION = 'RequestCertification{certified_request_headers:', STRIN
...
 
-CERTIFICATION = 'Certification{', ('no_request_certification: Empty{}' | 'request_certification:', REQUEST-CERTIFICATION), ',response_certification:', RESPONSE-CERTIFICATION, '}'
+CERTIFICATION = 'Certification{', ('no_request_certification:Empty{}' | 'request_certification:', REQUEST-CERTIFICATION), ',response_certification:', RESPONSE-CERTIFICATION, '}'
 
-VALIDATION-ARGS = 'ValidationArgs{', ('no_certification: Empty{}' | 'certification:', CERTIFICATION), '}'
+VALIDATION-ARGS = 'ValidationArgs{', ('no_certification:Empty{}' | 'certification:', CERTIFICATION), '}'
 
 HEADER-VALUE = 'default_certification(', VALIDATION-ARGS, ')'
 
 HEADER = 'IC-CertificateExpression: ', HEADER-VALUE
```
oggy-dfin, line 283: Should this whitespace be removed too? 
```suggestion
HEADER = 'IC-CertificateExpression:', HEADER-VALUE
```

nathanosdev:
Thanks @oggy-dfin, I've applied your suggestions

Dfinity-Bjoern (approved):

nathanosdev:
Thanks @Dfinity-Bjoern

PR #239, opened 2023-10-13
Limit subnet certificate delegations to depth 1

oggy-dfin:
As discussed in various meetings, the nested certificate delegations allowed by the spec aren't used, but have complicated the reasoning about some features (see discussions around delegations in https://github.com/dfinity/interface-spec/pull/163 for example), and stating some properties (like the new subnet `read_state` requests). But we don't use nested delegations in practice and we don't have any plans to use them in the future. So this PR removes them.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -379,7 +379,7 @@ This section makes forward references to other concepts in this document, in par
 
     where `signing_canister_id` is the id of the signing canister and `reconstruct` is a function that computes a root-hash for the tree.
 
-    -   If the `certificate` includes subnet delegations (possibly nested), then the `signing_canister_id` must be included in each delegation's canister id range (see [Delegation](#certification-delegation)).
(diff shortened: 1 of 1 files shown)


krpeacock:
This speaks to a slightly broader issue with the specification, but if there are going to be different rules for a valid **subnet delegation**, as opposed to what is allowed in a valid **delegation chain** used for a client `identity` to make calls to the Internet Computer, we should break down those differences more clearly. 

In the actual agent implementation, this change makes `delegation` logic more complicated, by introducing differences in validation between responses and `DelegationChain` identities

krpeacock:
I think all I really want is a section in the specification that offers a clear disambiguation about the different "delegation" types and what they should contain. It doesn't conflict with this proposal, but came to mind while I was reading it, based on my recent project that called for a close reading of the spec

Dfinity-Bjoern (approved):

PR #240, opened 2023-10-17
feat: update representation-independent hash computation

mraszyk:
This MR updates the specification of the representation-independent hash computation for the case of maps with nested maps.


adamspofford-dfinity:
Should it not be the other way around, `H(key) . hash_of_map(...)`? It seems inconsistent that arrays are hashed once but maps are hashed twice.

mraszyk:
> Should it not be the other way around, `H(key) . hash_of_map(...)`? It seems inconsistent that arrays are hashed once but maps are hashed twice. (Plus if there's a double hash it can no longer be implemented as a single stateless serde serializer.)

Sounds good! I've updated the PR and its example computation to reflect your suggestion.

Dfinity-Bjoern (approved):
As discussed, looks good, thanks @mraszyk!

PR #241, opened 2023-10-18
fix: bound delegation depth in verify_response

mraszyk:
The current definition of `verify_response` for replica-signed queries allows for nested delegations in which case the canister ranges fetched from the delegation might not be signed by the root subnet. This PR fixes this security vulnerability by limiting the delegation depth to one.

There's a separate [PR](https://github.com/dfinity/interface-spec/pull/239) bounding the delegation depth to one globally. This PR is void if the delegation depth is bounded globally.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -802,7 +802,7 @@ Given a query (the `content` map from the request body) `Q`, a response `R`, and
     verify_response(Q, R, Cert)
       = verify_cert(Cert) ∧
         ((Cert.delegation = NoDelegation ∧ SubnetId = RootSubnetId ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert) = Found Ranges) ∨
-         (SubnetId = Cert.delegation.subnet_id ∧ lookup(["subnet",SubnetId,"canister_ranges"], Cert.delegation.certificate) = Found Ranges)) ∧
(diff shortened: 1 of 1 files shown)


oggy-dfin:
For the read_state request, in the informal part we already have this condition:
> certificate (blob): A certificate (see [Certification](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certification)).
>
 >  If this certificate includes subnet delegations (possibly nested), then the effective_canister_id must be included in each delegation's canister id range (see [Delegation](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certification-delegation)).
 
Since the formal part states that the certificate is obtained in a separate read_state request, I'd assume that the condition is basically checked there?

mraszyk:
> For the read_state request, in the informal part we already have this condition:
> 
> > certificate (blob): A certificate (see [Certification](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certification)).
> > If this certificate includes subnet delegations (possibly nested), then the effective_canister_id must be included in each delegation's canister id range (see [Delegation](https://internetcomputer.org/docs/current/references/ic-interface-spec/#certification-delegation)).
> 
> Since the formal part states that the certificate is obtained in a separate read_state request, I'd assume that the condition is basically checked there?

We need to obtain the subnet_id for subsequent certificate lookup so we need this code explicitly in `verify_response`. However, I agree that security-wise the read_state verification checks that the certificate is authorized for the given effective_canister_id and thus we don't need to bound the delegation chain to be of length one here.

PR #245, opened 2023-11-01
fix: update deprecation note on read state paths

mraszyk:
This PR reformulates a note on deprecating paths in read state requests for `/api/v2/canister/...` with prefix `/subnet`. They're no longer marked as "will be deprecated", but rather "might be deprecated" and the advise to update is softened.


Dfinity-Bjoern (approved):
Thank you for clarifying, @mraszyk!

PR #249, opened 2023-11-14
restrict nonce to be of length at most 32

mraszyk:
This MR restrict the maximum length of a nonce in a request to 32 since 32 bytes are supposed to provide enough randomness.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -858,7 +858,7 @@ In development instances of the Internet Computer Protocol (e.g. testnets), the 
 
 All requests coming in via the HTTPS interface need to be either *anonymous* or *authenticated* using a cryptographic signature. To that end, the following fields are present in the `content` map in all cases:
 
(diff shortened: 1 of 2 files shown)


spec/index.md
```diff
@@ -858,7 +858,7 @@ In development instances of the Internet Computer Protocol (e.g. testnets), the
 
 All requests coming in via the HTTPS interface need to be either *anonymous* or *authenticated* using a cryptographic signature. To that end, the following fields are present in the `content` map in all cases:
 
--   `nonce` (`blob`, optional): Arbitrary user-provided data, typically randomly generated. This can be used to create distinct requests with otherwise identical fields.
+-   `nonce` (`blob`, optional): Arbitrary user-provided data of length at most 32, typically randomly generated. This can be used to create distinct requests with otherwise identical fields.
```
Dfinity-Bjoern, line 861:
```suggestion
-   `nonce` (`blob`, optional): Arbitrary user-provided data of length at most 32 bytes, typically randomly generated. This can be used to create distinct requests with otherwise identical fields.
```

PR #250, opened 2023-11-15
feat: add root_subnet_id to HTTP status endpoint

mraszyk:
This PR adds the `root_subnet_id` as an additional field returned by the HTTP `/api/v2/status` endpoint. This is useful to obtain the root subnet ID along with the root subnet's public key in development instances. However, we propose to not restrict the field to development instances only so that the IC mainnet does not deviate from test instances.


Dfinity-Bjoern:
I am not convinced. You can easily compute that from the root public key.

PR #251, opened 2023-11-20
Only controllers can install chunked code

bogwar:
Added the restriction that only controllers can install code uploaded in chunks


spec/index.md
```diff
@@ -1977,6 +1977,8 @@ This method traps if the canister's cycle balance decreases below the canister's
 
 This method installs code that had previously been uploaded in chunks.
 
+Only controllers of the target canister can install code.
```
mraszyk, line 1980:
```suggestion
Only controllers of the target canister can call this method.
```

mraszyk (approved):
LGTM

PR #256, opened 2023-11-23
chore: always run PR cleanup job

mraszyk:
This PR removes the concurrency group for the PR cleanup job so that this job is not cancelled by the master preview deployment job.


bogwar (approved):
thanks, Martin!

PR #257, opened 2023-11-28
Specify the resource reservation mechanism

ulan:
Context:

  - https://forum.dfinity.org/t/23447
  - https://dashboard.internetcomputer.org/proposal/126094

This adds new fields to the management canister interface:

  - `reserved_cycles` to the result of `canister_status`.
  - `reserved_cycles_limit` to `canister_settings` and `definite_canister_settings`.

The corresponding fields are also added to the abstract state:

  - `reserved_balances`,
  - `reserved_balance_limits`.

Note that the naming matches the exiting naming of the main cycle balance, which is `cycles` in the management canister and `balances` in the abstract state.

A new unspecified function `cycles_to_reserve` models cycle reservation for allocated resources depending on the old abstract state and the new state of the canister (and its new allocation settings).

Since the reserved cycles also affect canister freezing, a new helper function called `liquid_balance` is used to compute how many cycles are available after taking the freezing limit and reserved cycles into account.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -227,6 +227,8 @@ Dapps on the Internet Computer are called *canisters*. Conceptually, they consis
 
 -   A cycle balance
 
+-   A reserved cycle balance, which are cycles set aside from the main cycle balance for resource payments.
+
 -   The *canister status*, which is one of `running`, `stopping` or `stopped`.
 
 -   Resource reservations
@@ -245,19 +247,19 @@ If an empty canister receives a response, that response is dropped, as if the ca
 
 #### Canister cycles {#canister-cycles}
 
-The IC relies on *cycles*, a utility token, to manage its resources. A canister pays for the resources it uses from its *cycle balance*. The *cycle\_balance* is stored as 128-bit unsigned integers and operations on them are saturating. In particular, if *cycles* are added to a canister that would bring its total balance beyond 2<sup>128</sup>-1, then the balance will be capped at 2<sup>128</sup>-1 and any additional cycles will be lost.
+The IC relies on *cycles*, a utility token, to manage its resources. A canister pays for the resources it uses from its *cycle balances*. A *cycle\_balance* is stored as 128-bit unsigned integers and operations on them are saturating. In particular, if *cycles* are added to a canister that would bring its main cycle balance beyond 2<sup>128</sup>-1, then the balance will be capped at 2<sup>128</sup>-1 and any additional cycles will be lost.
 
-When the cycle balance of a canister falls to zero, the canister is *deallocated*. This has the same effect as
+When both the main and the reserved cycle balances of a canister fall to zero, the canister is *deallocated*. This has the same effect as
 
 -   uninstalling the canister (as described in [IC method](#ic-uninstall_code))
 
 -   setting all resource reservations to zero
 
-Afterwards the canister is empty. It can be reinstalled after topping up its balance.
+Afterwards the canister is empty. It can be reinstalled after topping up its main balance.
 
 :::note
 
-Once the IC frees the resources of a canister, its id, *cycles* balance, *controllers*, canister *version*, and the total number of canister changes are preserved on the IC for a minimum of 10 years. What happens to the canister after this period is currently unspecified.
+Once the IC frees the resources of a canister, its id, *cycle* balances, *controllers*, canister *version*, and the total number of canister changes are preserved on the IC for a minimum of 10 years. What happens to the canister after this period is currently unspecified.
 
 :::
 
@@ -1660,7 +1662,7 @@ Example: To accept all cycles provided in a call, invoke `ic0.msg_cycles_accept(
 
     It burns no more cycles than the amount obtained by combining `amount_high` and `amount_low`. Cycles are represented by 128-bit values.
 
-    It burns no more cycles than `balance` - `freezing_limit`, where `freezing_limit` is the amount of idle cycles burned by the canister during its `freezing_threshold`.
+    It burns no more cycles than the amount of cycles available for spending `liquid_balance(balance, reserved_balance, freezing_limit)`, where `reserved_balance` are cycles reserved for resource payments and `freezing_limit` is the amount of idle cycles burned by the canister during its `freezing_threshold`.
 
     It can be called multiple times, each time possibly burning more cycles from the balance.
 
@@ -1946,6 +1948,14 @@ The optional `settings` parameter can be used to set the following settings:
 
     Default value: 2592000 (approximately 30 days).
 
+-   `reserved_cycles_limit` (`nat`)
+
+    Must be a number between 0 and 2<sup>128</sup>-1, inclusively, and indicates the upper limit on `reserved_cycles` of the canister.
+
+    An operation that allocates resources such as compute and memory will fail of the new value of `reserved_cycles` exceeds this limit.
+
+    Default value: 5_000_000_000_000 (5 trillion cycles).
+
 The optional `sender_canister_version` parameter can contain the caller's canister version. If provided, its value must be equal to `ic0.canister_version`.
 
 Until code is installed, the canister is `Empty` and behaves like a canister that has no public methods.
@@ -2027,7 +2037,7 @@ Uninstalling a canister's code will reject all calls that the canister has not y
 
 The canister is now [empty](#canister-lifecycle). In particular, any incoming or queued calls will be rejected.
 
-A canister after *uninstalling* retains its *cycles* balance, *controllers*, history, status, and allocations.
+A canister after *uninstalling* retains its *cycle* balances, *controllers*, history, status, and allocations.
 
 The optional `sender_canister_version` parameter can contain the caller's canister version. If provided, its value must be equal to `ic0.canister_version`.
 
@@ -2636,6 +2646,8 @@ The [WebAssembly System API](#system-api) is relatively low-level, and some of i
           controllers : List Principal;
           global_timer : Nat;
           balance : Nat;
+          reserved_balance : Nat;
+          reserved_balance_limit : Nat;
           compute_allocation : Nat;
           memory_allocation : Nat;
           memory_usage_raw_module : Nat;
@@ -2931,6 +2943,8 @@ Finally, we can describe the state of the IC as a record having the following fi
       time : CanisterId ↦ Timestamp;
       global_timer : CanisterId ↦ Timestamp;
       balances: CanisterId ↦ Nat;
+      reserved_balances: CanisterId ↦ Nat;
+      reserved_balance_limits: CanisterId ↦ Nat;
       certified_data: CanisterId ↦ Blob;
       canister_history: CanisterId ↦ CanisterHistory;
       system_time : Timestamp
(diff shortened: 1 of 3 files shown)


spec/index.md
```diff
@@ -245,19 +247,19 @@ If an empty canister receives a response, that response is dropped, as if the ca
 
 #### Canister cycles {#canister-cycles}
 
-The IC relies on *cycles*, a utility token, to manage its resources. A canister pays for the resources it uses from its *cycle balance*. The *cycle\_balance* is stored as 128-bit unsigned integers and operations on them are saturating. In particular, if *cycles* are added to a canister that would bring its total balance beyond 2<sup>128</sup>-1, then the balance will be capped at 2<sup>128</sup>-1 and any additional cycles will be lost.
+The IC relies on *cycles*, a utility token, to manage its resources. A canister pays for the resources it uses from its *cycle balances*. A *cycle\_balance* is stored as 128-bit unsigned integers and operations on them are saturating. In particular, if *cycles* are added to a canister that would bring its main cycle balance beyond 2<sup>128</sup>-1, then the balance will be capped at 2<sup>128</sup>-1 and any additional cycles will be lost.
```
mraszyk, line 250: I suppose reserved cycles would also be capped at u128::max

mraszyk: Due to the reserved cycles limit, they can never exceed u128::max.

spec/index.md
```diff
@@ -1946,6 +1948,14 @@ The optional `settings` parameter can be used to set the following settings:
 
     Default value: 2592000 (approximately 30 days).
 
+-   `reserved_cycles_limit` (`nat`)
+
+    Must be a number between 0 and 2<sup>128</sup>-1, inclusively, and indicates the upper limit on `reserved_cycles` of the canister.
+
+    An operation that allocates resources such as compute and memory will fail of the new value of `reserved_cycles` exceeds this limit.
```
mraszyk, line 1955:
```suggestion
    An operation that allocates resources such as compute and memory will fail if the new value of `reserved_cycles` exceeds this limit.
```

spec/index.md
```diff
@@ -2959,6 +2973,27 @@ To convert `CallOrigin` into `ChangeOrigin`, we define the following conversion
         canister_version = sender_canister_version
       }
 
+#### Cycle bookkeeping and resource consumption
+
+The main cycle balance of canister `A` in state `S` can be obtained with `S.balances(A)`.
+In addition to the main balance, each canister has a reserved balance `S.reserved_balances(A)`.
+The reserved balance contains cycles that were set aside from the main balance for future payments for the consumption of resources such as memory and compute.
```
mraszyk, line 2980:
```suggestion
The reserved balance contains cycles that were set aside from the main balance for future payments for the consumption of resources such as compute and memory.
```
for the sake of consistency

spec/index.md
```diff
@@ -2959,6 +2973,27 @@ To convert `CallOrigin` into `ChangeOrigin`, we define the following conversion
...
+The reserved cycles can only be used for resource payments and cannot be transferred back to the main balance.
+
+The (unspecified) function `idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size)` determines the idle resource consumption rate in cycles per day of a canister given its current compute and memory allocation, memory usage, and subnet size. The function `freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size)` determines the freezing limit in cycles of a canister given its current compute and memory allocation, freezing threshold in seconds, memory usage & and subnet size. The value `freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size)` is derived from `idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size)` and `freezing_threshold` as follows:
```
mraszyk, line 2983:
```suggestion
The (unspecified) function `idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size)` determines the idle resource consumption rate in cycles per day of a canister given its current compute and memory allocation, memory usage, and subnet size. The function `freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size)` determines the freezing limit in cycles of a canister given its current compute and memory allocation, freezing threshold in seconds, memory usage, and subnet size. The value `freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size)` is derived from `idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size)` and `freezing_threshold` as follows:
```

spec/index.md
```diff
@@ -2959,6 +2973,27 @@ To convert `CallOrigin` into `ChangeOrigin`, we define the following conversion
...
+
+        freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size) = idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size) * freezing_threshold / (24 * 60 * 60)
+
+The (unspecified) functions `memory_usage_wasm_state(wasm_state)`, `memory_usage_raw_module(raw_module)`, and `memory_usage_canister_history(canister_history)` determine the canister's memory usage in bytes consumed by its Wasm state, raw Wasm binary, and canister history, respectively.
+
+The amount of cycles that is available for spending in calls and execution is computed by the `liquid_balance(balance, reserved_balance, freezing_limit)` function:
```
mraszyk, line 2989:
```suggestion
The amount of cycles that is available for spending in calls and execution is computed by the function `liquid_balance(balance, reserved_balance, freezing_limit)`:
```

spec/index.md
```diff
@@ -3124,15 +3157,19 @@ is_effective_canister_id(E.content, ECID)
...
-    S.memory_allocation[E.content.canister_id],
-    S.freezing_threshold[E.content.canister_id],
-    memory_usage_wasm_state(S.canisters[E.content.canister_id].wasm_state) +
-      memory_usage_raw_module(S.canisters[E.content.canister_id].raw_module) +
-      memory_usage_canister_history(S.canister_history[E.content.canister_id]),
-    S.canister_subnet[E.content.canister_id].subnet_size,
-  )
+  liquid_balance(
+    S.balances[E.content.canister_id],
+    S.reserved_balance(E.content.canister_id),
```
mraszyk, line 3162:
```suggestion
    S.reserved_balances[E.content.canister_id],
```
same everywhere else

spec/index.md
```diff
@@ -5026,12 +5163,32 @@ S with
 
 ```
 
-The canister cycle balance similarly depletes at an unspecified rate, but stays non-negative:
+The canister cycle balances similarly deplete at an unspecified rate, but stays non-negative.
```
mraszyk, line 5166:
```suggestion
The canister cycle balances similarly deplete at an unspecified rate, but stay non-negative.
```

spec/index.md
```diff
@@ -3680,14 +3744,6 @@ if A.settings.controllers is not null:
 else:
   New_controllers = [M.caller]
 
-if New_compute_allocation > 0 or New_memory_allocation > 0:
```
mraszyk, line 3683: I recall that the freezing limit was not checked if the compute and memory allocation did not change. Now you specify it to be always checked. Did the implementation change in this regard?

spec/index.md
```diff
@@ -3732,6 +3808,8 @@ S with
     memory_allocation[CanisterId] = New_memory_allocation
     freezing_threshold[CanisterId] = New_freezing_threshold
     balances[CanisterId] = M.transferred_cycles
+    reserved_balances[Canister_id] = 0
```
mraszyk, line 3811:
```suggestion
    reserved_balances[Canister_id] = New_reserved_balance
```
please double check

spec/index.md
```diff
@@ -3781,15 +3859,19 @@ M.arg = candid(A)
...
-      memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
-      memory_usage_canister_history(New_canister_history),
-    S.canister_subnet[A.canister_id].subnet_size,
-  ) ≤ S.balances[A.canister_id]
+  liquid_balance(
+    S.balances[A.canister_id],
+    S.reserved_balances[A.canister_id],
+    freezing_limit(
+      New_compute_allocation,
+      New_memory_allocation,
```
mraszyk, line 3867: I guess a new memory allocation could also cause new cycles to become reserved here.

spec/index.md
```diff
@@ -4747,10 +4862,29 @@ if A.settings.freezing_threshold is not null:
...
+  New_reserved_balance_limit = 5_000_000_000_000
+
+Cycles_reserved = cycles_to_reserve(S, New_compute_allocation, New_memory_allocation, EmptyCanister)
 if A.amount is not null:
-  New_balance = A.amount
+  New_balance = A.amount - Cycles_reserved
 else:
-  New_balance = DEFAULT_PROVISIONAL_CYCLES_BALANCE
+  New_balance = DEFAULT_PROVISIONAL_CYCLES_BALANCE - Cycles_reserved
+New_reserved_balance  = Cycles_reserved
```
mraszyk, line 4875:
```suggestion
New_reserved_balance = Cycles_reserved
```

spec/index.md
```diff
@@ -4723,15 +4847,6 @@ if A.settings.controllers is not null:
   New_controllers = A.settings.controllers
 else:
   New_controllers = [M.caller]
-
-if New_compute_allocation > 0 or New_memory_allocation > 0:
```
mraszyk, line 4727: same comment as for canister creation applies here: it used to be that the implementation would only check the freezing limit under this if-condition

spec/index.md
```diff
@@ -4173,6 +4277,8 @@ Env = {
   time = S.time[A.canister_id];
   controllers = S.controllers[A.canister_id];
   balance = S.balances[A.canister_id];
+  reserved_balance = S.reserved_balances[A.canister_id];
+  reserved_balance_limit = S.reserved_balance_limits[A.canister_id];
```
mraszyk, line 4281: I suppose these values can change in Env2 (defined below) passed to post-upgrade, no?

spec/index.md
```diff
@@ -3703,6 +3759,26 @@ if A.settings.freezing_threshold is not null:
   New_freezing_threshold = A.settings.freezing_threshold
 else:
   New_freezing_threshold = 2592000
+if A.settings.reserved_cycles_limit is not null:
+  New_reserved_balance_limit = A.settings.reserved_cycles_limit
+else:
+  New_reserved_balance_limit = 5_000_000_000_000
+
+Cycles_reserved = cycles_to_reserve(S, New_compute_allocation, New_memory_allocation, EmptyCanister)
```
mraszyk, line 3767:
```suggestion
Cycles_reserved = cycles_to_reserve(S, New_compute_allocation, New_memory_allocation, EmptyCanister.wasm_state)
```

spec/index.md
mraszyk: The implementation does not seem to have changed:
https://github.com/dfinity/ic/blob/b53c7fff38f7e5adc324283b162abd86bf356d4e/rs/execution_environment/src/execution/install_code.rs#L230

spec/index.md
mraszyk: https://github.com/dfinity/ic/blob/b53c7fff38f7e5adc324283b162abd86bf356d4e/rs/execution_environment/src/execution/install_code.rs#L230

spec/index.md
ulan: I think that might be unintentional in the implementation and a side-effect of that helper being used both for canister creation and changing settings.

If we just look at canister creation, it makes sense to refuse creating canister that would immediately become frozen. I would suggest adjusting the implementation for that. Wdyt?

If you prefer to add the check here in spec, I can also do that.

spec/index.md
ulan: Yep, Env2 should use the initial reserved cycles balance since we calculate cycles to reserve at the end of the entire execution.

spec/index.md
```diff
@@ -3723,26 +3801,28 @@ State after
 ```html
 
 S with
-    canisters[CanisterId] = EmptyCanister
```
ulan, line 3726: I changed `CanisterId` to `Canister_id` to have consistent naming with the other canister creation algorith.

spec/index.md
mraszyk: Thank you for adjusting the spec! I agree that creating a canister that is immediately frozen doesn't make much sense, but I'm not sure how relevant this is in practice (so I'm also fine if we keep the existing behavior now that we properly specify it).

Dfinity-Bjoern (approved):

infu:
@ulan 
I've got a problem with this code inside a Motoko canister (dfx 17.0)
```
await ic.create_canister({
      settings = ?{
          controllers = ?[Principal.fromActor(this)];
          freezing_threshold = ?2_000_000_000_000;
          memory_allocation = ?1000000000; // ~1gb
          compute_allocation = null;
          reserved_cycles_limit = ?(2_000_000_000_000);
      };
  sender_canister_version = null
});
```
Produces error: 
IC0503: Canister toj6n-haaaa-aaaal-qdika-cai trapped explicitly: Canister creation failed "Cannot increase memory allocation to 953.67 MiB due to insufficient cycles. At least 236_552_033_935_185_185 additional cycles are required."

I've tried setting reserved_cycles_limit to 0 to null and to 2_000_000_000_000, but nothing changes. It asks for 236,552TC for 1GB ? and asks for 500TC for 2MB 

This happens with canisters on subnets that show 91GB and 120GB states (from dashboard)
https://dashboard.internetcomputer.org/canister/toj6n-haaaa-aaaal-qdika-cai
https://dashboard.internetcomputer.org/canister/pnhvg-fiaaa-aaaam-abzia-cai

PR #259, opened 2023-12-01
feat: Canister backup and restore 

AlexandraZapuc:
Adds new API to the management canister that facilitates backup and restore operations.  Design doc: [backup and restore](https://docs.google.com/document/d/1vIZGpdYALSkeOugK9EQiaTiiAGKw8EOn__etQjGJUwM/edit)

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -113,6 +113,14 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type snapshot_id = nat64;
+type snapshot = record {
+  id: snapshot_id;
+  taken_on: timestamp;
+  total_size: bytes;
+  checksum: blob;   
+};
+
 service ic : {
   create_canister : (record {
     settings : opt canister_settings;
@@ -214,6 +222,18 @@ service ic : {
   bitcoin_get_utxos_query: (get_utxos_request) -> (get_utxos_response) query;
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
+
+  // Canister backup and restore
+  take_snapshot: (record { 
+    canister_id: canister_id;
+    duration: opt nat64;
+  }) -> (snapshot_id);
+  load_snapshot: (record { 
+    snapshot_id: snapshot_id;
+    canister_id: canister_id
+  }) -> ();
+  list_snapshots: (record {canister_id : canister_id}) -> (vec snapshot);
+  delete_snapshot: (record {snapshot_id : snapshot_id}) -> ();
 
   // provisional interfaces for the pre-ledger world
   provisional_create_canister_with_cycles : (record {
--- a/spec/index.md
+++ b/spec/index.md
@@ -2301,6 +2301,34 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `take_snapshot` {#ic-take_snapshot}
+
+This method takes a snapshot of the canister. Subsequent `take_snapshot` calls will replace the existing snapshot with a new one.
+
+The optional `duration` parameter can be used to limit the storage duration of a snapshot. When provided, the snapshot will be stored for the specified duration (in seconds) and will be deleted automatically at the end of this duration. In this case,
+subsequent calls to `load_snapshot` or `delete_snapshot` will fail.  
+
+It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the balance must have the right amount of cycles to support the new freezing threshold.
+
+Only controllers can take a snapshot of a canister and restore it.
+
+### IC method `load_snapshot` {#ic-take_snapshot}
+
+This method loads a snapshot identified by `snapshot_id` onto the canister.  
+
+Only controllers can take a snapshot of a canister and restore it.
+
+### IC method `list_snapshots` {#ic-take_snapshot}
+
+This method lists the snapshots of the canister identified by `canister_id`. Currently, only one snapshot per canister will be stored.
+
+### IC method `delete_snapshot` {#ic-take_snapshot}
+
+This method deletes the snapshot of the canister. An error will be returned if the snapshot is not found. A snapshot cannot be 
+found if it was previously deleted, replaced by a new snapshot through a `take_snapshot` request, or if the canister itself has been deleted.
+
+A snapshot may be deleted only by the controllers of the canister which owns the snapshot.
+
 ## Certification {#certification}
 
 Some parts of the IC state are exposed to users in a tamperproof way via certification: the IC can reveal a *partial state tree* which includes just the data of interest, together with a signature on the root hash of the state tree. This means that a user can be sure that the response is correct, even if the user happens to be communicating with a malicious node, or has received the certificate via some other untrusted way.
@@ -3079,7 +3107,7 @@ The (unspecified) function `idle_cycles_burned_rate(compute_allocation, memory_a
 
         freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size) = idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size) * freezing_threshold / (24 * 60 * 60)
 
-The (unspecified) functions `memory_usage_wasm_state(wasm_state)`, `memory_usage_raw_module(raw_module)`, and `memory_usage_canister_history(canister_history)` determine the canister's memory usage in bytes consumed by its Wasm state, raw Wasm binary, and canister history, respectively.
+The (unspecified) functions `memory_usage_wasm_state(wasm_state)`, `memory_usage_raw_module(raw_module)`,`memory_usage_canister_history(canister_history)` and `memory_usage_snapshot(snapshot)` determine the canister's memory usage in bytes consumed by its Wasm state, raw Wasm binary, and canister history, respectively.
 
 Submitted request  
 `E : Envelope`
@@ -3128,7 +3156,8 @@ is_effective_canister_id(E.content, ECID)
     S.freezing_threshold[E.content.canister_id],
     memory_usage_wasm_state(S.canisters[E.content.canister_id].wasm_state) +
       memory_usage_raw_module(S.canisters[E.content.canister_id].raw_module) +
-      memory_usage_canister_history(S.canister_history[E.content.canister_id]),
+      memory_usage_canister_history(S.canister_history[E.content.canister_id]) +
+      memory_usage_snapshot(S.snapshots[E.content.canister_id]),
     S.canister_subnet[E.content.canister_id].subnet_size,
   )
   S.canisters[E.content.canister_id].module.inspect_message
@@ -3228,7 +3257,8 @@ S.canister_status[CM.callee] = Stopped or S.canister_status[CM.callee] = Stoppin
   S.freezing_threshold[CM.callee],
   memory_usage_wasm_state(S.canisters[CM.callee].wasm_state) +
     memory_usage_raw_module(S.canisters[CM.callee].raw_module) +
-    memory_usage_canister_history(S.canister_history[CM.callee]),
+    memory_usage_canister_history(S.canister_history[CM.callee]) +
+    memory_usage_snapshot(S.snapshots[CM.callee]),
   S.canister_subnet[CM.callee].subnet_size,
 )
 
@@ -3272,7 +3302,8 @@ S.balances[CM.callee] ≥ freezing_limit(
   S.freezing_threshold[CM.callee],
   memory_usage_wasm_state(S.canisters[CM.callee].wasm_state) +
     memory_usage_raw_module(S.canisters[CM.callee].raw_module) +
-    memory_usage_canister_history(S.canister_history[CM.callee]),
+    memory_usage_canister_history(S.canister_history[CM.callee]) +
+    memory_usage_snapshot(S.snapshots[CM.callee]),
   S.canister_subnet[CM.callee].subnet_size,
 ) + MAX_CYCLES_PER_MESSAGE
 Ctxt_id ∉ dom(S.call_contexts)
@@ -3320,7 +3351,8 @@ S.balances[C] ≥ freezing_limit(
   S.freezing_threshold[C],
   memory_usage_wasm_state(S.canisters[C].wasm_state) +
     memory_usage_raw_module(S.canisters[C].raw_module) +
-    memory_usage_canister_history(S.canister_history[C]),
+    memory_usage_canister_history(S.canister_history[C]) +
+    memory_usage_snapshot(S.snapshots[C]),
   S.canister_subnet[C].subnet_size,
 ) + MAX_CYCLES_PER_MESSAGE
 Ctxt_id ∉ dom(S.call_contexts)
@@ -3369,7 +3401,8 @@ S.balances[C] ≥ freezing_limit(
   S.freezing_threshold[C],
   memory_usage_wasm_state(S.canisters[C].wasm_state) +
     memory_usage_raw_module(S.canisters[C].raw_module) +
-    memory_usage_canister_history(S.canister_history[C]),
+    memory_usage_canister_history(S.canister_history[C])
+    memory_usage_snapshot(S.snapshots[C]),
   S.canister_subnet[C].subnet_size,
 ) + MAX_CYCLES_PER_MESSAGE
 Ctxt_id ∉ dom(S.call_contexts)
@@ -3486,12 +3519,14 @@ if
     S.freezing_threshold[M.receiver],
     memory_usage_wasm_state(res.new_state) +
       memory_usage_raw_module(S.canisters[M.receiver].raw_module) +
-      memory_usage_canister_history(S.canister_history[M.receiver]),
+      memory_usage_canister_history(S.canister_history[M.receiver]) +
+      memory_usage_snapshot(S.snapshots[M.receiver]),
     S.canister_subnet[M.receiver].subnet_size,
   );
   (S.memory_allocation[M.receiver] = 0) or (memory_usage_wasm_state(res.new_state) +
     memory_usage_raw_module(S.canisters[M.receiver].raw_module) +
-    memory_usage_canister_history(S.canister_history[M.receiver]) ≤ S.memory_allocation[M.receiver])
+    memory_usage_canister_history(S.canister_history[M.receiver] +
+    memory_usage_snapshot(S.snapshots[M.receiver])) ≤ S.memory_allocation[M.receiver])
   (res.response = NoResponse) or S.call_contexts[M.call_context].needs_to_respond
 then
   S with
@@ -3785,13 +3820,15 @@ if New_compute_allocation > S.compute_allocation[A.canister_id] or New_memory_al
     New_freezing_threshold,
     memory_usage_wasm_state(S.canisters[A.canister_id].wasm_state) +
       memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
-      memory_usage_canister_history(New_canister_history),
+      memory_usage_canister_history(New_canister_history) +
+      memory_usage_snapshot(S.snapshots[A.canister_id]),
     S.canister_subnet[A.canister_id].subnet_size,
   ) ≤ S.balances[A.canister_id]
 if New_memory_allocation > 0:
   memory_usage_wasm_state(S.canisters[A.canister_id].wasm_state) +
     memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
-    memory_usage_canister_history(New_canister_history) ≤ New_memory_allocation
+    memory_usage_canister_history(New_canister_history) +
+    memory_usage_snapshot(S.snapshots[A.canister_id]) ≤ New_memory_allocation
 
 if A.settings.compute_allocation is not null:
   New_compute_allocation = A.settings.compute_allocation
@@ -3892,7 +3929,8 @@ S with
             S.memory_allocation[A.canister_id],
             memory_usage_wasm_state(S.canisters[A.canister_id].wasm_state) +
               memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
-              memory_usage_canister_history(S.canister_history[A.canister_id]),
+              memory_usage_canister_history(S.canister_history[A.canister_id]) +
+              memory_usage_snapshot(S.snapshots[A.canister_id]),
             S.freezing_threshold[A.canister_id],
             S.canister_subnet[A.canister_id].subnet_size,
           );
@@ -4078,7 +4116,8 @@ freezing_limit(
   S.freezing_threshold[A.canister_id],
   memory_usage_wasm_state(S.canisters[A.canister_id].wasm_state) +
     memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
-    memory_usage_canister_history(S.canister_history[A.canister_id]),
+    memory_usage_canister_history(S.canister_history[A.canister_id]) +
+    memory_usage_snapshot(S.snapshots[A.canister_id]),
   S.canister_subnet[A.canister_id].subnet_size,
 ) + MAX_CYCLES_PER_MESSAGE ≤ S.balances[A.canister_id]
 freezing_limit(
@@ -4087,13 +4126,15 @@ freezing_limit(
   S.freezing_threshold[A.canister_id],
   memory_usage_wasm_state(New_state) +
     memory_usage_raw_module(A.wasm_module) +
-    memory_usage_canister_history(New_canister_history),
+    memory_usage_canister_history(New_canister_history) +
+    memory_usage_snapshot(S.snapshots[A.canister_id]),
   S.canister_subnet[A.canister_id].subnet_size,
 ) + Cycles_used ≤ S.balances[A.canister_id]
 if S.memory_allocation[A.canister_id] > 0:
   memory_usage_wasm_state(New_state) +
     memory_usage_raw_module(A.wasm_module) +
-    memory_usage_canister_history(New_canister_history),
+    memory_usage_canister_history(New_canister_history) +
+    memory_usage_snapshot(S.snapshots[A.canister_id]),
   ) ≤ S.memory_allocation[A.canister_id]
 
 S.canister_history[A.canister_id] = {
@@ -4211,7 +4252,8 @@ freezing_limit(
   S.freezing_threshold[A.canister_id],
   memory_usage_wasm_state(S.canisters[A.canister_id].wasm_state) +
     memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
-    memory_usage_canister_history(S.canister_history[A.canister_id]),
+    memory_usage_canister_history(S.canister_history[A.canister_id]) +
+    memory_usage_snapshot(S.snapshots[A.canister_id]),
   S.canister_subnet[A.canister_id].subnet_size,
 ) + MAX_CYCLES_PER_MESSAGE ≤ S.balances[A.canister_id]
 freezing_limit(
@@ -4220,13 +4262,15 @@ freezing_limit(
   S.freezing_threshold[A.canister_id],
   memory_usage_wasm_state(New_state) +
     memory_usage_raw_module(A.wasm_module) +
-    memory_usage_canister_history(New_canister_history),
+    memory_usage_canister_history(New_canister_history) +
+    memory_usage_snapshot(S.snapshots[A.canister_id]),
   S.canister_subnet[A.canister_id].subnet_size,
 ) + Cycles_used + Cycles_used' ≤ S.balances[A.canister_id]
 if S.memory_allocation[A.canister_id] > 0:
   memory_usage_wasm_state(New_state) +
     memory_usage_raw_module(A.wasm_module) +
-    memory_usage_canister_history(New_canister_history) ≤ S.memory_allocation[A.canister_id]
+    memory_usage_canister_history(New_canister_history) +
+    memory_usage_snapshot(S.snapshots[A.canister_id]) ≤ S.memory_allocation[A.canister_id]
 
 S.canister_history[A.canister_id] = {
   total_num_changes = N;
@@ -5157,8 +5201,9 @@ We define an auxiliary method that handles calls from composite query methods by
            S.memory_allocation[Canister_id],
            S.freezing_threshold[Canister_id],
            memory_usage_wasm_state(S.canisters[Canister_id].wasm_state) +
-             memory_usage_raw_module(S.canisters[Canister_id].raw_module) +
-             memory_usage_canister_history(S.canister_history[Canister_id]),
+              memory_usage_raw_module(S.canisters[Canister_id].raw_module) +
+              memory_usage_canister_history(S.canister_history[Canister_id]) +
+              memory_usage_snapshot(S.snapshots[Canister_id]),
            S.canister_subnet[Canister_id].subnet_size,
          ) and
          (Method_name ∈ dom(Mod.query_methods) or Method_name ∈ dom(Mod.composite_query_methods)) and
@@ -5956,7 +6001,10 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
         es.params.sysenv.compute_allocation,
         es.params.sysenv.memory_allocation,
         es.params.sysenv.freezing_threshold,
-        memory_usage_wasm_state(es.wasm_state) + es.params.sysenv.memory_usage_raw_module + es.params.sysenv.memory_usage_canister_history,
+        memory_usage_wasm_state(es.wasm_state) + 
+          es.params.sysenv.memory_usage_raw_module + 
+          es.params.sysenv.memory_usage_canister_history +
+          es.params.sysenv.memory_usage_snapshot,
         es.params.sysenv.subnet_size,
       ))
       es.balance := es.balance - burned_amount
@@ -6063,7 +6111,10 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
         es.params.sysenv.compute_allocation,
         es.params.sysenv.memory_allocation,
         es.params.sysenv.freezing_threshold,
-        memory_usage_wasm_state(es.wasm_state) + es.params.sysenv.memory_usage_raw_module + es.params.sysenv.memory_usage_canister_history,
+        memory_usage_wasm_state(es.wasm_state) + 
+          es.params.sysenv.memory_usage_raw_module + 
+          es.params.sysenv.memory_usage_canister_history +
+          es.params.sysenv.memory_usage_snapshot,
         es.params.sysenv.subnet_size,
       ) then Trap {cycles_used = es.cycles_used;}
 
@@ -6079,7 +6130,10 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
         es.params.sysenv.compute_allocation,
         es.params.sysenv.memory_allocation,
         es.params.sysenv.freezing_threshold,
-        memory_usage_wasm_state(es.wasm_state) + es.params.sysenv.memory_usage_raw_module + es.params.sysenv.memory_usage_canister_history,
+        memory_usage_wasm_state(es.wasm_state) + 
+          es.params.sysenv.memory_usage_raw_module + 
+          es.params.sysenv.memory_usage_canister_history +
+          es.params.sysenv.memory_usage_snapshot,
         es.params.sysenv.subnet_size,
       ) then Trap {cycles_used = es.cycles_used;}
 
@@ -6096,7 +6150,10 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
         es.params.sysenv.compute_allocation,
         es.params.sysenv.memory_allocation,
         es.params.sysenv.freezing_threshold,
-        memory_usage_wasm_state(es.wasm_state) + es.params.sysenv.memory_usage_raw_module + es.params.sysenv.memory_usage_canister_history,
+        memory_usage_wasm_state(es.wasm_state) + 
+          es.params.sysenv.memory_usage_raw_module + 
+          es.params.sysenv.memory_usage_canister_history +
+          es.params.sysenv.memory_usage_snapshot,
         es.params.sysenv.subnet_size,
       ) or system_cannot_do_this_call_now()
       then


nomeata:
Will this require a stopped canister, or will it work even with outstanding callbacks?

spec/index.md
```diff
@@ -2301,6 +2301,34 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `take_snapshot` {#ic-take_snapshot}
+
+This method takes a snapshot of the canister. Subsequent `take_snapshot` calls will replace the existing snapshot with a new one.
```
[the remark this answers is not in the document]
Dfinity-Bjoern: Absolutely!

spec/index.md
```diff
@@ -2301,6 +2301,34 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+The optional `duration` parameter can be used to limit the storage duration of a snapshot. When provided, the snapshot will be stored for the specified duration (in seconds) and will be deleted automatically at the end of this duration. In this case,
```
Dfinity-Bjoern, line 2308: I think `duration` is a weird name. It sounds like it would be the duration of the `take_snapshot`. Would it maybe be easier to name this `delete_at_timestamp` or `preserve_until_timestamp` or so?

spec/_attachments/ic.did
```diff
@@ -215,6 +223,18 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // Canister backup and restore
+  take_snapshot: (record { 
+    canister_id: canister_id;
```
mraszyk, line 228:
```suggestion
    canister_id: principal;
```

spec/_attachments/ic.did
```diff
@@ -215,6 +223,18 @@ service ic : {
...
+    duration: opt nat64;
+  }) -> (snapshot_id);
+  load_snapshot: (record { 
+    snapshot_id: snapshot_id;
+    canister_id: canister_id
```
mraszyk, line 233:
```suggestion
    canister_id: principal
```

spec/_attachments/ic.did
```diff
@@ -215,6 +223,18 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // Canister backup and restore
+  take_snapshot: (record { 
+    canister_id: canister_id;
+    duration: opt nat64;
+  }) -> (snapshot_id);
+  load_snapshot: (record { 
+    snapshot_id: snapshot_id;
```
mraszyk, line 232: In the previous method, we have `canister_id` as the first field - why not do the same here?

spec/index.md
```diff
@@ -3079,7 +3107,7 @@ The (unspecified) function `idle_cycles_burned_rate(compute_allocation, memory_a
 
         freezing_limit(compute_allocation, memory_allocation, freezing_threshold, memory_usage, subnet_size) = idle_cycles_burned_rate(compute_allocation, memory_allocation, memory_usage, subnet_size) * freezing_threshold / (24 * 60 * 60)
 
-The (unspecified) functions `memory_usage_wasm_state(wasm_state)`, `memory_usage_raw_module(raw_module)`, and `memory_usage_canister_history(canister_history)` determine the canister's memory usage in bytes consumed by its Wasm state, raw Wasm binary, and canister history, respectively.
+The (unspecified) functions `memory_usage_wasm_state(wasm_state)`, `memory_usage_raw_module(raw_module)`,`memory_usage_canister_history(canister_history)` and `memory_usage_snapshot(snapshot)` determine the canister's memory usage in bytes consumed by its Wasm state, raw Wasm binary, and canister history, respectively.
```
mraszyk, line 3110:
```suggestion
The (unspecified) functions `memory_usage_wasm_state(wasm_state)`, `memory_usage_raw_module(raw_module)`, `memory_usage_canister_history(canister_history)` and `memory_usage_snapshot(snapshot)` determine the canister's memory usage in bytes consumed by its Wasm state, raw Wasm binary, and canister history, respectively.
```

spec/index.md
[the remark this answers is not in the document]
mraszyk: Moreover, ideally we'd also have model snapshots in the IC state and have dedicated transitions in the formal semantics.

spec/index.md
```diff
@@ -2301,6 +2301,34 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+Only controllers can take a snapshot of a canister and restore it.
+
+### IC method `list_snapshots` {#ic-take_snapshot}
+
+This method lists the snapshots of the canister identified by `canister_id`. Currently, only one snapshot per canister will be stored.
+
+### IC method `delete_snapshot` {#ic-take_snapshot}
+
+This method deletes the snapshot of the canister. An error will be returned if the snapshot is not found. A snapshot cannot be 
+found if it was previously deleted, replaced by a new snapshot through a `take_snapshot` request, or if the canister itself has been deleted.
```
mraszyk, line 2328: > replaced by a new snapshot through a `take_snapshot` request

it sounds weird that the API is formulated as if multiple snapshots per canister were allowed but there's also the implicit replacement of an old snapshot if `take_snapshot` is called - wouldn't it make more sense to fail `take_snapshot` if there already exists a snapshot (forcing the user to call `delete_snapshot` before calling `take_snapshot` again)? I guess this is how it's gonna work once the 1-snapshot restriction is lifted in the future (if it's not gonna be lifted, then we should have vectors and plural in the API).

spec/index.md
```diff
@@ -2301,6 +2301,34 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+A snapshot may be deleted only by the controllers of the canister which owns the snapshot.
```
mraszyk, line 2330:
```suggestion
A snapshot may be deleted only by the controllers of the canister for which the snapshot was taken.
```

spec/_attachments/ic.did
```diff
@@ -113,6 +113,14 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type snapshot_id = nat64;
+type snapshot = record {
+  id: snapshot_id;
+  taken_on: timestamp;
+  total_size: bytes;
```
mraszyk, line 120:
```suggestion
  total_size: nat64;
```

mraszyk:
> Will this require a stopped canister, or will it work even with outstanding callbacks?

In the design doc, there's a note that "While it is not explicitly enforced, creating a snapshot only after stopping the canister is recommended. This follows the same principle as upgrading a canister, as making sense of the callbacks may not be possible.". I'd suggest to add this recommendation as a note to the spec, too.

spec/index.md
AlexandraZapuc: Sure, we rename it, `preserve_until_timestamp` sounds better, or `preserve_duration`?

AlexandraZapuc:
> Will this require a stopped canister, or will it work even with outstanding callbacks?  

As discussed with @ielashi and @bogwar, we will not force the canister to be stopped, but similar to `install_code`, that would be the recommendation.  

> > Will this require a stopped canister, or will it work even with outstanding callbacks?

>In the design doc, there's a note that "While it is not explicitly enforced, creating a snapshot only after stopping the canister is recommended. This follows the same principle as upgrading a canister, as making sense of the callbacks may not be possible.". I'd suggest to add this recommendation as a note to the spec, too.
 
 
 @mraszyk  sure, will do!

spec/index.md
mraszyk: cc @AlexandraZapuc

oggy-dfin:
> As discussed with @ielashi and @bogwar, we will not force the canister to be stopped, but similar to `install_code`, that would be the recommendation.

I think it is indeed a good choice to allow taking snapshots even with open call context, in order to cover disaster scenarios. But it might be good to expose a flag to control whether this is enabled (maybe even make disable it by default).

spec/index.md
```diff
@@ -2330,6 +2330,34 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `take_snapshot` {#ic-take_snapshot}
+
+This method takes a snapshot of the canister. Subsequent `take_snapshot` calls will replace the existing snapshot with a new one.
+
+The optional `duration` parameter can be used to limit the storage duration of a snapshot. When provided, the snapshot will be stored for the specified duration (in seconds) and will be deleted automatically at the end of this duration. In this case,
```
ielashi, line 2337: Is this feature necessary for the V1 of snapshotting? I think the concept of a duration for deleting snapshots may also clash with the ongoing work on pre-emptive storage charging that's being done by the runtime team, so it might be better to avoid it for now and potentially introduce it later if the need arises.

spec/_attachments/ic.did
```diff
@@ -113,6 +113,14 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type snapshot_id = nat64;
```
ielashi, line 116: I remember the recommendation that we should be using `nat` for all numbers unless there's a strong reason not to. @chenyan-dfinity can you confirm if that's correct?

spec/index.md
AlexandraZapuc: I had a discussion today with Ulan. I will remove `duration` altogether. I was informed that there had been some work on `reserved_cycles,` which reserves cycles from the balance for resource allocation for a longer period of time. We will integrate the snapshots with `reserve_cycles` mechanism, so the duration does not make sense anymore. Will update the design.

spec/index.md
AlexandraZapuc: Yes, that is the case. I discussed with Ulan about this today. So I removed it. I will update the doc about the behavior considering the `reserve_cycles.`

spec/index.md
AlexandraZapuc: > it sounds weird that the API is formulated as if multiple snapshots per canister were allowed  

@mraszyk  Which part is formulated as multiple snapshots?  Do you mean this :  `list_snapshots: (record {canister_id : canister_id}) -> (vec snapshot)`?

mraszyk:
Let me also point out that in my opinion if the new argument
```
replace_snapshot: opt snapshot_id
```
is `null` and there's no available slot for the new snapshot, then `take_snapshot` should fail.

spec/_attachments/ic.did
chenyan-dfinity: Another minor point is that `nat` is more space efficient on the wire. `nat64` always takes 8 bytes, while `nat` can have fewer bytes if the actual number is small.

spec/_attachments/ic.did
```diff
@@ -224,6 +232,17 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // Canister backup and restore
+  take_snapshot: (record { 
+    canister_id: canister_id;
+  }) -> (snapshot_id);
+  load_snapshot: (record { 
+    canister_id: canister_id;
```
crusso, line 240:
```suggestion
    canister_id: principal;
```

spec/_attachments/ic.did
```diff
@@ -224,6 +232,17 @@ service ic : {
...
+    snapshot_id: snapshot_id;
+  }) -> ();
+  list_snapshots: (record {canister_id : canister_id}) -> (vec snapshot);
```
crusso, line 243:
```suggestion
  list_snapshots: (record {canister_id : principal}) -> (vec snapshot);
```

spec/_attachments/ic.did
crusso: (assuming @mraszyk suggestions was correct)

spec/_attachments/ic.did
crusso: (assuming @mraszyk suggestions was correct)

spec/index.md
```diff
@@ -2340,6 +2340,38 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+:::note
+
+It's important to stop a canister before taking a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored using the snapshot.
+It is expected that the canister admin (or their tooling) does this separately.
+
+:::
+
+### IC method `load_snapshot` {#ic-take_snapshot}
+
+This method loads a snapshot identified by `snapshot_id` onto the canister.  This method fails if the snapshot was previously deleted.
```
oggy-dfin, line 2360: Is it recommended to try and stop the canister before loading the snapshot? Are there any downsides to loading a snapshot on a non-stopped canister?

mraszyk: > Is it recommended to try and stop the canister before loading the snapshot?
Yes, it should be.

> Are there any downsides to loading a snapshot on a non-stopped canister?
The loaded snapshot might not know how to deal with outstanding callbacks.

spec/_attachments/ic.did
mraszyk: Actually, we have a type synonym `type canister_id = principal;` at the top of the `ic.did` file so it's fine as is.

spec/_attachments/ic.did
mraszyk: Actually, we have a type synonym `type canister_id = principal;` at the top of the `ic.did` file so it's fine as is.

AlexandraZapuc: Canister id is a principal. Would you rather prefer to use explicit `principal`?

spec/index.md
AlexandraZapuc: I have changed the api following the discussion with @mraszyk. I will add the rest of the semantics.

spec/index.md
```diff
@@ -2340,6 +2340,38 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+This method takes a snapshot of the canister. Subsequent `take_snapshot` calls will replace the existing snapshot with a new one.
+
+It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the balance must have the right amount of cycles to support the new freezing threshold.
+
+Only controllers can take a snapshot of a canister and restore it.
+
+:::note
+
+It's important to stop a canister before taking a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored using the snapshot.
```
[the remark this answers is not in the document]
AlexandraZapuc: I added info about what is saved. Let me know if it's not enough.

spec/index.md
AlexandraZapuc: So basically, the requests and responses that are in progress will continue to flow as expected, they will not be touched by the snapshots. However, the canister state can change arbitrarily, so it's up to the developer to decide when to take a snapshot and to restore it. 
 It's a behavior similar to installing code. Loading a snapshot will update the canister module, so again, the canister might not be able to make sense of the callbacks.

spec/index.md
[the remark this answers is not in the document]
AlexandraZapuc: @mraszyk I have updated the semantics too. And I think I have also covered all the other suggestions. Please take a look when you have time.

spec/index.md
```diff
@@ -2359,17 +2359,22 @@ It is expected that the canister admin (or their tooling) does this separately.
...
-This method loads a snapshot identified by `snapshot_id` onto the canister. This method fails if the snapshot was previously deleted.
+This method loads a snapshot identified by `snapshot_id` onto the canister. It fails if the snapshot was previously deleted.
 
 Only controllers can take a snapshot of a canister and restore it.
 
+:::note
+
+It's important to stop a canister before loading a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored.
+It is expected that the canister admin (or their tooling) does this separately.
+
```
mraszyk, line 2370: We should close the note environment here to fix markdown deployment.

spec/index.md
```diff
@@ -2359,17 +2359,22 @@ It is expected that the canister admin (or their tooling) does this separately.
...
 
-This method loads a snapshot identified by `snapshot_id` onto the canister. This method fails if the snapshot was previously deleted.
+This method loads a snapshot identified by `snapshot_id` onto the canister. It fails if the snapshot was previously deleted.
 
 Only controllers can take a snapshot of a canister and restore it.
 
+:::note
+
+It's important to stop a canister before loading a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored.
+It is expected that the canister admin (or their tooling) does this separately.
```
mraszyk, line 2369: canister admin -> canister controllers

spec/index.md
```diff
@@ -2342,7 +2342,9 @@ The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Perc
 
 ### IC method `take_snapshot` {#ic-take_snapshot}
 
-This method takes a snapshot of the canister. Subsequent `take_snapshot` calls will replace the existing snapshot with a new one.
+This method takes a snapshot of the specified canister. A snapshot consists of the wasm memory, stable memory, certified variables, wasm chunk store and binary.
```
mraszyk, line 2345: and binary -> , and wasm binary

mraszyk, line 2345: and binary -> , and wasm binary

spec/index.md
```diff
@@ -2368,7 +2370,8 @@ This method lists the snapshots of the canister identified by `canister_id`. Cur
 ### IC method `delete_snapshot` {#ic-take_snapshot}
 
 This method deletes a specified snapshot of the canister. An error will be returned if the snapshot is not found. 
-A snapshot cannot be found if it was previously deleted, replaced by a new snapshot through a `take_snapshot` request, or if the canister itself has been deleted.
+
+A snapshot cannot be found if it was previously deleted, replaced by a new snapshot through a `take_snapshot` request, or if the canister itself has been deleted or depleted of cycles.
```
mraszyk, line 2379: What if the canister is uninstalled? Do the snapshots get deleted?

spec/index.md
```diff
@@ -5078,6 +5082,61 @@ S with
...
 
+#### IC Management Canister: List snapshots
+
+Only the controllers of the given canister can get a list of the existing snapshots.
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.method_name = 'list_snapshots'
```
mraszyk, line 5208: the receiver is missing - it should be the mgmt canister

spec/index.md
```diff
@@ -5078,6 +5082,61 @@ S with
...
+M.arg = candid(A)
+S.snapshots = Other_snapshots ∪ S.snapshots[A.snapshot_id]
```
mraszyk, line 5210: Why are you updating snapshots here?

mraszyk, line 5210: Why are you updating snapshots here?

spec/index.md
```diff
@@ -5078,6 +5082,61 @@ S with
 
 ```
 
+#### IC Management Canister: List snapshots
+
+Only the controllers of the given canister can get a list of the existing snapshots.
```
mraszyk, line 5202: this condition is missing in the formal semantics below

spec/index.md
```diff
@@ -5078,6 +5082,61 @@ S with
...
+```
+#### IC Management Canister: Delete snapshot
+
+A snapshot may be deleted only by the controllers of the canister for which the snapshot was taken.
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.method_name = 'delete_snapshot'
```
mraszyk, line 5234: the receiver is missing - it should be the mgmt canister

spec/index.md
```diff
@@ -5078,6 +5082,61 @@ S with
...
+M.arg = candid(A)
+M.caller ∈ S.controllers[S.snapshots[A.snapshot_id].canister_id]
+S.snapshots = Other_snapshots ∪ S.snapshots[A.snapshot_id]
```
mraszyk, line 5237: I'd drop this line and define
```
S.snapshots[A.snapshot_id] = (deleted)
```
in the state after (see canister deletion for reference)

spec/index.md
```diff
@@ -5078,6 +5082,61 @@ S with
 
 ```
 
+#### IC Management Canister: List snapshots
```
mraszyk, line 5200: Take snapshot is missing as an IC mgmt canister method here in the formal semantics

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'take_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[A.canister_id]
+
+Cycles_reserved = cycles_to_reserve(S, A.canister_id, S.compute_allocation[A.canister_id], S.memory_allocation[A.canister_id] + memory_usage_snapshot(New_snapshot), S.canisters[A.canister_id])
```
mraszyk, line 5099:
```
S.memory_allocation[A.canister_id] + memory_usage_snapshot(New_snapshot)
```
the addition wouldn't make sense if memory allocation is fixed (i.e., not best-effort)

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+New_balance = S.balances[A.canister_id] - Cycles_used - Cycles_reserved
```
mraszyk, line 5100: What are `Cycles_used`? Let's introduce them (abstractly) in the text above (before "Only controllers...").

spec/index.md
mraszyk: We should pass S.memory_allocation[A.canister_id] here and add a new argument to cycles_to_reserve representing the snapshots in the IC state.

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+) ≥ 0
+
+```
+
+State after  
+
+```html
+
+S with
+    snapshots = Other_snapshots · New_snapshot
```
mraszyk, line 5126:
```suggestion
    snapshots[SnapshotId] = New_snapshot
```
and we should formally define `New_snapshot` above.

spec/_attachments/ic.did
```diff
@@ -115,6 +115,14 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type snapshot_id = nat;
+type snapshot = record {
+  id: snapshot_id;
+  taken_on_timestap: nat;
```
mraszyk, line 121:
```suggestion
  taken_at_timestamp: nat64;
```

spec/_attachments/ic.did
```diff
@@ -115,6 +115,14 @@ type send_transaction_request = record {
...
+  total_size: nat;
```
mraszyk, line 122:
```suggestion
  total_size: nat64;
```

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+    balances[A.canister_id] = New_balance
+    reserved_balances[A.canister_id] = New_reserved_balance
+    messages = Older_messages · Younger_messages ·
+      ResponseMessage {
+        origin = M.origin;
+        response = Reply (candid());
```
mraszyk, line 5132:
```suggestion
        response = Reply (candid(SnapshotId));
```

spec/index.md
```diff
@@ -2340,6 +2340,46 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+Subsequent `take_snapshot` calls will create a new snapshot. However, a `take_snapshot` call might fail if the maximum number of snapshots per canister is reached. This error can be avoided by providing a snapshot ID via the optional `replace_snapshot` parameter. The snapshot identified by the specified ID will be deleted. Currently, only one snapshot per canister is allowed.
+
+It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the balance must have the right amount of cycles to support the new freezing threshold.
+
+Only controllers can take a snapshot of a canister and restore it.
+
+:::note
+
+It's important to stop a canister before taking a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored using the snapshot.
+It is expected that the canister admin (or their tooling) does this separately.
```
mraszyk, line 2356:
```suggestion
It is expected that the canister controllers (or their tooling) do this separately.
```

spec/index.md
```diff
@@ -3583,7 +3630,7 @@ if
   ) ≥ 0
   (S.memory_allocation[M.receiver] = 0) or (memory_usage_wasm_state(res.new_state) +
     memory_usage_raw_module(S.canisters[M.receiver].raw_module) +
-    memory_usage_canister_history(S.canister_history[M.receiver]) ≤ S.memory_allocation[M.receiver])
+    memory_usage_canister_history(S.canister_history[M.receiver] ≤ S.memory_allocation[M.receiver])
```
mraszyk, line 3633: why are we ignoring snapshot memory here?

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+Only the controllers of the given canister can load a snapshot.
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'load_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[S.snapshots[A.snapshot_id].canister_id]
```
mraszyk, line 5151: Above, S.snapshots was indexed by canister ID, here it's indexed by snapshot ID. I suppose, the snapshots will be part of the canister's state in the replica implementation and thus it makes sense to index S.snapshots by canister IDs (and also require a canister ID in the mgmt calls arguments so that the mapping snapshot ID -> canister ID does not need to be computed by the replica separately)

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+
+Env = { 
+  canister_id = S.snapshots[A.snapshot_id].canister_id;
+  balance = S.balances[canister_id];
+  Public_custom_sections = parse_public_custom_sections(S.snapshots[canister_id].wasm_module);
+  Private_custom_sections = parse_private_custom_sections(S.snapshots[canister_id].wasm_module);
+}
+New_balance = balance - Cycles_used - S.reserved_balances[canister_id]
```
mraszyk, line 5159: why deduct the reserved balance here?

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'load_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[S.snapshots[A.snapshot_id].canister_id]
+
+Env = {
```
mraszyk, line 5153: Why are you defining an Env here that is not used later? It seems like you only need to parse the custom sections.

spec/index.md
mraszyk, line 5159: `Cycles_used` should be defined in the text above (before "Only controllers...")

spec/index.md
```diff
@@ -5033,6 +5083,175 @@ S with
...
+State after
+
+```html
+
+S with
+    snapshots = Other_snapshots
+    messages = Older_messages · Younger_messages ·
+      ResponseMessage {
+        origin = M.origin
+        response = candid()
```
mraszyk, line 5250: refunded cycles are missing - please also checkout for the other cases

spec/index.md
```diff
@@ -3193,7 +3234,8 @@ is_effective_canister_id(E.content, ECID)
       S.freezing_threshold[E.content.canister_id],
       memory_usage_wasm_state(S.canisters[E.content.canister_id].wasm_state) +
         memory_usage_raw_module(S.canisters[E.content.canister_id].raw_module) +
-        memory_usage_canister_history(S.canister_history[E.content.canister_id]),
+        memory_usage_canister_history(S.canister_history[E.content.canister_id]) +
+        memory_usage_snapshot(S.snapshots[E.content.canister_id]),
```
mraszyk, line 3238: S.snapshots should be introduced (along with the type) in the definition of the IC state (abbreviated by S here) and initiliazed in the definition of the initial IC state.

AlexandraZapuc: I am not sure where exactly you mean. You mean here: https://github.com/dfinity/interface-spec/blob/master/spec/index.md?plain=1#L2960?

spec/index.md
AlexandraZapuc: No, the controllers can still use the snapshot to restore the state of the canister.

spec/index.md
AlexandraZapuc: How do we introduce this? I mean, I have looked at other definitions, and no method has this defined this.

spec/index.md
mraszyk: yes and also here: https://github.com/dfinity/interface-spec/blob/125ca124ddfcfcda520ca107e266e8faaeb627bc/spec/index.md?plain=1#L3028

spec/index.md
mraszyk: OK. Then let's leave this as is for now.

spec/index.md
AlexandraZapuc: Discussing offline.

spec/index.md
AlexandraZapuc: Clarifying it offline.

spec/index.md
mraszyk: This should also be reflected in the formal semantics - clearing `S.snapshots[A.canister_id]` if a canister is deleted or runs out of cycles.

spec/index.md
AlexandraZapuc: I think I forgot top reply to this one. We don't have canister Id here. The snapshots are stored in the replica, not on canister. If you have any idea how to reflect that in the semantics, please let me know. Otherwise, I will consider this as done.

mraszyk:
Let's not check in the didc binary, please.

spec/_attachments/ic.did
```diff
@@ -224,6 +231,18 @@ service ic : {
...
+  take_snapshot: (record { 
+    canister_id: canister_id;
+    replace_snapshot: opt snapshot_id;
+  }) -> (snapshot_id);
+  load_snapshot: (record { 
+    canister_id: canister_id;
+    snapshot_id: snapshot_id;
+  }) -> ();
+  list_snapshots: (record {canister_id : canister_id}) -> (vec snapshot);
+  delete_snapshot: (record {snapshot_id : snapshot_id}) -> ();
```
mraszyk, line 244: In load_snapshot, we pass canister and snapshot ID while here we only pass snapshot ID. I believe the former is preferred, also for the sake of effective canister ID checks.

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `take_snapshot` {#ic-take_snapshot}
+
+This method takes a snapshot of the specified canister. A snapshot consists of the wasm memory, stable memory, certified variables, wasm chunk store and wasm binary.
```
mraszyk, line 2345: certified variables are missing in the formal definition of `Snapshot = {`

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+Subsequent `take_snapshot` calls will create a new snapshot. However, a `take_snapshot` call might fail if the maximum number of snapshots per canister is reached. This error can be avoided by providing a snapshot ID via the optional `replace_snapshot` parameter. The snapshot identified by the specified ID will be deleted. Currently, only one snapshot per canister is allowed.
+
+It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the balance must have the right amount of cycles to support the new freezing threshold.
+
+Only controllers can take a snapshot of a canister and restore it.
```
mraszyk, line 2351:
```suggestion
Only controllers can take a snapshot of a canister and load it back to the canister.
```

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `take_snapshot` {#ic-take_snapshot}
+
+This method takes a snapshot of the specified canister. A snapshot consists of the wasm memory, stable memory, certified variables, wasm chunk store and wasm binary.
+
+Subsequent `take_snapshot` calls will create a new snapshot. However, a `take_snapshot` call might fail if the maximum number of snapshots per canister is reached. This error can be avoided by providing a snapshot ID via the optional `replace_snapshot` parameter. The snapshot identified by the specified ID will be deleted. Currently, only one snapshot per canister is allowed.
+
+It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the balance must have the right amount of cycles to support the new freezing threshold.
```
mraszyk, line 2349:
```suggestion
It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the canister's balance must have a sufficient amount of cycles to support the new freezing threshold.
```

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `take_snapshot` {#ic-take_snapshot}
+
+This method takes a snapshot of the specified canister. A snapshot consists of the wasm memory, stable memory, certified variables, wasm chunk store and wasm binary.
+
+Subsequent `take_snapshot` calls will create a new snapshot. However, a `take_snapshot` call might fail if the maximum number of snapshots per canister is reached. This error can be avoided by providing a snapshot ID via the optional `replace_snapshot` parameter. The snapshot identified by the specified ID will be deleted. Currently, only one snapshot per canister is allowed.
```
mraszyk, line 2347:
```suggestion
Subsequent `take_snapshot` calls will create a new snapshot. However, a `take_snapshot` call might fail if the maximum number of snapshots per canister is reached. This error can be avoided by providing a snapshot ID via the optional `replace_snapshot` parameter. The snapshot identified by the specified ID will be deleted once a new snapshot has been successfully created. Currently, only one snapshot per canister is allowed.
```

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+:::note
+
+It's important to stop a canister before taking a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored using the snapshot.
+It is expected that the canister controllers (or their tooling) do this separately.
+
+:::
+
+### IC method `load_snapshot` {#ic-take_snapshot}
+
+This method loads a snapshot identified by `snapshot_id` onto the canister. It fails if the snapshot was previously deleted.
```
mraszyk, line 2362:
```suggestion
This method loads a snapshot identified by `snapshot_id` onto the canister. It fails if no snapshot with the specified `snapshot_id` was created or if the snapshot was already deleted.
```

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+Only controllers can take a snapshot of a canister and restore it.
```
mraszyk, line 2364:
```suggestion
Only controllers can take a snapshot of a canister and load it back to the canister.
```

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+:::note
+
+It's important to stop a canister before loading a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored.
+It is expected that the canister controllers (or their tooling) do this separately.
+
+:::
+
+### IC method `list_snapshots` {#ic-take_snapshot}
+
+This method lists the snapshots of the canister identified by `canister_id`. Currently, only one snapshot per canister will be stored.
```
mraszyk, line 2375:
```suggestion
This method lists the snapshots of the canister identified by `canister_id`. Currently, at most one snapshot per canister will be stored.
```

spec/index.md
```diff
@@ -2340,6 +2340,48 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+### IC method `delete_snapshot` {#ic-take_snapshot}
+
+This method deletes a specified snapshot that belongs to an existing canister. An error will be returned if the snapshot is not found. 
+
+A snapshot cannot be found if it was previously deleted, replaced by a new snapshot through a `take_snapshot` request, or if the canister itself has been deleted or depleted of cycles.
```
mraszyk, line 2381:
```suggestion
A snapshot cannot be found if it was never created, it was previously deleted, replaced by a new snapshot through a `take_snapshot` request, or if the canister itself has been deleted or run out of cycles.
```

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+New_reserved_balance ≤ S.reserved_balance_limits[A.canister_id]
+
+liquid_balance(
+  New_balance,
+  New_reserved_balance,
+  freezing_limit(
+    S.compute_allocation[A.canister_id],
+    S.memory_allocation[A.canister_id],
+    S.freezing_threshold[A.canister_id],
+    memory_usage_wasm_state(New_state) +
```
mraszyk, line 5128: there's no new state after taking a snapshot

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+      memory_usage_raw_module(A.wasm_module) +
```
mraszyk, line 5129: A.wasm_module is undefined, the wasm_module must be fetched from the IC state `S...`

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+      memory_usage_canister_history(New_canister_history) +
```
mraszyk, line 5130: Taking snapshots need not be recorded in the canister history, right?

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'load_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[A.canister_id]
+
+New_balance = S.balances[A.canister_id] - Cycles_used
+liquid_balance(
+  New_balance,
+  S.reserved_balances[A.canister_id],
```
mraszyk, line 5173: Loading a snapshot can increase the canister's memory footprint and thus trigger cycle reservation => we need to compute a New_reserved_balance here.

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+  freezing_limit(
+    S.compute_allocation[A.canister_id],
+    S.memory_allocation[A.canister_id],
+    S.freezing_threshold[A.canister_id],
+    memory_usage_wasm_state(S.canisters[A.canister_id]) +
+      memory_usage_raw_module(A.wasm_module) +
+      memory_usage_canister_history(New_canister_history) +
```
mraszyk, line 5180: How is New_canister_history defined? You'd also need to extend the canister history entry enum with a new variant and append it to the canister history.

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+
+New_balance = S.balances[A.canister_id] - Cycles_used
+liquid_balance(
+  New_balance,
+  S.reserved_balances[A.canister_id],
+  freezing_limit(
+    S.compute_allocation[A.canister_id],
+    S.memory_allocation[A.canister_id],
+    S.freezing_threshold[A.canister_id],
+    memory_usage_wasm_state(S.canisters[A.canister_id]) +
```
mraszyk, line 5178:
```suggestion
    memory_usage_wasm_state(S.canisters[A.canister_id].wasm_state) +
```

mraszyk: same everywhere else

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+      memory_usage_raw_module(A.wasm_module) +
```
mraszyk, line 5179:
```suggestion
      memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
```

mraszyk: same everywhere else

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'delete_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[S.snapshots[A.snapshot_id].canister_id]
+S.snapshots[A.snapshot_id] = (deleted)
```
mraszyk, line 5251: Here we need a canister ID instead of snapshot ID since keys of the map are canister IDs.

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+A snapshot may be deleted only by the controllers of the canister for which the snapshot was taken.
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'delete_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[S.snapshots[A.snapshot_id].canister_id]
```
mraszyk, line 5250:
```suggestion
M.caller ∈ S.controllers[A.canister_id]
```
note that a snapshot does not store a canister ID according to the definition of `Snapshot = {` above.

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+```html
+
+S with
+    canisters[A.canister_id] = {
+      wasm_state = S.snapshots[A.canister_id].wasm_state;
+      raw_module = S.snapshots[A.canister_id].raw_module;
+      module = S.snapshots[A.canister_id].wasm_module;
+      chunk_store = S.snapshots[A.canister_id].chunk_store;
+      public_custom_sections = parse_public_custom_sections(S.snapshots[A.canister_id].wasm_module);
+      private_custom_sections = parse_private_custom_sections(S.snapshots[A.canister_id].wasm_module);
```
mraszyk, line 5199:
```suggestion
      public_custom_sections = parse_public_custom_sections(S.snapshots[A.canister_id].raw_module);
      private_custom_sections = parse_private_custom_sections(S.snapshots[A.canister_id].raw_module);
```

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+
+State after  
+
+```html
+
+S with
+    canisters[A.canister_id] = {
+      wasm_state = S.snapshots[A.canister_id].wasm_state;
+      raw_module = S.snapshots[A.canister_id].raw_module;
+      module = S.snapshots[A.canister_id].wasm_module;
```
mraszyk, line 5196: since we're already parsing the custom section, we could parse the wasm module from the raw module and drop wasm_module from `Snapshot = {`: module = parse_wasm_mod(S.snapshots[A.canister_id].raw_module)

spec/_attachments/ic.did
```diff
@@ -115,6 +115,13 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type snapshot_id = nat;
+type snapshot = record {
+  id: snapshot_id;
+  taken_at_timestamp: nat64;
```
mraszyk, line 121: This field seems never used - is it supposed to be read by developers when listing snapshots?

mraszyk: although they could also note down the time when they took the snapshot themselves

spec/index.md
```diff
@@ -4942,7 +5003,7 @@ if A.settings.reserved_cycles_limit is not null:
 else:
   New_reserved_balance_limit = 5_000_000_000_000
 
-Cycles_reserved = cycles_to_reserve(S, Canister_id, New_compute_allocation, New_memory_allocation, EmptyCanister.wasm_state)
+Cycles_reserved = cycles_to_reserve(S, Canister_id, New_compute_allocation, New_memory_allocation,  S.snapshots[A.canister_id], EmptyCanister.wasm_state)
```
mraszyk, line 5006: the canister has just been created so there's no snapshot yet, we should use `null` here and also define `snapshots[Canister_id] = null` in the "State after" section.

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+```html
+
+S with
+    snapshots[A.canister_id] = New_snapshot
+    balances[A.canister_id] = New_balance
+    reserved_balances[A.canister_id] = New_reserved_balance
+    messages = Older_messages · Younger_messages ·
+      ResponseMessage {
+        origin = M.origin;
+        response = Reply (candid(SnapshotId));
```
mraszyk, line 5149: Could we maybe leave a comment that `SnapshotId` is an opaque nat generated by the system?

It is also not stored anywhere, so listing snapshots by their IDs is impossible.

spec/index.md
```diff
@@ -5033,6 +5094,179 @@ S with
...
+
+State after
+
+```html
+
+S with
+    messages = Older_messages · Younger_messages ·
+      ResponseMessage {
+        origin = M.origin
+        response = Reply( candid(dom(S.snapshots[A.canister_id])))
```
mraszyk, line 5234: Using `dom(...)` wouldn't work here since you define `snapshots` to be a single snapshot (not a map of snapshot IDs to their corresponding snapshots).

spec/index.md
AlexandraZapuc: Thanks for the thorough review. I think we need to go through the semantics for the other (already) defined management methods. Based on your comments,  I am updating the other methods(in this PR), but I think there are more inconsistencies.

spec/index.md
AlexandraZapuc: Yes, it will not be. Will fix this.

spec/index.md
AlexandraZapuc: Ok, I was thinking that this could be interpreted as an empty list. But I will use null to make it clearer.

spec/_attachments/ic.did
AlexandraZapuc: > This field seems never used - is it supposed to be read by developers when listing snapshots? 
Yes, that is the case. This field, I think, will be especially useful when we allow more snapshots per canister. I will leave this field as unless there is a strong preference to remove it.

spec/_attachments/ic.did
```diff
@@ -224,6 +231,21 @@ service ic : {
   bitcoin_send_transaction: (send_transaction_request) -> ();
   bitcoin_get_current_fee_percentiles: (get_current_fee_percentiles_request) -> (vec millisatoshi_per_byte);
 
+  // Canister backup and restore
```
ielashi, line 234: super nitpick: we're not using the term "backup" and "restore" in the API. I'd try to not introduce new vocabulary in the documentation to prevent confusion.

```suggestion
  // Canister snapshots
```

spec/_attachments/ic.did
```diff
@@ -224,6 +231,21 @@ service ic : {
...
+  take_snapshot: (record {
```
ielashi, line 235: Would it make sense to call these "canister snapshots" to make it clear that it's snapshotting a canister? Maybe in the future we'll take snapshots of other things (e.g. the whole subnet, or a canister group). @mraszyk WDYT?

```suggestion
  take_canister_snapshot: (record { 
```

AlexandraZapuc: I think so, I can add `canister` to all methods.

mraszyk: Sounds good to me.

spec/index.md
```diff
@@ -5162,26 +5167,59 @@ Only the controllers of the given canister can load a snapshot.
...
+S.canister_history[A.canister_id] = {
+  total_num_changes = N;
+  recent_changes = H;
+}
+New_canister_history = {
+  total_num_changes = N + 1;
+  recent_changes = H · {
+    timestamp_nanos = S.time[A.canister_id];
+    canister_version = S.canister_version[A.canister_id] + 1
+    origin = change_origin(M.caller, null, M.origin);
```
AlexandraZapuc, line 5196: `change_origin` usually has a sender_canister_version. I set parameter as null, is this correct?

mraszyk: The sender_canister_version is null for ingress messages and for inter-canister calls, it is parsed from the candid argument - please see other mgmt canister methods updating history for how it's done there.

AlexandraZapuc: Yes, I have checked the other methods. Basically, I wanted to ask if I should add it to the candid and pass it to this function after parsing it?

spec/index.md
```diff
@@ -2340,6 +2340,50 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+:::note
+
+It's important to stop a canister before taking a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored using the snapshot.
+It is expected that the canister controllers (or their tooling) do this separately.
+
+:::
+
+### IC method `load_canister_snapshot` {#ic-load_canister_snapshot}
+
+This method loads a snapshot identified by `snapshot_id` onto the canister. It fails if no snapshot with the specified `snapshot_id` was created or if the snapshot was already deleted.
```
mraszyk, line 2362:
```suggestion
This method loads a snapshot identified by `snapshot_id` onto the canister. It fails if no snapshot with the specified `snapshot_id` can be found. 

A snapshot cannot be found if it was never created, it was previously deleted, replaced by a new snapshot through a `take_canister_snapshot` request, or if the canister itself has been deleted or run out of cycles.```

File: spec/index.md [hunk 67]
```diff
@@ -2940,6 +2985,9 @@ Finally, we can describe the state of the IC as a record having the following fi
           mode : CodeDeploymentMode;
           module_hash : Blob;
         }
+      | LoadSnapshot {
```
mraszyk, line 2988: Please also add this variant to spec/_attachments/ic.did

spec/index.md
```diff
@@ -2940,6 +2985,9 @@ Finally, we can describe the state of the IC as a record having the following fi
...
+          module_hash : Blob;
```
mraszyk, line 2989: should we maybe also include the timestamp at which the snapshot was taken?

spec/index.md
```diff
@@ -5033,6 +5101,215 @@ S with
...
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'delete_canister_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[A.canister_id]
+S.snapshots[A.canister_id] = (deleted)
```
mraszyk, line 5294: this belongs to State after; instead we should check here that the snapshot stored for the canister has a matching snapshot ID according to A.snapshot_id

spec/index.md
```diff
@@ -5033,6 +5101,215 @@ S with
...
+  recent_changes = H;
+}
+New_canister_history = {
+  total_num_changes = N + 1;
+  recent_changes = H · {
+    timestamp_nanos = S.time[A.canister_id];
+    canister_version = S.canister_version[A.canister_id] + 1
+    origin = change_origin(M.caller, A.sender_canister_version, M.origin);
+    details = LoadSnapshot {
+      module_hash = SHA-256(New_state.wasm_module);
```
mraszyk, line 5201:
```suggestion
      module_hash = SHA-256(New_state.raw_module);
```

spec/index.md
```diff
@@ -5033,6 +5101,215 @@ S with
...
+
+S with
+    canisters[A.canister_id] = {
+      wasm_state = S.snapshots[A.canister_id].wasm_state;
+      raw_module = S.snapshots[A.canister_id].raw_module;
+      module = parse_wasm_mod(S.snapshots[A.canister_id].raw_module;
+      chunk_store = S.snapshots[A.canister_id].chunk_store;
+      public_custom_sections = parse_public_custom_sections(S.snapshots[A.canister_id].raw_module);
+      private_custom_sections = parse_private_custom_sections(S.snapshots[A.canister_id].raw_module);
+    }
```
mraszyk, line 5240:
```suggestion
    canisters[A.canister_id] = New_state
```

spec/index.md
```diff
@@ -5033,6 +5101,215 @@ S with
...
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'load_canister_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[A.canister_id]
+
```
mraszyk, line 5176: We should check here that the snapshot stored for the canister has a matching snapshot ID according to A.snapshot_id

spec/index.md
```diff
@@ -5033,6 +5101,215 @@ S with
...
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'take_canister_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[A.canister_id]
+
```
mraszyk, line 5116: We should check here that no snapshot is stored for a canister or A.replace_snapshot is set to the snapshot ID of the currently stored snapshot.

spec/index.md
```diff
@@ -2957,9 +3005,16 @@ Finally, we can describe the state of the IC as a record having the following fi
       subnet_id : Principal;
       subnet_size : Nat;
     }
+    Snapshot = {
```
mraszyk, line 3008: maybe we should store the system-generated snapshot ID here so that we can check it later

mraszyk:
Could you please also update canister history sections in the spec mentioning that loading snapshots is also recorded?

spec/index.md
```diff
@@ -2679,6 +2723,7 @@ The [WebAssembly System API](#system-api) is relatively low-level, and some of i
           memory_allocation : Nat;
           memory_usage_raw_module : Nat;
           memory_usage_canister_history : Nat;
+          memory_usage_snapshots: Nat;
```
mraszyk, line 2726: this field does not seem to be initialized

spec/index.md
```diff
@@ -2957,9 +3005,16 @@ Finally, we can describe the state of the IC as a record having the following fi
...
+    Snapshot = {
+      wasm_state : WasmState;
+      raw_module : Blob;
+      chunk_store : ChunkStore;
+      certified_data: Blob;
+    }
     S = {
       requests : Request ↦ (RequestStatus, Principal);
       canisters : CanisterId ↦ CanState;
+      snapshots: SnapshotId ↦ Snapshot;
```
mraszyk, line 3017:
```suggestion
      snapshots: CanisterId ↦ Snapshot;
```

spec/index.md
```diff
@@ -5033,6 +5101,215 @@ S with
...
+      memory_usage_canister_history(New_canister_history) +
+      memory_usage_snapshot(S.snapshots[A.canister_id]),
+    S.canister_subnet[A.canister_id].subnet_size,
+  )
+) ≥ 0
+
+if S.memory_allocation[A.canister_id] > 0:
+  memory_usage_wasm_state(New_state.wasm_state) +
+    memory_usage_raw_module(New_state.raw_module) +
+    memory_usage_canister_history(New_canister_history) ≤ S.memory_allocation[A.canister_id]
```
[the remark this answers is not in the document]
AlexandraZapuc: Discussed offline.

spec/_attachments/ic.did
```diff
@@ -36,6 +36,10 @@ type change_details = variant {
     mode : variant {install; reinstall; upgrade};
     module_hash : blob;
   };
+  load_snapshot : record {
+    module_hash : blob;
```
mraszyk, line 40: We discussed the information stored in load_snapshot in the Interface spec meeting and concluded that the module hash need not be stored since it can be derived from the canister history if we additionally store the canister version of the canister when its snapshot was taken. So our suggestion is to replace the module hash here with the canister version so that a load snapshot record would contain the canister version and timestamp when the snapshot was taken.

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+Only the controllers of the given canister can take a snapshot.
+
+```html
+
+S.messages = Older_messages · CallMessage M · Younger_messages
+(M.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+M.callee = ic_principal
+M.method_name = 'take_canister_snapshot'
+M.arg = candid(A)
+is_system_assigned Snapshot_id    // Snapshot ID is an opaque generated by the system
```
mraszyk, line 5118: The function `is_system_assigned` is defined for principals => let's introduce the variable `Snapshot_id` informally after the sentence "Only the controllers of the given canister can take a snapshot." above.

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+  S.snapshots[A.canister_id] = null
+
+New_snapshot = Snapshot {
+  snapshot_id = Snapshot_id;
+  wasm_state = canisters[A.canister_id].wasm_state;
+  raw_module = canisters[A.canister_id].raw_module;
+  chunk_store = canisters[A.canister_id].chunk_store;
+  certified_data = S.certified_data[A.canister_id];
+  canister_version = S.canister_version[A.canister_id];
+  take_at_timestamp = CurrentTime;
```
mraszyk, line 5132:
```suggestion
  take_at_timestamp = S.time[A.canister_id];
```

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+if A.replace_snapshot is not null:
+  S.snapshots[A.canister_id].snapshot_id = A.replace_snapshot
+else:
+  S.snapshots[A.canister_id] = null
+
+New_snapshot = Snapshot {
+  snapshot_id = Snapshot_id;
+  wasm_state = canisters[A.canister_id].wasm_state;
+  raw_module = canisters[A.canister_id].raw_module;
+  chunk_store = canisters[A.canister_id].chunk_store;
```
mraszyk, line 5129:
```suggestion
  wasm_state = S.canisters[A.canister_id].wasm_state;
  raw_module = S.canisters[A.canister_id].raw_module;
  chunk_store = S.canisters[A.canister_id].chunk_store;
```

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+
+State after
+
+```html
+
+S with
+    messages = Older_messages · Younger_messages ·
+      ResponseMessage {
+        origin = M.origin
+        response = Reply (candid(S.snapshots[A.canister_id]))
```
mraszyk, line 5283:
```suggestion
        response = Reply (candid({
          id = S.snapshots[A.canister_id].snapshot_id;
          taken_at_timestamp = S.snapshots[A.canister_id].taken_at_timestamp;
          total_size = memory_usage_snapshot(S.snapshots[A.canister_id]);
        }))
```

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+M.method_name = 'load_canister_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[A.canister_id]
+S.snapshots[A.canister_id].snapshot_id  = A.snapshot_id
+
+New_state = {
+  wasm_state = S.snapshots[A.canister_id].wasm_state;
+  raw_module = S.snapshots[A.canister_id].raw_module;
+  module = parse_wasm_mod(S.snapshots[A.canister_id].raw_module;
+  chunk_store = S.snapshots[A.canister_id].chunk_store;
```
mraszyk, line 5193:
```suggestion
```
chunk_store is modeled separately from canister state in this spec

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+M.callee = ic_principal
+M.method_name = 'load_canister_snapshot'
+M.arg = candid(A)
+M.caller ∈ S.controllers[A.canister_id]
+S.snapshots[A.canister_id].snapshot_id  = A.snapshot_id
+
+New_state = {
+  wasm_state = S.snapshots[A.canister_id].wasm_state;
+  raw_module = S.snapshots[A.canister_id].raw_module;
+  module = parse_wasm_mod(S.snapshots[A.canister_id].raw_module;
```
mraszyk, line 5192:
```suggestion
  module = parse_wasm_mod(S.snapshots[A.canister_id].raw_module);
```

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+}
+New_canister_history = {
+  total_num_changes = N + 1;
+  recent_changes = H · {
+    timestamp_nanos = S.time[A.canister_id];
+    canister_version = S.canister_version[A.canister_id] + 1
+    origin = change_origin(M.caller, A.sender_canister_version, M.origin);
+    details = LoadSnapshot {
+      canister_version: S.snapshots[A.canister_id].canister_version
+      taken_at_timestamp: S.snapshots[A.canister_id].take_at_timestamp
```
mraszyk, line 5214:
```suggestion
      canister_version = S.snapshots[A.canister_id].canister_version;
      taken_at_timestamp = S.snapshots[A.canister_id].take_at_timestamp;
```

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+    memory_usage_canister_history(New_canister_history) ≤ S.memory_allocation[A.canister_id]
+
+```
+
+State after  
+
+```html
+
+S with
+    canisters[A.canister_id] = New_state
```
mraszyk, line 5246:
```suggestion
    canisters[A.canister_id] = New_state
    chunk_store[A.canister_id] = S.snapshots[A.canister_id].chunk_store
```

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+```html
+
+S with
+    snapshots[A.canister_id] = New_snapshot
+    balances[A.canister_id] = New_balance
+    reserved_balances[A.canister_id] = New_reserved_balance
+    messages = Older_messages · Younger_messages ·
+      ResponseMessage {
+        origin = M.origin;
+        response = Reply (candid(R));
```
mraszyk, line 5167:
```suggestion
        response = Reply (candid({
          id = Snapshot_id;
          taken_at_timestamp = S.time[A.canister_id];
          total_size = memory_usage_snapshot(New_snapshot);
        }))
```

spec/index.md
```diff
@@ -5033,6 +5104,218 @@ S with
...
+    S.memory_allocation[A.canister_id],
+    S.freezing_threshold[A.canister_id],
+    memory_usage_wasm_state(S.canisters[A.canister_id].wasm_state) +
+      memory_usage_raw_module(S.canisters[A.canister_id].raw_module) +
+      memory_usage_canister_history(S.canister_history[A.canister_id]) +
+      memory_usage_snapshot(New_snapshot),
+    S.canister_subnet[A.canister_id].subnet_size,
+  )
+) ≥ 0
+R = <implementation-specific>
```
mraszyk, line 5153:
```suggestion
```
we can actually specify it (see suggestion below)

spec/index.md
```diff
@@ -2470,6 +2470,50 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `take_canister_snapshot` {#ic-take_canister_snapshot}
+
+This method takes a snapshot of the specified canister. A snapshot consists of the wasm memory, stable memory, certified variables, wasm chunk store and wasm binary.
+
+Subsequent `take_canister_snapshot` calls will create a new snapshot. However, a `take_canister_snapshot` call might fail if the maximum number of snapshots per canister is reached. This error can be avoided by providing a snapshot ID via the optional `replace_snapshot` parameter. The snapshot identified by the specified ID will be deleted once a new snapshot has been successfully created. Currently, only one snapshot per canister is allowed.
```
chenyan-dfinity, line 2477: The spec says I can only take one snapshot per canister, but I tried the local replica, and I can take more than 1 snapshots. My local replica is configured as system subnet, not sure if that makes a difference.

spec/index.md
```diff
@@ -2470,6 +2470,50 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the canister's balance must have a sufficient amount of cycles to support the new freezing threshold.
+
+Only controllers can take a snapshot of a canister and load it back to the canister.
+
+:::note
+
+It's important to stop a canister before taking a snapshot to ensure that all outstanding callbacks are completed. Failing to do so may cause the canister to not make sense of the callbacks if its state is restored using the snapshot.
```
chenyan-dfinity, line 2485: Does it make sense for this method to stop the canister, take the snapshot and resume the canister? This way I can save two update calls, and it seems like we always need to do this three steps, right?

spec/index.md
```diff
@@ -2470,6 +2470,50 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+It is expected that the canister controllers (or their tooling) do this separately.
+
+:::
+
+### IC method `load_canister_snapshot` {#ic-load_canister_snapshot}
+
+This method loads a snapshot identified by `snapshot_id` onto the canister. It fails if no snapshot with the specified `snapshot_id` can be found.
+
+Only controllers can take a snapshot of a canister and load it back to the canister.
```
chenyan-dfinity, line 2494: There is a corner case when the canister itself is the controller. Do we want to allow canister itself to load a snapshot? The await callback is almost guaranteed to fail in this case.

spec/index.md
```diff
@@ -2470,6 +2470,50 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+### IC method `list_canister_snapshots` {#ic-list_canister_snapshots}
+
+This method lists the snapshots of the canister identified by `canister_id`. Only controllers of the canister can list its snapshots. Currently, at most one snapshot per canister will be stored.
+
+### IC method `delete_canister_snapshot` {#ic-delete_canister_snapshot}
+
+This method deletes a specified snapshot that belongs to an existing canister. An error will be returned if the snapshot is not found. 
+
+A snapshot cannot be found if it was never created, it was previously deleted, replaced by a new snapshot through a `take_canister_snapshot` request, or if the canister itself has been deleted or run out of cycles.
```
chenyan-dfinity, line 2513: > run out of cycles

Just to double check, if the canister is out of cycles, we lose all the snapshots?

spec/index.md
```diff
@@ -2191,7 +2191,7 @@ This method can only be called by canisters, i.e., it cannot be called by extern
...
+The canister history consists of a list of canister changes (canister creation, code uninstallation, code deployment, snapshot restoration, or controllers change). Every canister change consists of the system timestamp at which the change was performed, the canister version after performing the change, the change's origin (a user or a canister), and its details. The change origin includes the principal (called *originator* in the following) that initiated the change and, if the originator is a canister, the originator's canister version when the originator initiated the change (if available). Code deployments are described by their mode (code install, code reinstall, code upgrade) and the SHA-256 hash of the newly deployed canister module. Loading a snapshot is described by the canister version, snapshot ID and timestamp at which the snapshot was taken. Canister creations and controllers changes are described by the full new set of the canister controllers after the change. The order of controllers stored in the canister history may vary depending on the implementation.
```
dsarlis, line 2194: > Loading a snapshot is described by the canister version, snapshot ID and timestamp at which the snapshot was taken.

This might need to be clarified a bit. The canister version in the history is the version after loading the snapshot but the current phrasing sounds like it might be the old version of the canister when the snapshot was taken.

spec/index.md
dsarlis: Not sure this is true anymore, it depends which replica version you used. The latest should only allow 1 snapshot to be present at a time. You have to call `take_canister_snapshot` with `replace_snapshot_id` to succeed (which will overwrite the previous snapshot).

spec/index.md
dsarlis: This is modelled similarly to `install_code`. Canisters without outgoing calls can safely take a snapshot without stopping (same as they can safely upgrade without stopping). If they do need to stop, then it's up to the canister to do the right thing.

spec/index.md
dsarlis: Same corner case as when the canister is a controller of itself and wants to upgrade itself. (Don't do it is my hot take here).

spec/index.md
dsarlis: Yes.

spec/index.md
dsarlis: Clarified with Alexandra offline. There's a small bug in the replica which I will fix.

spec/index.md
chenyan-dfinity: Double check again, `uninstall_code` keeps the snapshot, only `delete_canister` removes the snapshot?

berestovskyy: The `uninstall_code` method is sort of reseting the canister state. The canister ID is still the same, so it makes sense to keep the snapshots. On the other hand, `delete_canister` method deletes the canister ID, so at the moment there is no need to keep the snapshots any longer.

This behaviors might change in the future though, when the snapshot load across canisters is supported...

mraszyk: > This behaviors might change in the future though, when the snapshot load across canisters is supported...

Could you please elaborate on what exactly might change?

spec/index.md
```diff
@@ -2488,6 +2488,50 @@ A single metric entry is a record with the following fields:
 
 - `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
 
+### IC method `take_canister_snapshot` {#ic-take_canister_snapshot}
+
+This method takes a snapshot of the specified canister. A snapshot consists of the wasm memory, stable memory, certified variables, wasm chunk store and wasm binary.
+
+Subsequent `take_canister_snapshot` calls will create a new snapshot. However, a `take_canister_snapshot` call might fail if the maximum number of snapshots per canister is reached. This error can be avoided by providing a snapshot ID via the optional `replace_snapshot` parameter. The snapshot identified by the specified ID will be deleted once a new snapshot has been successfully created. Currently, only one snapshot per canister is allowed.
+
+It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the canister's balance must have a sufficient amount of cycles to support the new freezing threshold.
```
mraszyk, line 2497:
```suggestion
It's important to note that a snapshot will increase the memory footprint of the canister. Thus, the canister's balance must have a sufficient amount of cycles to not become frozen.
```

spec/index.md
```diff
@@ -5540,6 +5621,230 @@ S with
...
+
+New_state = {
+  wasm_state = S.snapshots[A.canister_id].wasm_state;
+  raw_module = S.snapshots[A.canister_id].raw_module;
+  module = parse_wasm_mod(S.snapshots[A.canister_id].raw_module);
+  public_custom_sections = parse_public_custom_sections(S.snapshots[A.canister_id].raw_module);
+  private_custom_sections = parse_private_custom_sections(S.snapshots[A.canister_id].raw_module);
+}
+Cycles_reserved = cycles_to_reserve(S, A.canister_id, S.compute_allocation[A.canister_id], S.memory_allocation[A.canister_id], S.snapshots[A.canister_id], New_state)
+New_balance = S.balances[A.canister_id] - Cycles_used - Cycles_reserved
```
mraszyk, line 5718: > Cycles_used

Does loading snapshot cost cycles?

spec/index.md
```diff
@@ -5540,6 +5621,230 @@ S with
...
+
+```html
+
+S with
+    canisters[A.canister_id] = New_state
+    chunk_store[A.canister_id] = S.snapshots[A.canister_id].chunk_store
+    certified_data[A.canister_id] = S.snapshots[A.canister_id].certified_data
+    balances[A.canister_id] = New_balance
+    reserved_balances[Canister_id] = New_reserved_balance
+    canister_history[Canister_id] = New_canister_history
```
mraszyk, line 5773:
```suggestion
    reserved_balances[A.canister_id] = New_reserved_balance
    canister_history[A.canister_id] = New_canister_history
    canister_version[A.canister_id] = S.canister_version[A.canister_id] + 1
```

spec/index.md
berestovskyy: It's not even scoped yet. However, sharing snapshots across canister might require some changes in snapshot ownership...

Dfinity-Bjoern (approved):
As this is now available in production, we should merge this to the spec as well.

PR #262, opened 2023-12-13
Update the changelog with the certificate delegation depth limit.

oggy-dfin:


spec/_attachments/interface-spec-changelog.md
```diff
@@ -1,6 +1,7 @@
 ## Changelog {#changelog}
 
 ### ∞ (unreleased)
+- Set the maximum depth of a delegation in a read_state response/certified variable certificate to 1.
```
mraszyk, line 4: I always append the new entry to the bottom of the list. Could you please move it so that we're consistent?

oggy-dfin: Done

mraszyk (approved):
Thank you! @oggy-dfin

PR #263, opened 2023-12-14
[IC-272] canister logging

maksymar:


spec/_attachments/ic.did
```diff
@@ -121,6 +128,16 @@ type node_metrics = record {
     num_block_failures_total : nat64;
 };
 
+type log_record = record {
+  idx: nat;
+  timestamp: nat;
```
sesi200, line 133: Inconsistent with the ICRC standards. The standards use `nat64`, see [here](https://github.com/dfinity/ICRC-1/blob/main/standards/ICRC-1/ICRC-1.did#L2) for example.

spec/_attachments/ic.did
```diff
@@ -242,4 +259,8 @@ service ic : {
   }) -> (record {canister_id : canister_id});
   provisional_top_up_canister :
     (record { canister_id: canister_id; amount: nat }) -> ();
+
+  // canister loging
```
sesi200, line 263:
```suggestion
  // canister logging
```

spec/index.md
```diff
@@ -2340,6 +2340,30 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+### IC method `fetch_logs_query` {#ic-fetch_logs_query}
+
+This method is identical to [`fetch_logs`](#ic-fetch_logs), but exposed as a query.
+
+:::note
+
+This query is only accessible in non-replicated mode. Calls in replicated mode are rejected.
+
+:::
```
sesi200, line 2359: Can you explain the rationale behind this? Why wouldn't it be ok to run this in replicated mode instead of duplicating functionality in an update call?

spec/_attachments/ic.did
```diff
@@ -121,6 +128,16 @@ type node_metrics = record {
     num_block_failures_total : nat64;
 };
 
+type log_record = record {
+  idx: nat;
```
mraszyk, line 132: given the comment on timestamps, we could also consider using nat64 here for consistency

spec/_attachments/ic.did
```diff
@@ -242,4 +259,8 @@ service ic : {
   }) -> (record {canister_id : canister_id});
   provisional_top_up_canister :
     (record { canister_id: canister_id; amount: nat }) -> ();
+
+  // canister loging
+  fetch_logs : (record { canister_id }) -> (fetch_logs_response);
+  fetch_logs_query : (record { canister_id }) -> (fetch_logs_response) query;
```
mraszyk, line 265:
```suggestion
  fetch_logs : (record { canister_id: canister_id }) -> (fetch_logs_response);
  fetch_logs_query : (record { canister_id: canister_id }) -> (fetch_logs_response) query;
```

spec/index.md
```diff
@@ -5033,6 +5057,10 @@ S with
 
 ```
 
+#### IC Management Canister: Canister logs
+
+TBD
```
mraszyk, line 5062: also changes to do the abstract IC state and system API specifications are missing yet (and thus TBD)

spec/index.md
```diff
@@ -2340,6 +2340,30 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+:::note
+
+This query is only accessible in non-replicated mode. Calls in replicated mode are rejected.
+
+:::
+
+:::warning
+
+The response of a query comes from a single replica, and is therefore not appropriate for security-sensitive applications.
```
mraszyk, line 2363: I wonder if we should mention signed-queries here: if you repeat the query and get responses from sufficiently many distinct nodes, then you can actually trust the response.

spec/index.md
mraszyk: which speaks against having the `fetch_logs` method defined above as an update call whatsoever

spec/index.md
```diff
@@ -2340,6 +2340,30 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `fetch_logs` {#ic-fetch_logs}
+
+Provides the logs of the canister including trap messages.
+Logs are stored in a circular buffer of 4KiB.
```
mraszyk, line 2346: Do the `4KiB` also include storage overhead of the underlying data structure? Ideally that's not the case and we can say that the total length of all log contents is at most `4KB`. Then we should drop also the reference to the "circular buffer" and instead say that oldest logs get dropped if appending new logs would exceed the log memory capacity.

spec/index.md
```diff
@@ -2340,6 +2340,30 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `fetch_logs` {#ic-fetch_logs}
+
+Provides the logs of the canister including trap messages.
```
mraszyk, line 2345: Could you please describe the semantics of the individual fields in the response to this mgmt canister call? Please refer to `node_metrics_history` for an example.

mraszyk, line 2345: > including trap messages

if these are automatically logged by the system, then devs should have a way to silence them so that they don't need to pay for the log memory consumed by those logs

mraszyk (commented):
@maksymar Thank you for kicking off the work on the specification of canister logging. I've left some comments on your PR.

spec/_attachments/ic.did
ielashi: FYI the above is valid candid syntax

ielashi, line 265: Why do we need separate endpoints for the query and non-query? It's sufficient to have the single query endpoint, since all queries can be called in replicated mode.

spec/index.md
ielashi: I think we can remove the replication version of `fetch_logs` and only start with the non-replicated version. We can always introduce the replicated mode later.

spec/_attachments/ic.did
mraszyk: I see! Then we can leave it as is if you prefer.

spec/index.md
mraszyk: > We can always introduce the replicated version later.

i.e., do not reject replicated calls, introducing a separate method sounds weird;

I'd also drop the suffix _query

spec/_attachments/ic.did
maksymar: Fixed, switched to `nat64` and also changed the name to `timestamp_nanos` to be consistent with the other examples in this file.

spec/_attachments/ic.did
maksymar: Done.

spec/_attachments/ic.did
maksymar: Done.

spec/_attachments/ic.did
maksymar: To keep the same style I switched to `canister_id : canister_id`. It can be prettified later on for all the instances at once.

maksymar: I added two endpoints looking at the bitcoin endpoints as an example:

```
  bitcoin_get_balance: (get_balance_request) -> (satoshi);
  bitcoin_get_balance_query: (get_balance_request) -> (satoshi) query;
  bitcoin_get_utxos: (get_utxos_request) -> (get_utxos_response);
  bitcoin_get_utxos_query: (get_utxos_request) -> (get_utxos_response) query;
```

spec/index.md
maksymar: do you mean I should drop the replicated endpoint `fetch_logs` and rename the non-replicated by dropping the `_query` suffix?

what if we later decide to add a replicated endpoint, would it be possible to have two endpoints with the same name?

spec/index.md
maksymar: Done.

spec/index.md
ielashi: The same endpoint can be used for replicated and non-replicated executions, just like regular canister queries.

spec/_attachments/ic.did
```diff
@@ -1,12 +1,18 @@
 type canister_id = principal;
 type wasm_module = blob;
 
+type log_visibility = variant {
```
oggy-dfin, line 4: How about the case where the canister would store the logs on the IC? In that case, it might make sense to have the logs accessible by a separate canister that only indexes them, i.e., adding a third option, a (bounded) list of principals who can access the logs.

I guess the most flexible option would be to expose reading the logs as a system API instruction instead of a management call, and then have the canister handle its own access control.

spec/index.md
```diff
@@ -2340,6 +2340,36 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `fetch_logs` {#ic-fetch_logs}
+
+Given a canister ID as input provides the logs of that canister including trap messages.
+Logs are stored in a circular buffer of 4KiB.
+Log memory persists across upgrades and it is deleted if the canister is reinstalled or uninstalled.
+The log visibility is defined in `log_visibility` field of `canister_settings`.
+Users cannot call this method.
```
oggy-dfin, line 2349: The last line means that only canisters can call this method? In that case, is every canister allowed to call this method on its own canister_id?

spec/index.md
oggy-dfin: Long-term, what do we envision in terms of charging for this endpoint (e.g., once we have charging for queries in place)? Would it be free, or charged to the canister?

spec/_attachments/ic.did
mraszyk: >  a (bounded) list of principals who can access the logs.

that used to be part of the initial design, but then it was dropped to keep it simple

spec/index.md
mraszyk: > The last line means that only canisters can call this method?

Yes.

> In that case, is every canister allowed to call this method on its own canister_id?

Only if the visibility is public in the current spec.

spec/index.md
mraszyk: > do you mean I should drop the replicated endpoint fetch_logs and rename the non-replicated by dropping the _query suffix?

yes

spec/index.md
maksymar: Done.

spec/_attachments/ic.did
maksymar: Ack. I'll resolve this conversation since there's no action points.

spec/index.md
maksymar: Done.

spec/index.md
maksymar: I've added a replica-signed query comment to this particular warning, because there's no specific part in the spec describing it. and describing replica-signed queries is not in the scope of this PR.

spec/index.md
maksymar: I suppose the first implementation does not have any means to control the logs verbosity to make it simple.
additionally the log size of 4KB is small enough not to care about the price. which can be (potentially) be even free.

spec/index.md
maksymar: Removed 'no user calls' restriciton.

spec/index.md
mraszyk: > the log size of 4KB is small enough

agreed, this is the order of magnitude taken by the canister history as well (which cannot be turned off either)

spec/index.md
```diff
@@ -2343,30 +2343,28 @@ The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Perc
...
 
-### IC method `fetch_logs_query` {#ic-fetch_logs_query}
-
-This method is identical to [`fetch_logs`](#ic-fetch_logs), but exposed as a query.
-
 :::note
 
-This query is only accessible in non-replicated mode. Calls in replicated mode are rejected.
+This method is exposed as a query and is only accessible in non-replicated mode. 
+Calls in replicated mode are rejected.
```
mraszyk, line 2360: If this is (supposed to be) equivalent to "Users cannot call this method.", could you please also update the other occurrence of this phrase?

spec/index.md
maksymar: I added the abstract IC state description.
but I'm not sure what other system API specifications are missing, could you please clarify?

spec/index.md
mraszyk: in the canister_info section

maksymar: 1) `Calls in replicated mode are rejected` is not an equivalent to `Users cannot call this method`, because we want users to call this method, if fact we want ONLY users to call this method (not canisters). and since canisters can make only update calls to other canisters we want `calls in replicated mode are rejected`.

2) I don't feel comfortable modifying specs of other methods in this PR, just not to mess around with the PR scope.

mraszyk: > Calls in replicated mode are rejected is not an equivalent to Users cannot call this method

My bad, I'm sorry!

spec/_attachments/ic.did
```diff
@@ -121,6 +128,16 @@ type node_metrics = record {
     num_block_failures_total : nat64;
 };
 
+type log_record = record {
+  idx: nat64;
+  timestamp_nanos: nat64;
+  contents: blob;
```
mraszyk, line 134: Should the field name be in plural?

spec/_attachments/ic.did
```diff
@@ -241,5 +258,8 @@ service ic : {
     sender_canister_version : opt nat64;
   }) -> (record {canister_id : canister_id});
   provisional_top_up_canister :
-    (record { canister_id: canister_id; amount: nat }) -> ();
+    (record { canister_id : canister_id; amount: nat }) -> ();
+
+  // canister logging
+  fetch_logs : (record { canister_id : canister_id }) -> (fetch_logs_response) query;
```
mraszyk, line 264: Just to double check: are we fine with the fact that the query results are not certified?

spec/index.md
```diff
@@ -2340,6 +2340,34 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+A single log entry is a record with the following fields:
+
+- `idx` (`nat64`): the unique incremental index of an entry;
+- `timestamp_nanos` (`nat64`): the timestamp as nanoseconds since 1970-01-01 at which the entry was recorded;
+- `contents` (`blob`): the content of a log entry;
+
+:::note
+
+This method is exposed as a query and is only accessible in non-replicated mode.
```
mraszyk, line 2359: > is only accessible in non-replicated mode

does it mean that composite queries can invoke `fetch_logs` as a nested call?

mraszyk:
Will `ic00.canister_status` return log visibility of the canister?

spec/index.md
```diff
@@ -5348,8 +5416,70 @@ S with
...
+A.canister_id = effective_canister_id
+S[A.canister_id].log_visibility = Public or M.caller ∈ S.controllers[A.canister_id]
+
+```
+
+Query response `R`:
+
+```html
+
+{status: "replied"; reply: {arg: candid(S.canister_logs[A.canister_id])}, signatures: Sigs}
```
mraszyk, line 5467: Will we include node signatures?

spec/index.md
```diff
@@ -4994,6 +5059,8 @@ S with
     reserved_balance_limits[Canister_id] = New_reserved_balance_limit
     certified_data[Canister_id] = ""
     canister_history[Canister_id] = New_canister_history
+    canister_log_visibility[Canister_id] = Public
```
mraszyk, line 5062: Is Public gonna be the default log visibility?

spec/index.md
```diff
@@ -3639,6 +3691,8 @@ Depending on whether this is a call message and a response messages, we have eit
 
 The cycle consumption of executing this message is modeled via the unspecified `cycles_used` variable; the variable takes some value between 0 and `MAX_CYCLES_PER_MESSAGE`/`MAX_CYCLES_PER_RESPONSE` (for call execution and response execution, respectively).
 
+The logs produced by the canister during message execution are modeled via the unspecified `canister_logs` variable; the variable stores a list of logs (each of type `CanisterLog`) with consecutive sequence numbers, timestamps equal to `S.time[M.receiver]`, and contents produced by the canister calling `ic0.debug_print`, `ic0.trap`, or produced by the WebAssembly runtime when the canister WebAssembly module traps.
```
mraszyk, line 3694: > contents produced by the canister calling `ic0.debug_print`, `ic0.trap`, or produced by the WebAssembly runtime when the canister WebAssembly module traps.

is this list of events producing logs correct?

spec/index.md
```diff
@@ -844,6 +844,14 @@ It must be contained in the canister ranges of a subnet, otherwise the correspon
 
     -   Otherwise, the call is rejected by the system independently of the effective canister id.
 
+-   If the request is a query call to the Management Canister (`aaaaa-aa`), then:
+
+    -   If the call is to the `bitcoin_get_balance_query` or `bitcoin_get_utxos_query` method, then the effective canister id for this call must be the Management Canister (`aaaaa-aa`).
```
mraszyk, line 849: @r-birkner Do the BN expect and enforce the condition on bitcoin query calls?

spec/index.md
```diff
@@ -2053,6 +2061,8 @@ Indicates various information about the canister. It contains:
 
 -   The cycle balance of the canister.
 
+-   The canister log visibility of the canister.
```
mraszyk, line 2064: You  specified the canister log visibility as part of `definite_canister_settings` so it'll be returned here.

spec/index.md
mraszyk: I've refined your initial spec PR so let's continue the discussion on the detailed questions I raised below.

spec/index.md
mraszyk: From the spec perspective, I'd go for not allowing calls to `fetch_logs` as nested calls in composite query evaluation.

mraszyk:
I suppose that logs produced during `install_code` will also be recorded, right?

spec/index.md
mraszyk: I think so since the HTTP handler attaches them for all requests to the query endpoint, but I'd like to double check.

spec/_attachments/ic.did
maksymar: Renamed to singular.

spec/index.md
maksymar: Ack.

spec/index.md
maksymar: Added "only users can call, not canisters, no nested composite queries".

spec/index.md
maksymar: I suppose it's safer to go with controllers by default.

spec/index.md
```diff
@@ -5064,7 +5065,7 @@ S with
     reserved_balance_limits[Canister_id] = New_reserved_balance_limit
     certified_data[Canister_id] = ""
     canister_history[Canister_id] = New_canister_history
-    canister_log_visibility[Canister_id] = Public
+    canister_log_visibility[Canister_id] = New_controllers
```
mraszyk, line 5068: The default to Public is set at multiple places - could you please update them all?

spec/index.md
maksymar: @ielashi PTAL

spec/index.md
maksymar: Done. Also I was not sure if order is important, but just in case I switched to `Controllers | Public`.

spec/index.md
```diff
@@ -5462,7 +5462,7 @@ Q.canister_id = ic_principal
 Q.method_name = 'fetch_logs'
 Q.arg = candid(A)
 A.canister_id = effective_canister_id
-S[A.canister_id].canister_log_visibility = Public or M.caller ∈ S.controllers[A.canister_id]
+S[A.canister_id].canister_log_visibility = M.caller ∈ S.controllers[A.canister_id] or Public
```
mraszyk, line 5465: why did you change this line? it looks weird now?

spec/index.md
```diff
@@ -3899,7 +3899,7 @@ S with
     if A.settings.log_visibility is not null:
       canister_log_visibility[Canister_id] = A.settings.log_visibility
     else:
-      canister_log_visibility[Canister_id] = Public
+      canister_log_visibility[Canister_id] = New_controllers
```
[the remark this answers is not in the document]
maksymar: Good catch, thank you. Fixed all the occurrences from list to enum variant.

spec/index.md
maksymar: first, I thought that oder might matter (I was not sure if it was) and the most left value should be considered as the default one, that's why I moved to have controllers first and public the second.

and secondly I realised that `log_visibility` is an enum, not the actual controllers list so changed to `Controllers or Public`.

spec/_attachments/ic.did
ielashi: Yes, I think we're fine with that.

spec/index.md
ielashi: Yes, that is correct.

spec/_attachments/ic.did
ielashi, line 264:
```suggestion
  fetch_canister_logs : (record { canister_id : canister_id }) -> (fetch_logs_response) query;
```

Would it make sense to update the method name to make it clear that it's fetching the logs of a canister? Maybe in the future we want to add a way to fetch logs for the replicas, for instance.

maksymar: Done.

spec/index.md
```diff
@@ -5011,6 +5082,8 @@ S with
     reserved_balance_limits[Canister_id] = New_reserved_balance_limit
     certified_data[Canister_id] = ""
     canister_history[Canister_id] = New_canister_history
+    canister_log_visibility[Canister_id] = Controllers
```
mraszyk, line 5085: Here we should take the log visibility from the canister settings if provided.

spec/index.md
```diff
@@ -5365,8 +5439,70 @@ S with
...
+
+Conditions
+
+```html
+
+Q.canister_id = ic_principal
+Q.method_name = 'fetch_canister_logs'
+Q.arg = candid(A)
+A.canister_id = effective_canister_id
+S[A.canister_id].canister_log_visibility = Controllers or Public
```
mraszyk, line 5482:
```suggestion
S[A.canister_id].canister_log_visibility = Public or Q.sender in S[A.canister_id].controllers
```

spec/index.md
maksymar: Done.

spec/index.md
maksymar: `canister_log_visibility` is an enum, it should not contain the actual list of controllers, that's why I provided `Controllers or Public`.

or am I missing something? why did you add `Q.sender in S[A.canister_id].controllers`?

spec/index.md
```diff
@@ -5082,7 +5082,7 @@ S with
     reserved_balance_limits[Canister_id] = New_reserved_balance_limit
     certified_data[Canister_id] = ""
     canister_history[Canister_id] = New_canister_history
-    canister_log_visibility[Canister_id] = Controllers
+    canister_log_visibility[Canister_id] = New_canister_log_visibility
```
mraszyk, line 5085: Where is `New_canister_log_visibility` defined?

spec/index.md
mraszyk: The conditions should state that
- either the log visibility is _public_
- or the log visibility is _private_ and the sender of the message is a controller of the canister
Does this make sense?

spec/index.md
maksymar: @ielashi PTAL

spec/index.md
maksymar: Yes it does, thank you!

spec/index.md
maksymar: added `New_canister_log_visibility` definition to canister creation section

mraszyk (commented):
Thank you! LGTM (I'll approve once the feature is ready to be released so that we don't accidentally merge the spec changes earlier.)

maksymar:
> I suppose that logs produced during `install_code` will also be recorded, right?

since this is an update call, then it's one of the requirements from the [design doc](https://docs.google.com/document/d/10zx2qybpK9lmYBUn4r-AbNkcfA_k_zpM7XQ0yfyzBRc/edit):

> Developers are able to log messages in:
> - Heartbeats
> - Timers
> - Pre/post upgrade and canister_init
> - Update calls
> - Queries if called in replicated mode

spec/index.md
r-birkner: Sorry for replying so late. I somehow missed it and Maksym had to remind me to reply.

Just to make sure: 
> If the request is a query call to the Management Canister (`aaaaa-aa`), then:
This means that the canister ID in the request body is set to `aaaaa-aa`?

As far as I can tell, we don't enforce that. The only thing we do is that if it is a query to the management canister for the two specified methods, we rewrite the canister ID to either the testnet or mainnet BTC canister depending on the network argument in the request.

I will double-check with the team to make sure my response is really correct :)

mraszyk: > If the request is a query call to the Management Canister (aaaaa-aa), then:
> This means that the canister ID in the request body is set to aaaaa-aa?

Indeed, this means that the canister_id field in the CBOR map of the request's body is set to aaaaa-aa.

mraszyk: It looks like (see [here](https://github.com/dfinity/ic/blob/d970aa69a9286c34d8cbf34985e3d0c44ca2e9bc/rs/execution_environment/src/query_handler.rs#L209)) the replica expects to see aaaaa-aa as the query's canister_id. So all the BN could do is to route the request to the bitcoin subnet and maybe also update the (effective) canister ID in the URL.

mraszyk: > As far as I can tell, we don't enforce that.

I see. The replica doesn't enforce it either, but I guess I don't have a (security) argument to push for a change here.

spec/index.md
```diff
@@ -2374,6 +2384,35 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `fetch_canister_logs` {#ic-fetch_canister_logs}
+
+Given a canister ID as input, this method returns a vector of logs of that canister including its trap messages.
+The total length of all log contents does not exceed 4KiB.
+If new logs are added resulting in exceeding the maximum total log length of 4KiB, the oldest logs will be removed.
+Logs persist across canister upgrades and they are deleted if the canister is reinstalled or uninstalled.
+The log visibility is defined in the `log_visibility` field of `canister_settings`: logs can be either public (visible to everyone) or only visible to the canister's controllers.
```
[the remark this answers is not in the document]
maksymar: Done.

spec/index.md
```diff
@@ -2374,6 +2384,35 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+A single log is a record with the following fields:
+
+- `idx` (`nat64`): the unique sequence number of the log;
```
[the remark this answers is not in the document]
maksymar: Done.

maksymar:
hi all, the feature is going to be available on the next dfx release, I suppose we can move forward with merging this PR.

spec/index.md
ielashi: @mraszyk Can you point to where in the HTTP handler adds signatures to the queries? And does it matter whether or not it's attached to the `fetch_canister_logs` endpoint?

mraszyk: The signature is computed [here](https://github.com/dfinity/ic/blob/6c35853d3ac8330485068e0b2d8f44c2aa073ca4/rs/http_endpoints/public/src/query.rs#L252). So as long as the query execution service is used, we should be getting query signatures.

spec/index.md
```diff
@@ -2374,6 +2384,35 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `fetch_canister_logs` {#ic-fetch_canister_logs}
+
+Given a canister ID as input, this method returns a vector of logs of that canister including its trap messages.
+The total length of all log contents does not exceed 4KiB.
```
mraszyk, line 2390: @maksymar Could you please confirm that this is the constant actually used in code?

maksymar: yes, here's the constant [MAX_ALLOWED_CANISTER_LOG_BUFFER_SIZE](https://sourcegraph.com/github.com/dfinity/ic@03de843e2a01f57a8fb5c2ef928e91574396f6c5/-/blob/rs/types/management_canister_types/src/lib.rs?L2504) and also preparing an [MR with the test](https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/18416).

Dfinity-Bjoern:
In the Interface Spec meeting, we suggested that we may want to mark this feature as experimental.

Dfinity-Bjoern (approved):
Please merge in sync with release in prod

Stefs-2142:
Really looking forward to this feature!

spec/index.md
```diff
@@ -5536,8 +5623,76 @@ S with
...
+#### Trimming canister logs
+
+Canister logs can be trimmed if the total length of their contents exceeds 4KiB.
+
+Conditions
+
+```html
+
+S.canister_logs[CanisterId] = Older_logs · Newer_logs
+SUM { |l.content| | l <- Older_logs } > 4KiB
```
mraszyk, line 5635: The implementation uses the size of the log entry struct in Rust here instead of the binary size of the content.

mraszyk: I.e., the log entry size in the replica implementation is a constant away from the size specified here.

maksymar: yes, otherwise it can have unlimited number of entries with zero content size.

mraszyk: The relevant function in the replica code is [here](https://github.com/dfinity/ic/blob/8176ea68ae82cb2ba3eec346259e00bf318a171d/rs/types/management_canister_types/src/lib.rs#L2525): it adds the size of the two `u64` (16 bytes in total) on top of the log content size.

mraszyk: > yes, otherwise it can have unlimited number of entries with zero content size.

that makes a lot of sense - let's see how to phrase it in the spec

mraszyk:
Why are we ignoring many cases of `HypervisorError` [here](https://github.com/dfinity/ic/blob/8176ea68ae82cb2ba3eec346259e00bf318a171d/rs/embedders/src/wasm_executor.rs#L771)? I also wonder if a trap log message should describe what `ApiType` trapped (in particular, in light of [RUN-957](https://dfinity.atlassian.net/browse/RUN-957)).

spec/index.md
```diff
@@ -3745,6 +3804,8 @@ Depending on whether this is a call message and a response messages, we have eit
 
 The cycle consumption of executing this message is modeled via the unspecified `cycles_used` variable; the variable takes some value between 0 and `MAX_CYCLES_PER_MESSAGE`/`MAX_CYCLES_PER_RESPONSE` (for call execution and response execution, respectively).
 
+The logs produced by the canister during message execution are modeled via the unspecified `canister_logs` variable; the variable stores a list of logs (each of type `CanisterLog`) with consecutive sequence numbers, timestamps equal to `S.time[M.receiver]`, and contents produced by the canister calling `ic0.debug_print`, `ic0.trap`, or produced by the WebAssembly runtime when the canister WebAssembly module traps.
```
mraszyk, line 3807: Does `ic0.trap` correspond to `HypervisorError::CalledTrap(text)` and "produced by the WebAssembly runtime when the canister WebAssembly module traps" correspond to `HypervisorError::Trapped(trap_code)` in the replica code? What does a panic in unwrapping in Rust canister code correspond to then?

maksymar:
> Why are we ignoring many cases of `HypervisorError` [here](https://github.com/dfinity/ic/blob/8176ea68ae82cb2ba3eec346259e00bf318a171d/rs/embedders/src/wasm_executor.rs#L771)?

because by design doc we only log debug_prints and traps. all the other `HypervisorError`'s are specific to wasm execution, eg. Wasm engine errors, message rejected, invalid wasm, etc.

maksymar:
> I also wonder if a trap log message should describe what `ApiType` trapped (in particular, in light of [RUN-957](https://dfinity.atlassian.net/browse/RUN-957)).

we wanted to keep it simple at the beginning with just adding a `[TRAP]:` prefix. it can be extended later to the method/args/api_type etc.

spec/index.md
maksymar: in my early testing I had different prefixes for `CalledTrap` and `Trapped`, it was `Canister explicitly called trap:` and `Canister trapped:` correspondingly. So panic was `CalledTrap`, see https://github.com/maksymar/canister-logs:

```shell
... Canister explicitly called trap: Panicked at 'aaa!', demo_canister/src/lib.rs:52:5
```

mraszyk: > So panic was CalledTrap

That's interesting - a panic in Rust code results in `ic0.trap`? How does the Rust compiler know about that?

mraszyk: Thank you for the explanation!

spec/index.md
Dfinity-Bjoern: Wouldn't that just be `SUM { |l.content| + 8 | l <- Older_logs } > 4KiB`?

mraszyk: To keep the spec simple, I've decided to refer to canister log size rather than the size of its content and some overhead.

PR #265, opened 2024-01-11
feat(sync-call): [IC-1666] Endpoint for synchronous call requests

DSharifi:
As part of [IC-1666](https://dfinity.atlassian.net/browse/IC-1666) the proposal introduces a new HTTPS endpoint for synchronous update calls, or "call and await", where users can send call requests to the replica and get the response back in the same request. This differs from the current call approach where users have to continuously poll the status of a call request to get the response.

Adding a synchronous endpoint for update calls will significantly decrease the end-to-end (client-observed) latency for update calls. An initial prototype shows that the client-observed latency can be reduced by a minimum of 1 second which is ~40% improvement.

[IC-1666]: https://dfinity.atlassian.net/browse/IC-1666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -536,7 +536,7 @@ The concrete mechanism that users use to send requests to the Internet Computer 
 
 -   At `/api/v2/canister/<effective_canister_id>/call` the user can submit update calls that are asynchronous and might change the IC state.
 
--   At `/api/v2/canister/<effective_canister_id>/sync_call` the user can submit update calls and either get a synchronous HTTPS response with a certified status (if the update call completes within one execution round) or have the update call processed asynchronously.
+-   At `/api/v3/canister/<effective_canister_id>/call` the user can submit update calls and get a synchronous HTTPS response with a certificate for the call status.
 
 -   At `/api/v2/canister/<effective_canister_id>/read_state` or `/api/v2/subnet/<subnet_id>/read_state` the user can read various information about the state of the Internet Computer. In particular, they can poll for the status of a call here.
 
@@ -546,7 +546,7 @@ The concrete mechanism that users use to send requests to the Internet Computer 
 
 In these paths, the `<effective_canister_id>` is the [textual representation](#textual-ids) of the [*effective* canister id](#http-effective-canister-id).
 
-Requests to `/api/v2/canister/<effective_canister_id>/call`, `/api/v2/canister/<effective_canister_id>/sync_call`, `/api/v2/canister/<effective_canister_id>/read_state`, `/api/v2/subnet/<subnet_id>/read_state`, and `/api/v2/canister/<effective_canister_id>/query` are POST requests with a CBOR-encoded request body, which consists of a authentication envelope (as per [Authentication](#authentication)) and request-specific content as described below.
+Requests to `/api/v2/canister/<effective_canister_id>/call`, `/api/v3/canister/<effective_canister_id>/call`, `/api/v2/canister/<effective_canister_id>/read_state`, `/api/v2/subnet/<subnet_id>/read_state`, and `/api/v2/canister/<effective_canister_id>/query` are POST requests with a CBOR-encoded request body, which consists of a authentication envelope (as per [Authentication](#authentication)) and request-specific content as described below.
 
 :::note
 
@@ -559,10 +559,10 @@ This document does not yet explain how to find the location and port of the Inte
 Users interact with the Internet Computer by calling canisters. By the very nature of a blockchain protocol, they cannot be acted upon immediately, but only with a delay. Moreover, the actual node that the user talks to may not be honest or, for other reasons, may fail to get the request on the way.
 
 The Internet Computer has two HTTPS APIs for canister calling:
-- [*Asynchronous*](#http-async-call) canister calling, where the user must poll the Internet Computer for the status of the canister call by _separate_ HTTPS requests.
-- [*Synchronous*](#http-sync-call) canister calling, where the status of the canister call is in the response of the original HTTPS request.
+- [*Asynchronous*](#http-async-call-overview) canister calling, where the user must poll the Internet Computer for the status of the canister call by _separate_ HTTPS requests.
+- [*Synchronous*](#http-sync-call-overview) canister calling, where the status of the canister call is in the response of the original HTTPS request.
 
-#### Asynchronous canister calling {#http-async-call}
+#### Asynchronous canister calling {#http-async-call-overview}
 
 1.  A user submits a call via the [HTTPS Interface](#http-interface). No useful information is returned in the immediate response (as such information cannot be trustworthy anyways).
 
@@ -621,23 +621,15 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
 
 When asking the IC about the state or call of a request, the user uses the request id (see [Request ids](#request-id)) to read the request status (see [Request status](#state-tree-request-status)) from the state tree (see [Request: Read state](#http-read-state)).
 
-#### Synchronous canister calling {#http-sync-call}
+#### Synchronous canister calling {#http-sync-call-overview}
 
-A synchronous update call, also known as a "call and await", is a type of update call where the replica will respond with a certificate if the call completes within one execution round.
+A synchronous update call, also known as a "call and await", is a type of update call where the replica will respond with a certificate if the call completes within a timeout.
 
-The purpose of the synchronous call endpoint is to give the user a certified response for their canister call. This means that if a certificate is returned with a terminal state, i.e. `replied`, `rejected`, or `done`, then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call.
+The purpose of the synchronous call endpoint is to give the user a certified response for their canister call. This means that if a certificate is returned with the status of the update call in a terminal state, i.e. `replied`, `rejected`, or `done`, then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call.
 
 Replicas will reply back to users' submitted update calls with a certificate of the state of the call. A replica will keep the HTTPS connection for the request alive and respond once the state transitions to `replied`, `rejected`, or `done`. If a timeout for the request is reached,  then the replica will reply before the call completes, meaning a certificate with the state in in `unknown`, `received`, or `processing` is returned. In such cases, the user should poll the IC to determine the terminal state of the update call.
 
-A certificate is returned by the IC to the user if:
-
--   The execution of the call completes within one execution round, and the status is certified by the IC within the ingress expiry time. The status will be either `replied`, `rejected` or `done`.
-
--   The call expires before it is queued and is dropped by the IC. The status of the call will be `unknown` with the certificate time stamp at a later time than the ingress expiry time.
-
 The synchronous call endpoint is useful for users as it removes the networking overhead of polling the IC to determine the status of their call.
-
-Synchronous calls are therefore best suited for calls that are fast and expected to complete within one execution round. If a call does not complete within one execution round, the replica won't return a certificate and the user will need to poll the IC to determine the status of the call.
 
 ### Request: Call {#http-call}
 
@@ -2929,7 +2921,7 @@ A reference implementation would likely maintain a separate list of `messages` f
 
 #### API requests
 
-We distinguish between API requests (type `Request`) passed to `/api/v2/…/call` and `/api/v2/…/sync_call`, which may be present in the IC state, and the *read-only* API requests passed to `/api/v2/…/read_state` and `/api/v2/…/query`, which are only ephemeral.
+We distinguish between API requests (type `Request`) passed to `/api/v2/…/call` and `/api/v3/…/call`, which may be present in the IC state, and the *read-only* API requests passed to `/api/v2/…/read_state` and `/api/v2/…/query`, which are only ephemeral.
 
 These are the read-only messages:
 
@@ -3173,7 +3165,7 @@ The following is an incomplete list of invariants that should hold for the abstr
 
 Based on this abstract notion of the state, we can describe the behavior of the IC. There are three classes of behaviors:
 
--   Potentially state changing API requests that are submitted via `/api/v2/…/call` and `/api/v2/…/sync_call`. These transitions describe checks that the request must pass to be considered received.
+-   Potentially state changing API requests that are submitted via `/api/v2/…/call` and `/api/v3/…/call`. These transitions describe checks that the request must pass to be considered received.
 
 -   Spontaneous transitions that model the internal behavior of the IC, by describing conditions on the state that allow the transition to happen, and the state after.
 
@@ -3219,7 +3211,7 @@ A `Request` has an effective canister id according to the rules in [Effective ca
 
 #### API Request submission
 
-After a node accepts a request via `/api/v2/canister/<ECID>/call` or `/api/v2/canister/<ECID>/sync_call`, the request gets added to the IC state as `Received`.
+After a node accepts a request via `/api/v2/canister/<ECID>/call` or `/api/v3/canister/<ECID>/call`, the request gets added to the IC state as `Received`.
 
 This may only happen if the signature is valid and is created with a correct key. Due to this check, the envelope is discarded after this point.
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -643,7 +643,7 @@ When asking the IC about the state or call of a request, the user uses the reque
 
 #### Synchronous canister calling {#http-sync-call-overview}
 
-A synchronous update call, also known as a "call and await", is a type of update call where the replica will attempt to respond with a certificate of the state of the call to the HTTPS request. If the returned certificate indicates that the update call is in a terminal state (`replied`, `rejected`, or `done`), then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call. A terminal state means the call has completed its execution.
+A synchronous update call, also known as a "call and await", is a type of update call where the replica will attempt to respond to the HTTPS request with a certificate of the call status. If the returned certificate indicates that the update call is in a terminal state (`replied`, `rejected`, or `done`), then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call. A terminal state means the call has completed its execution.
 
 The synchronous call endpoint is useful for users as it removes the networking overhead of polling the IC to determine the status of their call.
 
@@ -671,17 +671,17 @@ In order to call a canister, the user makes a POST request to `/api/v3/canister/
 
 -   `arg` (`blob`): Argument to pass to the canister method.
 
-The HTTP response to this request can have the following responses:
+The HTTP response to this request can have the following forms:
 
--   200 HTTP status with a non-empty body. This status is returned if the replica received the message.
+-   200 HTTP status with a non-empty body. This status is returned if the replica successfully received the message.
     
-    -   A certificate for the state of the update call is produced, and returned as a CBOR (see [CBOR](#cbor)) map with the fields below. The user should use the certificate to determine the state of the call:
+    -   A certificate for the state of the update call is produced, and returned in a CBOR (see [CBOR](#cbor)) map with the fields specified below. The user should use the certificate to determine the state of the call:
 
         -   `status` (`text`): `"certified_state"`
 
-        -   `reply` (`blob`):  A certificate (see [Certification](#certification)) with subtrees at `/request_status/<request_id>` and `/time`. See [Request status](#state-tree-request-status) for more details on the request status.
+        -   `reply` (`blob`):  A certificate (see [Certification](#certification)) with subtrees at `/request_status/<request_id>` and `/time`, where `<request_id>` is the [request ID](#request-id) of the update call. See [Request status](#state-tree-request-status) for more details on the request status.
 
-    -   If a non-replicated pre-processing error occurred due to the [canister inspect message](#system-api-inspect-message), then a body with information about the IC specific error encountered is returned. The body is a CBOR map with the following fields:
+    -   If a non-replicated pre-processing error occurred (e.g., due to the [canister inspect message](#system-api-inspect-message)), then a body with information about the IC specific error encountered is returned. The body is a CBOR map with the following fields:
 
         -   `status` (`text`): `"canister_inspect_message_rejected"`
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -683,7 +683,7 @@ The HTTP response to this request can have the following forms:
 
     -   If a non-replicated pre-processing error occurred (e.g., due to the [canister inspect message](#system-api-inspect-message)), then a body with information about the IC specific error encountered is returned. The body is a CBOR map with the following fields:
 
-        -   `status` (`text`): `"canister_inspect_message_rejected"`
+        -   `status` (`text`): `"non_replicated_rejection"`
 
         -   `reject_code` (`nat`): The reject code (see [Reject codes](#reject-codes)).
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -623,7 +623,7 @@ When asking the IC about the state or call of a request, the user uses the reque
 
 #### Synchronous canister calling {#http-sync-call}
 
-Unlike asynchronous canister calling, synchronous canister calling does not require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. Still, the state transitions and semantics of the call itself are the same as for asynchronous canister calling.
+Unlike asynchronous canister calling, synchronous canister calling does not initially require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. However, if a certified response is not received (due to a timeout or pre-execution error), the user must poll the Internet Computer for the status. Still, the state transitions and semantics of the call itself are the same as for asynchronous canister calling.
 
 1.  A user submits a synchronous call via the [HTTPS Interface](#http-interface).
 
@@ -696,10 +696,12 @@ In order to make a synchronous update call to a canister, the user makes a POST 
 
 The HTTP response to this request can have the following responses:
 
--   202 HTTP status with a non-empty body. Implying the request was processed.
+-   201 HTTP status with a non-empty body. Implying the request was processed.
     -   `response` (`blob`): A certificate (see [Certification](#certification)).
 
     The returned certificate includes the subtree at `/request_status/<request_id>` and `/time`.
+
+-   202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use [`read_state`](#http-read-state) to determine the status of the call.
 
 -   200 HTTP status with a non-empty body. Implying an execution pre-processing error occurred. The body of the response contains more information about the IC specific error encountered. The body is a CBOR map with the following fields:
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -623,11 +623,11 @@ When asking the IC about the state or call of a request, the user uses the reque
 
 #### Synchronous canister calling {#http-sync-call-overview}
 
-A synchronous update call, also known as a "call and await", is a type of update call where the replica will attempt to respond with a certificate to the HTTPS request. If the returned certificate indicates the that the update call is in a terminal state, `replied`, `rejected`, or `done`, then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call.
+A synchronous update call, also known as a "call and await", is a type of update call where the replica will attempt to respond with a certificate to the HTTPS request. If the returned certificate indicates that the update call is in a terminal state, `replied`, `rejected`, or `done`, then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call.
 
 The synchronous call endpoint is useful for users as it removes the networking overhead of polling the IC to determine the status of their call.
 
-The replica will keep the HTTPS connection for the request alive and respond once the state transitions to a terminal state. If a timeout for the request is reached while waiting for the terminal state to be reached,  then the replica will reply before the call completes, meaning a certificate with the state in in `unknown`, `received`, or `processing` is returned.
+The replica will maintain the HTTPS connection for the request and will respond once the state transitions to a terminal state. If a timeout for the request is reached while waiting for the terminal state to be reached,  then the replica will reply before the call completes, meaning a certificate with the state in `unknown`, `received`, or `processing` is returned.
 
 The user should take the following actions for the returned certificate states:
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -673,7 +673,7 @@ In order to call a canister, the user makes a POST request to `/api/v3/canister/
 
 The HTTP response to this request can have the following responses:
 
--   200 HTTP status with a non-empty body. This status is returned if the canister call completed or was rejected.
+-   200 HTTP status with a non-empty body. This status is returned if the canister call completed or was rejected within an implementation-specific timeout.
     
     -   If the update call completes, meaning the call state is in `reply`, `reject`, or `done`, then the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -558,27 +558,9 @@ This document does not yet explain how to find the location and port of the Inte
 
 Users interact with the Internet Computer by calling canisters. By the very nature of a blockchain protocol, they cannot be acted upon immediately, but only with a delay. Moreover, the actual node that the user talks to may not be honest or, for other reasons, may fail to get the request on the way.
 
-The Internet Computer therefore has a two models for canister calling:
-- [*Synchronous*](#http-sync-call) canister calling, where the user waits for a certified response from the Internet Computer for the initial call.
+The Internet Computer has two HTTPS APIs for canister calling:
 - [*Asynchronous*](#http-async-call) canister calling, where the user must poll the Internet Computer for the status of the request.
-
-#### Synchronous canister calling {#http-sync-call}
-
-1.  A user submits a synchronous call via the [HTTPS Interface](#http-interface).
-
-2.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for normal calls. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
-
-3.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load).
-
-4.  If the call is processed (sufficient resources, call not yet expired), it will be executed, for some calls this may be atomic, for others this involves multiple internal steps.
-
-5.  Eventually, a response will be produced which will be replied the user. The response can be a `reply`, indicating success, a `reject`, indicating some form of error.
-
-6.  In the case that the call has been retained for long enough before a response is generated, but the request has not expired yet, the IC can forget the response data and only remember the call as `done`, to prevent a replay attack.
-
-7. The user can also afterwards retrieve the state of the request via the [HTTPS Interface](#http-interface) for a certain amount of time.
-
-8.  Once the expiry time is past, the IC can prune the call and its response, and completely forget about it.
+- [*Synchronous*](#http-sync-call) canister calling, where the user waits for a certified response from the Internet Computer for the initial call.
 
 #### Asynchronous canister calling {#http-async-call}
 
@@ -638,6 +620,27 @@ To avoid replay attacks, the transition from `done` or `received` to `pruned` mu
 Calls must stay in `replied` or `rejected` long enough for polling users to catch the response.
 
 When asking the IC about the state or call of a request, the user uses the request id (see [Request ids](#request-id)) to read the request status (see [Request status](#state-tree-request-status)) from the state tree (see [Request: Read state](#http-read-state)).
+
+#### Synchronous canister calling {#http-sync-call}
+
+Unlike asynchronous canister calling, synchronous canister calling does not require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. However, The state transitions and semantics of the call are the same as for asynchronous canister calling.
+
+1.  A user submits a synchronous call via the [HTTPS Interface](#http-interface).
+
+2.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for normal calls. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
+
+3.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load).
+
+4.  If the call is processed (sufficient resources, call not yet expired), it will be executed, for some calls this may be atomic, for others this involves multiple internal steps.
+
+5.  Eventually, a response will be produced which will be replied the user. The response can be a `reply`, indicating success, a `reject`, indicating some form of error.
+
+6.  In the case that the call has been retained for long enough before a response is generated, but the request has not expired yet, the IC can forget the response data and only remember the call as `done`, to prevent a replay attack.
+
+7. The user can also afterwards retrieve the state of the request via the [HTTPS Interface](#http-interface) for a certain amount of time.
+
+8.  Once the expiry time is past, the IC can prune the call and its response, and completely forget about it.
+
 
 ### Request: Call {#http-call}
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -623,13 +623,13 @@ When asking the IC about the state or call of a request, the user uses the reque
 
 #### Synchronous canister calling {#http-sync-call}
 
-Unlike asynchronous canister calling, synchronous canister calling does not require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. However, The state transitions and semantics of the call are the same as for asynchronous canister calling.
+Unlike asynchronous canister calling, synchronous canister calling does not require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. Still, the state transitions and semantics of the call itself are the same as for asynchronous canister calling.
 
 1.  A user submits a synchronous call via the [HTTPS Interface](#http-interface).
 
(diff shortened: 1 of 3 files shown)


spec/index.md
```diff
@@ -613,6 +621,27 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
 
 When asking the IC about the state or call of a request, the user uses the request id (see [Request ids](#request-id)) to read the request status (see [Request status](#state-tree-request-status)) from the state tree (see [Request: Read state](#http-read-state)).
 
+#### Synchronous canister calling {#http-sync-call}
+
+Unlike asynchronous canister calling, synchronous canister calling does not require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. However, The state transitions and semantics of the call are the same as for asynchronous canister calling.
```
mraszyk, line 626:
```suggestion
Unlike asynchronous canister calling, synchronous canister calling does not require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. Still, the state transitions and semantics of the call itself are the same as for asynchronous canister calling.
```

spec/index.md
```diff
@@ -613,6 +621,27 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+1.  A user submits a synchronous call via the [HTTPS Interface](#http-interface).
+
+2.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for normal calls. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
```
mraszyk, line 630:
```suggestion
2.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for calls to regular canisters. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
```

spec/index.md
```diff
@@ -613,6 +621,27 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+3.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load).
```
mraszyk, line 632:
```suggestion
3.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although the IC may still not start processing the request if it is under high load).
```

spec/index.md
```diff
@@ -613,6 +621,27 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
 
 When asking the IC about the state or call of a request, the user uses the request id (see [Request ids](#request-id)) to read the request status (see [Request status](#state-tree-request-status)) from the state tree (see [Request: Read state](#http-read-state)).
 
+#### Synchronous canister calling {#http-sync-call}
```
mraszyk, line 624: I don't get the point of this section - it looks like you copied the first eight steps from the previous section.

spec/index.md
```diff
@@ -651,6 +680,50 @@ The functionality exposed via the [The IC management canister](#ic-management-ca
...
+
+-   `canister_id` (`blob`): The principal of the canister to call.
+
+-   `method_name` (`text`): Name of the canister method to call
+
+-   `arg` (`blob`): Argument to pass to the canister method
+
+The HTTP response to this request can have the following responses:
+
+-   202 HTTP status with a non-empty body. Implying the request was processed.
```
mraszyk, line 699: 202 suggests that the request has been accepted for further processing which is not the case here - the status code 200 should be returned

spec/index.md
```diff
@@ -651,6 +680,50 @@ The functionality exposed via the [The IC management canister](#ic-management-ca
...
+
+-   4xx HTTP status for client errors (e.g. malformed request). Except for 429 HTTP status, retrying the request will likely have the same outcome.
+
+-   5xx HTTP status when the server has encountered an error or is otherwise incapable of performing the request. The request might succeed if retried at a later time.
+
+This request type can *also* be used to call a query method (but not a composite query method). A user may choose to go this way, instead of via the faster and cheaper [Request: Query call](#http-query) below, if they want to get a *certified* response. Note that the canister state will not be changed by sending a call request type for a query method (except for cycle balance change due to message execution).
+
+:::note
+
+The functionality exposed via the [The IC management canister](#ic-management-canister) can be used this way.
```
mraszyk, line 720: What's the point of this note?

spec/index.md
```diff
@@ -651,6 +680,50 @@ The functionality exposed via the [The IC management canister](#ic-management-ca
...
+-   `method_name` (`text`): Name of the canister method to call
+
+-   `arg` (`blob`): Argument to pass to the canister method
+
+The HTTP response to this request can have the following responses:
+
+-   202 HTTP status with a non-empty body. Implying the request was processed.
+    -   `response` (`blob`): A certificate (see [Certification](#certification)).
+
+The returned certificate includes the subtree at `/request_status/<request_id>` and `/time`.
```
mraszyk, line 702:
```suggestion
    The returned certificate includes the subtree at `/request_status/<request_id>` and `/time`.
```

spec/index.md
DSharifi: Good point. I changed it to `201` for the cases where a certification is returned. 

We can keep `202` for cases where the user needs to fall back to polling, for example if we time out the requests because no response has been generated within some time threshold.

spec/index.md
DSharifi: Removed this.

spec/index.md
mraszyk: Why not 200?

mraszyk: > We can keep 202 for cases where the user needs to fall back to polling, for example if we time out the requests because no response has been generated within some time threshold.

That makes sense!

spec/index.md
```diff
@@ -554,7 +556,13 @@ This document does not yet explain how to find the location and port of the Inte
 
 ### Overview of canister calling {#http-call-overview}
 
-Users interact with the Internet Computer by calling canisters. By the very nature of a blockchain protocol, they cannot be acted upon immediately, but only with a delay. Moreover, the actual node that the user talks to may not be honest or, for other reasons, may fail to get the request on the way. This implies the following high-level workflow:
+Users interact with the Internet Computer by calling canisters. By the very nature of a blockchain protocol, they cannot be acted upon immediately, but only with a delay. Moreover, the actual node that the user talks to may not be honest or, for other reasons, may fail to get the request on the way.
+
+The Internet Computer has two HTTPS APIs for canister calling:
+- [*Asynchronous*](#http-async-call) canister calling, where the user must poll the Internet Computer for the status of the request.
+- [*Synchronous*](#http-sync-call) canister calling, where the user waits for a certified response from the Internet Computer for the initial call.
```
mraszyk, line 563:
```suggestion
- [*Asynchronous*](#http-async-call) canister calling, where the user must poll the Internet Computer for the status of the canister call by _separate_ HTTPS requests.
- [*Synchronous*](#http-sync-call) canister calling, where the Internet Computer serves the response of the canister call as the response to the original HTTPS request.
```

spec/index.md
mraszyk: Shouldn't the note stay under the previous "Call" section?

spec/index.md
```diff
@@ -645,12 +674,52 @@ The HTTP response to this request can have the following responses:
...
+
+This request type can *also* be used to call a query method (but not a composite query method). A user may choose to go this way, instead of via the faster and cheaper [Request: Query call](#http-query) below, if they want to get a *certified* response. Note that the canister state will not be changed by sending a call request type for a query method (except for cycle balance change due to message execution).
+
 :::note
 
 The functionality exposed via the [The IC management canister](#ic-management-canister) can be used this way.
 
 :::
 
+See [The system state tree](#state-tree) for details on the state tree.
```
mraszyk, line 720: This looks like a copy-paste error. Could you please double check?

spec/index.md
```diff
@@ -613,6 +621,27 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
 
 When asking the IC about the state or call of a request, the user uses the request id (see [Request ids](#request-id)) to read the request status (see [Request status](#state-tree-request-status)) from the state tree (see [Request: Read state](#http-read-state)).
 
+#### Synchronous canister calling {#http-sync-call}
+
+Unlike asynchronous canister calling, synchronous canister calling does not initially require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. However, if a certified response is not received (due to a timeout or pre-execution error), the user must poll the Internet Computer for the status. Still, the state transitions and semantics of the call itself are the same as for asynchronous canister calling.
```
mraszyk, line 626: In case of a pre-execution error, polling makes no sense since the request status is not in the certified state tree (in such a case, the HTTP handler synchronously rejects the request).

mraszyk, line 626:
```suggestion
Unlike asynchronous canister calling, synchronous canister calling does not initially require the user to poll the Internet Computer for the status of the request. Instead, the user waits for a certified response from the Internet Computer for the initial call. However, if a certified response is not received (due to a timeout), the user must poll the Internet Computer for the status of the request. In that case, the state transitions and semantics of the call itself are the same as for asynchronous canister calling.
```

spec/index.md
mraszyk: Instead of repeating the steps, I'd just add a few sentences describing a synchronous call as the replica making an asynchronous call followed by polling for its status on behalf of the caller (until a timeout).

spec/index.md
```diff
@@ -645,12 +674,52 @@ The HTTP response to this request can have the following responses:
 
 This request type can *also* be used to call a query method (but not a composite query method). A user may choose to go this way, instead of via the faster and cheaper [Request: Query call](#http-query) below, if they want to get a *certified* response. Note that the canister state will not be changed by sending a call request type for a query method (except for cycle balance change due to message execution).
 
+### Request: Sync Call {#http-sync-call}
+A synchronous update call, or "call and await", is a type of update [call](#http-call) where the replica will wait with replying to the user until the call has been processed and the result has been added to the replicated state. This means the user will receive a certified result of the call, and thus __do not need to poll__ [`read_state`](#http-read-state) to determine the status of the call.
```
mraszyk, line 678:
```suggestion
A synchronous update call, or "call and await", is a type of update [call](#http-call) where the replica will wait with replying to the user until the call has been processed and the result has been added to the certified state. This means the user will receive a certified result of the call, and thus __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the status of the call.
```

spec/index.md
```diff
@@ -645,12 +674,52 @@ The HTTP response to this request can have the following responses:
...
+
+-   `canister_id` (`blob`): The principal of the canister to call.
+
+-   `method_name` (`text`): Name of the canister method to call
+
+-   `arg` (`blob`): Argument to pass to the canister method
+
+The HTTP response to this request can have the following responses:
+
+-   201 HTTP status with a non-empty body. Implying the request was processed.
```
mraszyk, line 693:
```suggestion
-   201 HTTP status with a non-empty body. Implying the request was processed.

```

spec/index.md
alin-at-dfinity: I strongly believe that the replica should be allowed to conclude a synchronous call before it times out. E.g. if it is stuck in `processing` for too long. Or even IMHO if it's in the received` state for more than a minute or so (someone could attempt to DoS a single canister with synchronous calls; we don't want a replica maintaining 300k concurrent synchronous calls).

So maybe we should be more explicit about replicas being able to terminate a synchronous call at any point before the actual call has timed out or completed.

Could be as simple as replacing "if a certified response is not received (due to a timeout)" with "if a certified response is not produced in reasonable time".

spec/index.md
DSharifi: Thanks for the feedback!

>I do not think HTTP 201 is appropriate here. While the request method is POST, it is not true that the request "has led to the creation of a resource".

Can't we argue that response status + Certification that's returned (which is added to the replicated state) is a "creation of a resource"?

I agree that making `200` generic over `reply` and `reject` would make things more consistent with how canister calls work internally. However, reserving `200` for `reject` makes the API of our sync endpoint more consistent with the async endpoint for the user.

I don't have a strong opinion about either option. We can go with a generic `200` if you think it makes more sense.

CC: @rumenov

spec/index.md
DSharifi: I added it since we are returning the state of the ingress call. It should be changed to:

```suggestion
See [Request status](#state-tree-request-status) for details on the request status.
```

spec/index.md
DSharifi: What if the replica is malicious? I.e. it sends a `reject` response back to the user, but actually submitted the ingress message.

mraszyk:
Note that if we return 200 for both a pre-execution error and a response certificate, then we should introduce a new enumeration over those two cases.

spec/index.md
DSharifi: I rewrote the section and removed this part.

DSharifi: Agreed. It's more explicit about it now.

spec/index.md
```diff
@@ -613,6 +621,27 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+2.  The IC asks the targeted canister if it is willing to accept this message and be charged for the expense of processing it. This uses the [Ingress message inspection](#system-api-inspect-message) API for normal calls. For calls to the management canister, the rules in [The IC management canister](#ic-management-canister) apply.
+
+3.  At some point, the IC may accept the call for processing and set its status to `received`. This indicates that the IC as a whole has received the call and plans on processing it (although it may still not get processed if the IC is under high load).
+
+4.  If the call is processed (sufficient resources, call not yet expired), it will be executed, for some calls this may be atomic, for others this involves multiple internal steps.
+
+5.  Eventually, a response will be produced which will be replied the user. The response can be a `reply`, indicating success, a `reject`, indicating some form of error.
+
+6.  In the case that the call has been retained for long enough before a response is generated, but the request has not expired yet, the IC can forget the response data and only remember the call as `done`, to prevent a replay attack.
```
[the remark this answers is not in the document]
DSharifi: Thanks. I tried to simplify the section. Could you take another look?

spec/index.md
DSharifi: I removed the step-by-step points. Can you take another look?

spec/index.md
```diff
@@ -651,6 +669,46 @@ The functionality exposed via the [The IC management canister](#ic-management-ca
 
 :::
 
+### Request: Sync Call {#http-sync-call}
+
+In order to make a synchronous update call to a canister, the user makes a POST request to `/api/v2/canister/<effective_canister_id>/sync_call`. The request body consists of an authentication envelope with a `content` map with the following fields:
+
+-   `request_type` (`text`): Always `sync-call`
```
mraszyk, line 676: or sync_call

spec/index.md
```diff
@@ -651,6 +669,46 @@ The functionality exposed via the [The IC management canister](#ic-management-ca
...
+
+-   202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use [`read_state`](#http-read-state) to determine the status of the call.
+
+-   200 HTTP status with a non-empty body. The response contains the canister response, which is either a `reply` or a `reject`.
+    
+    -   If the update call resulted in a reply, the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
+
+        -   `reply` (`blob`):  A certificate (see [Certification](#certification)) with the subtree at `/request_status/<request_id>` and `/time`.
+
+    -  If the update call resulted in a reject, the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
```
mraszyk, line 696: We should not return two different types of CBOR maps without giving the user a way to tell what to try parsing. Please see my suggestion on the PR: let's introduce an enumeration and wrap the two CBOR maps (for reply and reject) as variants of that enumeration.

spec/index.md
```diff
@@ -651,6 +669,46 @@ The functionality exposed via the [The IC management canister](#ic-management-ca
...
+
+        -   `error_code` (`text`): an optional implementation-specific textual error code (see [Error codes](#error-codes)).
+
+-   4xx HTTP status for client errors (e.g. malformed request). Except for 429 HTTP status, retrying the request will likely have the same outcome.
+
+-   5xx HTTP status when the server has encountered an error or is otherwise incapable of performing the request. The request might succeed if retried at a later time.
+
+This request type can *also* be used to call a query method (but not a composite query method). A user may choose to go this way, instead of via the faster and cheaper [Request: Query call](#http-query) below, if they want to get a *certified* response. Note that the canister state will not be changed by sending a call request type for a query method (except for cycle balance change due to message execution).
+
+See [Request status](#state-tree-request-status) for details on the request status.
```
mraszyk, line 710: Wouldn't this naturally be a comment next to the `reply` case?

mraszyk (commented):
Thank you for implementing the changes suggested so far!

spec/index.md
DSharifi: I added a `status` field to both CBOR maps with the value set to either `rejected` or `replied`. This is the same way we handle reject and reply for Query calls.  https://internetcomputer.org/docs/current/references/ic-interface-spec#http-query

Is this fine by you?

spec/index.md
```diff
@@ -651,6 +669,48 @@ The functionality exposed via the [The IC management canister](#ic-management-ca
...
+
+-   `canister_id` (`blob`): The principal of the canister to call.
+
+-   `method_name` (`text`): Name of the canister method to call
+
+-   `arg` (`blob`): Argument to pass to the canister method
+
+The HTTP response to this request can have the following responses:
+
+-   202 HTTP status with empty body. Implying the request was accepted by the IC for further processing. Users should use [`read_state`](#http-read-state) to determine the status of the call.
```
alin-at-dfinity, line 688: Nit: This may make more sense as the second entry, both because 202 comes after 200 and because it's the fallback behavior if a response is not produced in reasonable time.

Also, it may be worth explicitly stating that this is the fallback in case a response was not produced in reasonable time.

spec/index.md
```diff
@@ -2855,7 +2915,7 @@ A reference implementation would likely maintain a separate list of `messages` f
 
 #### API requests
 
-We distinguish between the *asynchronous* API requests (type `Request`) passed to `/api/v2/…/call`, which may be present in the IC state, and the *synchronous* API requests passed to `/api/v2/…/read_state` and `/api/v2/…/query`, which are only ephemeral.
+We distinguish between the *asynchronous* API requests (type `Request`) passed to `/api/v2/…/call` and `/api/v2/…/sync_call`, which may be present in the IC state, and the *synchronous* API requests passed to `/api/v2/…/read_state` and `/api/v2/…/query`, which are only ephemeral.
```
alin-at-dfinity, line 2918: This is slightly confusing: above (and in the API name) we explicitly call `sync_call` "synchronous". Yet here we categorize it as asynchronous.

Is there any possibility to use a different categorization? If not, maybe the endpoint could be named `call_and_await`.

spec/index.md
```diff
@@ -3099,7 +3159,7 @@ The following is an incomplete list of invariants that should hold for the abstr
 
 Based on this abstract notion of the state, we can describe the behavior of the IC. There are three classes of behaviors:
 
--   Asynchronous API requests that are submitted via `/api/v2/…/call`. These transitions describe checks that the request must pass to be considered received.
+-   Asynchronous API requests that are submitted via `/api/v2/…/call` and `/api/v2/…/sync_call`. These transitions describe checks that the request must pass to be considered received.
```
alin-at-dfinity, line 3162: Ditto.

spec/index.md
mraszyk: > Yet here we categorize it as asynchronous.

It is asynchronous in the worst-case.

spec/index.md
mraszyk: Yes. That makes sense - Thank you for taking inspiration from query calls!

spec/index.md
mraszyk: But I agree that it'd make sense to make this more precise and ideally also formally define the response to sync_call on the happy path: a certificate similar to the case of read state request responses.

DSharifi: [Changed ](https://github.com/dfinity/interface-spec/pull/265/commits/07a46146730e44a1f8ffd5a472c63f3116c81d44)

alin-at-dfinity: SGTM

I'm just wondering whether "call" or "update call" is the most precise term. "Call" seems too generic, since queries are also calls. And "update call" is too specific, as (I believe that) replicated queries are not technically update calls.

Maybe we should use "replicated call", which should cover all ingress messages. Although I suppose that certified queries could also be claimed to be "replicated calls".

That being said, the above is mostly nitpicking on my part. So feel free to ignore.

spec/index.md
```diff
@@ -643,14 +661,56 @@ The HTTP response to this request can have the following responses:
...
+
+-   200 HTTP status with a non-empty body. The response contains the canister response, which is either a `reply` or a `reject`.
+    
+    -   If the update call resulted in a (replicated) reply, the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
+
+        -   `status` (`text`): `"replied"`
+
+        -   `reply` (`blob`):  A certificate (see [Certification](#certification)) with subtrees at `/request_status/<request_id>` and `/time`. See [Request status](#state-tree-request-status) for more details on the request status.
+
+    -   If the update call resulted in a (non-replicated) reject, the response is a CBOR map with the following fields:
```
mraszyk, line 696: This is pretty subtle, a sync_call can be rejected in the replicated mode with a certificate or in the non-replicated mode without a certificate.

spec/index.md
mraszyk: > replicated queries are not technically update calls.

I believe they are so changed to "update calls".

Dfinity-Bjoern (approved):

spec/index.md
```diff
@@ -643,14 +663,56 @@ The HTTP response to this request can have the following responses:
...
+
+        -   `status` (`text`): `"rejected"`
+
+        -   `reject_code` (`nat`): The reject code (see [Reject codes](#reject-codes)).
+
+        -   `reject_message` (`text`): a textual diagnostic message.
+
+        -   `error_code` (`text`): an optional implementation-specific textual error code (see [Error codes](#error-codes)).
+
+-   202 HTTP status with an empty body. This status is returned as a fallback mechanism for when a response from the canister was not produced in reasonable time. This implies that the request was accepted by the IC for further processing. Users should use [`read_state`](#http-read-state) to determine the status of the call.
```
rumenov, line 708: What will be a "reasonable time "? Should we just use the `ingress_expiry` value. The annoying part is that when the ingress_expiry is hit then we will be returning 202, which I am not sure how intuitive it is.

frederikrothenberger: Just my two cents here: AFAIK `ingress_expiry` is usually set to 5 minutes in the future. Connection timeouts for Web Applications happen usually around ~30 seconds (in my experience). So the reasonable time should be well below the 30 seconds mark, otherwise clients will start seeing HTTP 408 / HTTP 504 and the like (instead of the eventually produced 202 status).

rumenov (approved):

spec/index.md
```diff
@@ -532,9 +532,11 @@ Users have the ability to learn about the hash of the canister's module, its cur
 
 ## HTTPS Interface {#http-interface}
 
-The concrete mechanism that users use to send requests to the Internet Computer is via an HTTPS API, which exposes three endpoints to handle interactions, plus one for diagnostics:
+The concrete mechanism that users use to send requests to the Internet Computer is via an HTTPS API, which exposes four endpoints to handle interactions, plus one for diagnostics:
 
--   At `/api/v2/canister/<effective_canister_id>/call` the user can submit (asynchronous, potentially state-changing) calls.
+-   At `/api/v2/canister/<effective_canister_id>/call` the user can submit update calls that are asynchronous and might change the IC state.
+
+-   At `/api/v2/canister/<effective_canister_id>/sync_call` the user can submit update calls and either get a synchronous HTTPS response with a certified status (if the update call completes within one execution round) or have the update call processed asynchronously.
```
mraszyk, line 539:
```suggestion
-   At `/api/v2/canister/<effective_canister_id>/sync_call` the user can submit update calls and either get a synchronous HTTPS response with a certified status or have the update call processed asynchronously.
```
I wouldn't specify those implementation details.

spec/index.md
```diff
@@ -613,6 +621,22 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
 
 When asking the IC about the state or call of a request, the user uses the request id (see [Request ids](#request-id)) to read the request status (see [Request status](#state-tree-request-status)) from the state tree (see [Request: Read state](#http-read-state)).
 
+#### Synchronous canister calling {#http-sync-call}
+
+A synchronous update call, also known as a "call and await", is a type of update call where the replica will respond with a certificate if the call completes within one execution round.
```
mraszyk, line 626:
```suggestion
A synchronous update call, also known as a "call and await", is a type of update call where the replica will respond with a certificate if the call completes within an implementation-specific timeout.
```

spec/index.md
```diff
@@ -613,6 +621,22 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+The purpose of the synchronous call endpoint is to give the user a certified response for their canister call. This means that if a certificate is returned, then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call.
+
+A certificate is returned by the IC to the user if:
+
+-   The execution of the call completes within one execution round, and the status is certified by the IC within the ingress expiry time. The status will be either `replied`, `rejected` or `done`.
```
mraszyk, line 632:
```suggestion
-   The execution of the call completes within an implementation-specific timeout. The status will be either `replied`, `rejected` or `done`.
```
Note that returning a certificate with `done` doesn't really reveal what happened.

spec/index.md
```diff
@@ -613,6 +621,22 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+-   The call expires before it is queued and is dropped by the IC. The status of the call will be `unknown` with the certificate time stamp at a later time than the ingress expiry time.
```
mraszyk, line 634:
```suggestion
```

spec/index.md
```diff
@@ -643,14 +669,56 @@ The HTTP response to this request can have the following responses:
...
+
+-   `method_name` (`text`): Name of the canister method to call.
+
+-   `arg` (`blob`): Argument to pass to the canister method.
+
+The HTTP response to this request can have the following responses:
+
+-   200 HTTP status with a non-empty body. The response contains the canister response, which is either a `reply` or a `reject`.
+    
+    -   If the update call completes within one execution round, or the call expires, the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
```
mraszyk, line 698:
```suggestion
    -   If the update call completes within an implementation-specific timeout, the response is a CBOR (see [CBOR](#cbor)) map with the following fields:
```

spec/index.md
```diff
@@ -643,14 +669,56 @@ The HTTP response to this request can have the following responses:
...
+
+        -   `status` (`text`): `"non_replicated_rejection"`
+
+        -   `reject_code` (`nat`): The reject code (see [Reject codes](#reject-codes)).
+
+        -   `reject_message` (`text`): a textual diagnostic message.
+
+        -   `error_code` (`text`): an optional implementation-specific textual error code (see [Error codes](#error-codes)).
+
+-   202 HTTP status with an empty body. This status is returned as a fallback mechanism for when the canister call is running for more than one execution round. This implies that the request was accepted by the IC for further processing. Users should use [`read_state`](#http-read-state) to determine the status of the call.
```
mraszyk, line 714:
```suggestion
-   202 HTTP status with an empty body. This status is returned as a fallback mechanism for when the canister call is running for more than an implementation-specific timeout. This implies that the request was accepted by the IC for further processing. Users should use [`read_state`](#http-read-state) to determine the status of the call.
```

spec/index.md
```diff
@@ -613,6 +621,22 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+A certificate is returned by the IC to the user if:
+
+-   The execution of the call completes within one execution round, and the status is certified by the IC within the ingress expiry time. The status will be either `replied`, `rejected` or `done`.
+
+-   The call expires before it is queued and is dropped by the IC. The status of the call will be `unknown` with the certificate time stamp at a later time than the ingress expiry time.
+
+The synchronous call endpoint is useful for users as it removes the networking overhead of polling the IC to determine the status of their call.
+
+Synchronous calls are therefore best suited for calls that are fast and expected to complete within one execution round. If a call does not complete within one execution round, the replica won't return a certificate and the user will need to poll the IC to determine the status of the call.
```
mraszyk, line 638:
```suggestion
Synchronous calls are therefore best suited for calls that are fast. If a call does not complete within an implementation-specific timeout, the replica won't return a certificate and the user will need to poll the IC to determine the status of the call.
```

spec/index.md
mraszyk: > The status of the call will be `unknown` with the certificate time stamp at a later time than the ingress expiry time.

Such a certificate isn't really useful.

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
 
 When asking the IC about the state or call of a request, the user uses the request id (see [Request ids](#request-id)) to read the request status (see [Request status](#state-tree-request-status)) from the state tree (see [Request: Read state](#http-read-state)).
 
+#### Synchronous canister calling {#http-sync-call-overview}
+
+A synchronous update call, also known as a "call and await", is a type of update call where the replica will attempt to respond with a certificate of the state of the call to the HTTPS request. If the returned certificate indicates that the update call is in a terminal state (`replied`, `rejected`, or `done`), then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call. A terminal state means the call has completed its execution.
```
mraszyk, line 646:
```suggestion
A synchronous update call, also known as a "call and await", is a type of update call where the replica will attempt to respond to the HTTPS request with a certificate of the call status. If the returned certificate indicates that the update call is in a terminal state (`replied`, `rejected`, or `done`), then the user __does not need to poll__ (using [`read_state`](#http-read-state) requests) to determine the result of the call. A terminal state means the call has completed its execution.
```

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+The synchronous call endpoint is useful for users as it removes the networking overhead of polling the IC to determine the status of their call.
+
+The replica will maintain the HTTPS connection for the request and will respond once the state transitions to a terminal state. If an implementation specific timeout for the request is reached while the replica waits for the terminal state, then the replica will reply before the call completes, meaning a certificate for the call state in `unknown`, `received`, or `processing` is returned.
```
mraszyk, line 650:
```suggestion
The replica will maintain the HTTPS connection for the request and will respond once the call status transitions to a terminal state. If an implementation specific timeout for the request is reached while the replica waits for the terminal state, then the replica will reply before the call completes.
```

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+-   `sender`, `nonce`, `ingress_expiry`: See [Authentication](#authentication)
+
+-   `canister_id` (`blob`): The principal of the canister to call.
+
+-   `method_name` (`text`): Name of the canister method to call.
+
+-   `arg` (`blob`): Argument to pass to the canister method.
+
+The HTTP response to this request can have the following responses:
```
mraszyk, line 674:
```suggestion
The HTTP response to this request can have the following forms:
```

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+-   200 HTTP status with a non-empty body. This status is returned if the replica received the message.
+    
+    -   A certificate for the state of the update call is produced, and returned as a CBOR (see [CBOR](#cbor)) map with the fields below. The user should use the certificate to determine the state of the call:
+
+        -   `status` (`text`): `"certified_state"`
+
+        -   `reply` (`blob`):  A certificate (see [Certification](#certification)) with subtrees at `/request_status/<request_id>` and `/time`. See [Request status](#state-tree-request-status) for more details on the request status.
+
+    -   If a non-replicated pre-processing error occurred due to the [canister inspect message](#system-api-inspect-message), then a body with information about the IC specific error encountered is returned. The body is a CBOR map with the following fields:
```
mraszyk, line 684:
```suggestion
    -   If a non-replicated pre-processing error occurred (e.g., due to the [canister inspect message](#system-api-inspect-message)), then a body with information about the IC specific error encountered is returned. The body is a CBOR map with the following fields:
```

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+        -   `status` (`text`): `"canister_inspect_message_rejected"`
```
mraszyk, line 686: this is too specific, there can be multiple reasons why a pre-processing error occurs, e.g., the canister is frozen

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+-   `canister_id` (`blob`): The principal of the canister to call.
+
+-   `method_name` (`text`): Name of the canister method to call.
+
+-   `arg` (`blob`): Argument to pass to the canister method.
+
+The HTTP response to this request can have the following responses:
+
+-   200 HTTP status with a non-empty body. This status is returned if the replica received the message.
```
mraszyk, line 676:
```suggestion
-   200 HTTP status with a non-empty body. This status is returned if the replica successfully received the message.
```

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+    
+    -   A certificate for the state of the update call is produced, and returned as a CBOR (see [CBOR](#cbor)) map with the fields below. The user should use the certificate to determine the state of the call:
```
mraszyk, line 678:
```suggestion
    -   A certificate for the state of the update call is produced, and returned in a CBOR (see [CBOR](#cbor)) map with the fields specified below. The user should use the certificate to determine the state of the call:
```

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+        -   `status` (`text`): `"certified_state"`
+
+        -   `reply` (`blob`):  A certificate (see [Certification](#certification)) with subtrees at `/request_status/<request_id>` and `/time`. See [Request status](#state-tree-request-status) for more details on the request status.
```
mraszyk, line 682:
```suggestion
        -   `reply` (`blob`):  A certificate (see [Certification](#certification)) with subtrees at `/request_status/<request_id>` and `/time`, where `<request_id>` is the [request ID](#request-id) of the update call. See [Request status](#state-tree-request-status) for more details on the request status.
```

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+
+The replica will maintain the HTTPS connection for the request and will respond once the state transitions to a terminal state. If an implementation specific timeout for the request is reached while the replica waits for the terminal state, then the replica will reply before the call completes, meaning a certificate for the call state in `unknown`, `received`, or `processing` is returned.
+
+The user should take the following actions for the returned certificate states:
+
+-    `processing`: The call is being executed by the IC. The user should poll the IC using [`read_state`](#http-read-state) requests to determine the result of the call.
+
+-    `received`: The call has been received by the IC. The user should poll the IC using [`read_state`](#http-read-state) requests to determine the result of the call.
+
+-    `unknown`: The user can conclude that the message was not received by the IC at the attached timestamp. The user can retry the call, or poll the IC using [`read_state`](#http-read-state) requests.
```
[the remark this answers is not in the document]
Dfinity-Bjoern: > The user can conclude that the message was not received by the IC at the attached timestamp.
This is *wrong*. The message may not be in the certified state tree at that timestamp, but it could be at any later time because the message may indeed be sitting in some artifact pool of some other node and be included in the next block. The message may also have been received earlier and may have been pruned. It's very hard to conclude anything from `unknown` (other than the user may want to poll until expiration to learn whether anything comes up).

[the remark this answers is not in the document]
DSharifi: I'll clarify the `unknown` case. But I still believe it is useful always to return the certificate. If the user/agent sees `processing`, or `received` then they have a signal for their message being received by the IC and that they don't need to retry the request.

spec/index.md
DSharifi: See my comment [here](https://github.com/dfinity/interface-spec/pull/265#discussion_r1576248114). We should return certificates for the other states as well as it allows the user to know whether their ingress message made it to the IC.

spec/index.md
```diff
@@ -633,8 +641,64 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+    -   If a non-replicated pre-processing error occurred due to the [canister inspect message](#system-api-inspect-message), then a body with information about the IC specific error encountered is returned. The body is a CBOR map with the following fields:
+
+        -   `status` (`text`): `"canister_inspect_message_rejected"`
+
+        -   `reject_code` (`nat`): The reject code (see [Reject codes](#reject-codes)).
+
+        -   `reject_message` (`text`): a textual diagnostic message.
+
+        -   `error_code` (`text`): an optional implementation-specific textual error code (see [Error codes](#error-codes)).
+
```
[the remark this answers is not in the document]
DSharifi: I deliberately removed the `202` status response.

I am proposing that we always return a certificate with a `200` HTTP status code if a replica successfully receives a message. It makes the API simpler, and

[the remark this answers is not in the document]
DSharifi: I deliberately removed the `202` response after a timeout.

I am proposing that we always return a certificate with a 200 HTTP status code in the new API, regardless of certification or timeout. It makes the API much simpler. 

From the user's perspective, the HTTP status code serves no value, as they can't trust the status code alone for what action to perform next. They need to inspect the body and drill down into the certificate from the subnet and c choose the next action based on the state that is certified.

Thus when the agent sees a `200` response with a certificate they can choose the next step of action, (completed, polling /read_state, retrying the HTTP request, etc.) based on the certificate.

spec/index.md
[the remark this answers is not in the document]
mraszyk: > they don't need to retry the request.

the request might still time out even after being "received"

spec/index.md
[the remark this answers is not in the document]
mraszyk: > they can't trust the status code alone

sure, but the certificate with "unknown", "received", "processing" doesn't let you conclude anything useful either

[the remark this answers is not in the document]
DSharifi: `Received` and `processing` lets you conclude that the ingress message made it to the IC.

spec/index.md
DSharifi: Changed it to "non_replicated_rejection"

spec/index.md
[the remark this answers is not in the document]
mraszyk: > Received and processing lets you conclude that the ingress message made it to the IC.

Sure, but here "making it to the IC" doesn't mean "being successfully executed".

spec/index.md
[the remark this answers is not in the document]
rumenov: I agree that in the current setup there is not much value in attaching a certificate for processing, received and unknown.

However, I am at odds with our long term plans. It was said that we should clearly not rely on the ingress history to figure out the if an update made it or not.  So in this case I would imagine that long term, if a user wants to check that his ingress message made it, then the user will have to query in certified fashion the canister state. All this is application specific. 

What is not application specific is that most likely the user will need to be able to distinguish between different timestamps. More specifically the user will need to know past what timestamp querying the canister will yield a correct result.

For example if the user queries the canister at time X to verify that an ingress message mutated the state, then X should be definitely bigger than the time at which a certificate was returned when in processing or received 
 (and unknown?!) state.

All this said, if we omit the certificate for those states now and we want to add them later, most likely it won't break much due to the cbor nature of the response.

spec/index.md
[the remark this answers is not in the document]
rumenov: If the user receives 200 with "non_replicated_rejection" the user can't conclude anything useful as well. 

The thinking here is that the user will get 200 iff it hits the wasm. In other words the IC is a transport for the execution layer in a way.

spec/index.md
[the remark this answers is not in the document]
oggy-dfin: > What is not application specific is that most likely the user will need to be able to distinguish between different timestamps. More specifically the user will need to know past what timestamp querying the canister will yield a correct result.
Based on the expiry time (which is set by the user), the user can figure out at which point the call must've made it, even without the certificate, though I'm not actually sure where the call will have made it to :) probably the input queue at best, I doubt we guarantee that the message processing actually starts. And I think this would be no better with a certificate. But even if the processing did start at this point, the canister can in principle take an arbitrarily long time to process the message. So the user will in general still need some canister-specific knowledge to figure out at which time the result of a query is definite.

spec/index.md
DSharifi: Changed it to return `202` and an empty body if it's not in a terminal state.

spec/index.md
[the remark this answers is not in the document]
DSharifi: I changed it to return 202 and an empty body if it's not in a terminal state. This means we will not return a certificate if the call status is not in `replied`, `rejected`, or `done`.

spec/index.md
```diff
@@ -639,8 +647,60 @@ Calls must stay in `replied` or `rejected` long enough for polling users to catc
...
+-   `canister_id` (`blob`): The principal of the canister to call.
+
+-   `method_name` (`text`): Name of the canister method to call.
+
+-   `arg` (`blob`): Argument to pass to the canister method.
+
+The HTTP response to this request can have the following forms:
+
+-   200 HTTP status with a non-empty body. This status is returned if the canister call completed or was rejected within an implementation-specific timeout.
```
mraszyk, line 676:
```suggestion
-   200 HTTP status with a non-empty body. This status is returned if the canister call completed.
```

DSharifi: The call can be rejected, and we still reply with `200`. Hence the `status` fields below.

mraszyk: > The call can be rejected, and we still reply with 200. Hence the status fields below.

What do you mean?

DSharifi: I mean the ingress filter can fail, and we never submit the ingress message to consensus+p2p, we reply with a `200` status code.

https://github.com/dfinity/interface-spec/pull/265/files/d6edfe5ab5d10c2b08e0d6267fe88c224841232e#diff-b9a833d6114611a080aee4cd57618feb45aa044bcc11105d44b155b5bbde00ebR684-R693

mraszyk: > I mean the ingress filter can fail, and we never submit the ingress message to consensus+p2p, we reply with a 200 status code.

This is covered by "the canister call completed" since the call completed in this case (with an error).

PR #267, opened 2024-01-29
Fix the ordering guarantees claim.

oggy-dfin:
The ordering guarantees specified in the spec are overpromising. The sender can't rely on the function calls being executed in order, and the calls are not really just like a FIFO queue, because there's no delivery guarantee for requests (the system can fail them at any time).

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1089,7 +1089,7 @@ The order in which the various messages between canisters are delivered and exec
 
 More precisely:
 
--   Method calls between any *two* canisters, if delivered, start executing in order. Note that function call delivery can fail for arbitrary reasons (e.g., high system load).
+-   Messages between any *two* canisters, if delivered to the canister, start executing in order. Note that message delivery can fail for arbitrary reasons (e.g., high system load).
 
 -   If a WebAssembly function, within a single invocation, makes multiple calls to the same canister, they are queued in the order of invocations to `ic0.call_perform`.
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -1085,11 +1085,11 @@ This section summarizes the format of the CBOR data passed to and from the entry
 
 ### Ordering guarantees
 
-The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that function calls between two canisters are executed in order, so that a canister that requires in-order execution need not wait for the response from an earlier message to a canister before sending a later message to that same canister.
+The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that successful function calls between two canisters are executed in order.
 
 More precisely:
 
--   Method calls between any *two* canisters are delivered in order, as if they were communicating over a single simple FIFO queue.
+-   Successful method calls between any *two* canisters are delivered in order. Note that function calls can fail for arbitrary reasons (e.g., high system load). Thus, canisters that need to ensure that two function calls are both executed, and moreover executed in a particular order, must wait for the completion of the first function call.
 
 -   If a WebAssembly function, within a single invocation, makes multiple calls to the same canister, they are queued in the order of invocations to `ic0.call_perform`.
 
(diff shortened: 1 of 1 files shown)


mraszyk (commented):
LGTM - let's approve after discussing in the spec meeting tomorrow

spec/index.md
```diff
@@ -1085,11 +1085,11 @@ This section summarizes the format of the CBOR data passed to and from the entry
 
 ### Ordering guarantees
 
-The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that function calls between two canisters are executed in order, so that a canister that requires in-order execution need not wait for the response from an earlier message to a canister before sending a later message to that same canister.
+The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that successful function calls between two canisters are executed in order.
 
 More precisely:
 
--   Method calls between any *two* canisters are delivered in order, as if they were communicating over a single simple FIFO queue.
+-   Successful method calls between any *two* canisters are delivered in order. Note that function calls can fail for arbitrary reasons (e.g., high system load). Thus, canisters that need to ensure that two function calls are both executed, and moreover executed in a particular order, must wait for the completion of the first function call.
```
mraszyk, line 1092:
```suggestion
-   Successful method calls between any *two* canisters are delivered in order. Note that function calls can fail for arbitrary reasons (e.g., high system load).
```

mraszyk: Discussed in the Interface spec meeting: there might be other ways for the canister to guarantee in-order execution, e.g., sequence numbers. CC @oggy-dfin

spec/index.md
```diff
@@ -1085,11 +1085,11 @@ This section summarizes the format of the CBOR data passed to and from the entry
 
 ### Ordering guarantees
 
-The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that function calls between two canisters are executed in order, so that a canister that requires in-order execution need not wait for the response from an earlier message to a canister before sending a later message to that same canister.
+The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that successful function calls between two canisters are executed in order.
```
mraszyk, line 1088:
```suggestion
The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that function calls between two canisters that start being executed do so in the order in which the corresponding messages were sent.
```

spec/index.md
mraszyk, line 1092:
```suggestion
-   Method calls between any *two* canisters start executing in order. Note that function calls can fail for arbitrary reasons (e.g., high system load). Thus, canisters that need to ensure that two function calls are both executed, and moreover executed in a particular order, must wait for the completion of the first function call.
```

oggy-dfin: I don't quite like that suggestion because it suggests that calls will start executing (whereas one of them might not); but I think I understand your point, so I'll give it another shot; lmk what you think.

spec/index.md
```diff
@@ -1085,11 +1085,11 @@ This section summarizes the format of the CBOR data passed to and from the entry
 
 ### Ordering guarantees
 
-The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that function calls between two canisters are executed in order, so that a canister that requires in-order execution need not wait for the response from an earlier message to a canister before sending a later message to that same canister.
+The order in which the various messages between canisters are delivered and executed is not fully specified. The guarantee provided by the IC is that if a canister sends two messages to a canister and they both start being executed by the receiving canister, then they do so in the order in which the messages were sent.
 
 More precisely:
 
--   Method calls between any *two* canisters are delivered in order, as if they were communicating over a single simple FIFO queue.
+-   Method calls between any *two* canisters, if delivered, start executing in order. Note that function call delivery can fail for arbitrary reasons (e.g., high system load).
```
mraszyk, line 1092:
```suggestion
-   Messages between any *two* canisters, if delivered to the canister, start executing in order. Note that message delivery can fail for arbitrary reasons (e.g., high system load).
```
unifying the wording a bit

PR #268, opened 2024-01-29
[FINAL] feat: Best-effort responses

oggy-dfin:
To ensure scalability, we want to allow canister authors to use best-effort responses to inter-canister calls, which time out after a period of time.

Details are in IC-1674 and the forum post:

https://forum.dfinity.org/t/scalable-messaging-model/26920


mraszyk (commented):
I believe we should also update the formal part of the spec and model the call timeouts there.

oggy-dfin:
@mraszyk yes, but I would like to first agree on the signatures and the informal description of the semantics with the stakeholders. Hence the draft status :)

spec/index.md
```diff
@@ -1550,6 +1557,16 @@ If this traps (e.g. runs out of cycles), the state changes from the `cleanup` fu
 
 There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` and `ic0.call_perform`.
 
+-   `ic0.call_set_timeout : (seconds : i32) -> ()`
+
+    Requests the system to generate a synthetic timeout reject if no reply is received within the specified amount of seconds. The returned reject code will be `SYS_TIMEOUT`.
+
+    The `seconds` parameter is only advisory. First, it is silently bounded by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. Second, while the system will attempt not to time out the call too early, it may in exceptional circumstances trigger the timeout earlier than requested; the calling canister should not rely on the timeout not occurring early. Finally, even under normal circumstances, the timeout may trigger significantly later than the specified time.
```
derlerd-dfinity, line 1564: I'd treat the "in exceptional circumstances" as an implementation detail and not mention it here. Mainly to avoid the situation where somebody reads this as "it is fine if the canister can't handle this case as it will never happen". 
```suggestion
    The `seconds` parameter is only advisory. First, it is silently bounded by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. Second, it may (if the load on the subnet demands it) trigger the timeout earlier than requested; the calling canister should not rely on the timeout not occurring early. Finally, even under normal circumstances, the timeout may trigger significantly later than the specified time.
```

derlerd-dfinity, line 1564: Could you maybe comment on why we need the
```
Finally, even under normal circumstances, the timeout may trigger significantly later than the specified time.
```
Is it just to give the implementation more flexibility, or is there a concrete concern that requires this formulation?

oggy-dfin: Well, it explicitly says in that sentence that you should not rely on the timeout not triggering early :) But I can change, I just wouldn't mention subnets since we so far don't have a real treatment of them in the document.

oggy-dfin: I was thinking of something like another DTS message handler already being scheduled; I don't know how we'd prioritize that, especially if we have several of those. In the extreme case, if you simultaneously make 10 calls that time out, and each timeout message handler does several DTS rounds, you could end up triggering the timeout significantly later (even though the system is running normally).

derlerd-dfinity: Good point.

Another reason is probably also the subnet time jumping in bigger increments than expected because of some issues leading to lower FR or similar. More generally, one could probably say that timeouts will be at most as fine grained as the increments in terms of subnet time.

alin-at-dfinity: Technically, if we're simply considering the effect of the subnet/canister being backlogged (as opposed e.g. optimizations in how we trigger timeouts that might result in timeouts only being triggered once ever N seconds), the timeout will still trigger on time. It's just that we cannot guarantee its immediate execution. Which is true for every kind of execution (including timers), so I would rather leave it out.

spec/index.md
```diff
@@ -1550,6 +1564,14 @@ If this traps (e.g. runs out of cycles), the state changes from the `cleanup` fu
 
 There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` and `ic0.call_perform`.
 
+-   `ic0.call_with_best_effort_response : (timeout_seconds : i32) -> ()`
+
+    Allows the system to generate a synthetic reject response without waiting for the actual response from the callee, even if the callee produces a response. The reject callback will be triggered with either `SYS_TRANSIENT` or `SYS_UNKNOWN` reject codes. If the code is `SYS_TRANSIENT`, the caller can be sure that the call never made it to the callee. If it is `SYS_UNKNOWN`, then the call **may or may not** have been processed by the callee; for example, it could be that processing the call took too long, that the reply took too long to deliver to the calling canister, or simply that the system decided that it needed to shed load. In this case, if the calling canister needs to know whether the call was successful, it must find an out-of-band way of doing so. For example, if the callee provides idempotent function calls, the calling canister can simply retry the call.
```
alin-at-dfinity, line 1569:
```suggestion
    Allows the system to generate a synthetic reject response without waiting for the actual response. The reject callback will be triggered with either `SYS_TRANSIENT` or `SYS_UNKNOWN` reject codes. If the code is `SYS_TRANSIENT`, the caller can be sure that the call never made it to the callee. If it is `SYS_UNKNOWN`, then the call **may or may not** have been processed by the callee; for example, it could be that processing the call took too long, that the reply took too long to deliver to the calling canister, or simply that the system decided that it needed to shed load. In this case, if the calling canister needs to know whether the call was successful, it must find an out-of-band way of doing so. For example, if the callee provides idempotent function calls, the calling canister can simply retry the call.
```

spec/index.md
```diff
@@ -1550,6 +1564,14 @@ If this traps (e.g. runs out of cycles), the state changes from the `cleanup` fu
...
+
+    The `timeout_seconds` parameter is a hint to the system when the reject should be produced. However, it is only advisory. First, it is silently bounded by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. Second, the system may invoke the reject callback earlier than requested (e.g., under high load); the calling canister should not rely on the timeout not occurring early. Finally, the reject callback may also be executed (possibly significantly) later than the specified time, e.g., if the canister is under high load.
```
alin-at-dfinity, line 1571: Personal opinion: I would separate timing out a call from dropping it to shed load. That way a timeout reject would only ever be produced after (when?) the deadline has expired.

But similar to how we can already reject a request with a `SYS_TRANSIENT` reject code at any time, because the caller explicitly selected to make a best-effort call, the system could also terminate the caller's end of the call at any time with a `SYS_UNKNOWN` reject code. Without it involving a time out.

spec/index.md
```diff
@@ -1424,6 +1432,12 @@ The canister can access an argument. For `canister_init`, `canister_post_upgrade
 
     The reject message. Traps if there is no reject message (i.e. if `reject_code` is `0`).
 
+-   `ic0.msg_deadline : () -> i64`
+
+    The deadline, in nanoseconds since 1970-01-01, by which the caller expect to see a response.
```
alin-at-dfinity, line 1437:
```suggestion
    The deadline, in nanoseconds since 1970-01-01, after which the caller will stop waiting for a response.
```

spec/index.md
oggy-dfin: Unless we want to define another reject code, I don't see how we can't distinguish between timeouts and premature dropping at the interface spec level, given the current choice not to put ICXXX codes that aren't specified in the interface spec?

alin-at-dfinity: We should definitely not define another reject code.

Since no IC*XXX* error codes are defined in the spec, there is no documented way of figuring out e.g. what kind of transient error has occurred (queue full, out of memory, subnet split, whatever). By that measure, we shouldn't concern ourselves with how to differentiate between the various `SYS_UNKNOWN` root causes (for now only "timed out" and "dropped").

All we need to do is state that there are 2 possible reasons why a best-effort call might end up in an unknown state: timeouts and load shedding. Bundling both under timeouts -- and having to explain that load shedding is essentially an early timeout and thus your specified timeout is meaningless -- is backwards IMHO. No other reject code has a single root cause behind it, so we shouldn't aim to make `SYS_UNKNOWN` different.

oggy-dfin: There's https://wiki.internetcomputer.org/wiki/Error_Codes_returned_by_Internet_Computer which documents the embedded error codes, we should add them here.

I've removed "timeout too early now", I hope that helps. Please suggest a rewording if you still think it's confusing. My ask is just that we clearly spell out to the spec user that the reject callback can be triggered both earlier and later than the specified timeout.

spec/index.md
```diff
@@ -1424,6 +1432,12 @@ The canister can access an argument. For `canister_init`, `canister_post_upgrade
 
     The reject message. Traps if there is no reject message (i.e. if `reject_code` is `0`).
 
+-   `ic0.msg_deadline : () -> i64`
+
+    The deadline, in nanoseconds since 1970-01-01, after which the caller will stop waiting for a response.
+
+    For calls with best-effort responses, the deadline is computed based on the time the call was made, and the `timeout_seconds` parameter provided by the caller. For other calls, the deadline will be set to 0.
```
alin-at-dfinity, line 1439:
```suggestion
    For calls with best-effort responses, the deadline is computed based on the time the call was made, and the `timeout_seconds` parameter provided by the caller. For other calls, a deadline of 0 will be returned.
```

spec/index.md
```diff
@@ -1550,6 +1564,14 @@ If this traps (e.g. runs out of cycles), the state changes from the `cleanup` fu
 
 There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` and `ic0.call_perform`.
 
+-   `ic0.call_with_best_effort_response : (timeout_seconds : i32) -> ()`
+
+    Allows the system to generate a synthetic reject response without waiting for the actual response. The reject callback will be triggered with either `SYS_TRANSIENT` or `SYS_UNKNOWN` reject codes. If the code is `SYS_TRANSIENT`, the caller can be sure that the call never made it to the callee. If it is `SYS_UNKNOWN`, then the call **may or may not** have been processed by the callee; for example, it could be that processing the call took too long, that the reply took too long to deliver to the calling canister, or simply that the system decided that it needed to shed load. In this case, if the calling canister needs to know whether the call was successful, it must find an out-of-band way of doing so. For example, if the callee provides idempotent function calls, the calling canister can simply retry the call.
```
alin-at-dfinity, line 1569:
```suggestion
    Allows the system to generate a synthetic reject response without waiting for the actual response. The reject callback will have a reject code of either `SYS_TRANSIENT` or `SYS_UNKNOWN`. If the code is `SYS_TRANSIENT`, the caller can be sure that the call never made it to the callee. If it is `SYS_UNKNOWN`, then the call **may or may not** have been processed by the callee; for example, it could be that processing the call did not complete in time, that delivering the reply to the calling canister took too long, or simply that the system found it necessary to shed load. In this case, if the calling canister needs to know whether the call was successful, it must find an out-of-band way of doing so. For example, if the callee provides idempotent function calls, the calling canister can simply retry the call.
```

spec/index.md
alin-at-dfinity:
```suggestion
    The `timeout_seconds` parameter specifies how long the system should wait for a response before producing a timeout reject. It is silently bounded by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The reject callback may be executed (possibly significantly) later than the specified time, e.g., if the canister is under high load. Also note that the system may reject a call for other reasons (e.g., under high load); the calling canister should not rely on the reject callback not being called before timeout expiration.
```

Maybe also squeeze in something about `SYS_UNKNOWN` and `SYS_TRANSIENT`. Both timeouts and other rejects may produce either reject code.

spec/index.md
```diff
@@ -1008,6 +1012,8 @@ Rejection codes are member of the following enumeration:
 
 -   `CANISTER_ERROR` (5): Canister error (e.g., trap, no response)
 
+-   `SYS_UNKNOWN` (6): Response unknown; aborted waiting for it (e.g., timed out, or system under high load).
```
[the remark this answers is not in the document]
oggy-dfin: Added "system stopped waiting for it"

spec/index.md
oggy-dfin: Yes I realize that the first paragraph was wrong in that respect 😬 I modified it again now...

spec/index.md
```diff
@@ -1566,9 +1566,11 @@ There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` a
...
 
-    The `timeout_seconds` parameter specifies how long the system should wait for a response before producing a timeout reject. It is silently bounded by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The reject callback may be executed (possibly significantly) later than the specified time, e.g., if the canister is under high load. Also note that the system may, at any time, reject a call for other reasons (e.g., under high load); the calling canister should not rely on the reject callback not being called before timeout expiration.
+     The `timeout_seconds` is silently bounded from above by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The reject callback may be executed (possibly significantly) later than the specified time, e.g., if the canister is under high load. The reject callback can also executed before timeout expiration, and possibly with a `SYS_UNKNOWN` code if the system is under load.
```
alin-at-dfinity, line 1571:
```suggestion
     The `timeout_seconds` is silently bounded from above by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The reject callback may be executed (possibly significantly) later than the specified time, e.g., if the canister is under high load. The reject callback may also be executed before timeout expiration, with a `SYS_UNKNOWN` code, if the system is under load.
```

Or

```suggestion
     The `timeout_seconds` is silently bounded from above by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The reject callback may be executed (possibly significantly) later than the specified time, e.g., if the canister is under high load. If the system is under load, it may also execute the reject callback with a `SYS_UNKNOWN` code before timeout expiration.
```

alin-at-dfinity (dismissed):
LGTM

spec/index.md
```diff
@@ -2929,7 +2957,7 @@ Finally, we can describe the state of the IC as a record having the following fi
     }
     CanStatus
       = Running
-      | Stopping (List (CallOrigin, Nat))
+      | Stopping (List (CallOrigin, Nat, NoDeadline | Timestamp | Expired))
```
alin-at-dfinity, line 2960: Could you please explain what's going on here?

I'm too lazy to figure it out myself, but at first sight it seems to me like when a canister is stopping we're doing or not doing something based on whether the deadline of a call context has expired. I wouldn't expect us to have to do anything under any circumstances when a call context's deadline has expired (except maybe immediately drop any response).

oggy-dfin: This is just a list of stopping calls; each of these calls can have a deadline attached, so we track the deadline. These are basically management canister call contexts, just for this one call.

But generally the main complication in the abstract system in this PR is that once we respond with `SYS_UNKNOWN` to a request, we need to drop any further responses for this request. In the implementation we do this through `CallContextManager::calbacks`, but we don't have this structure in the spec. Instead, we are allowed spooky action at a distance, so we can issue a reject message and simultaneously change in flight messages or downstream call contexts. `Expired` basically means that a `SYS_UNKNOWN` response has already been generated, so while we can proceed with whatever work we're doing, in the end we'll just drop the response (`Expired` feels a bit off as a name, happy to take suggestions)

alin-at-dfinity: [Damn, was going to reply and figured that pressing *Resolve conversation* would both post and resolve. Instead it just lost my post.]

Thanks a lot for the clarification.

And `Expired` seems quite OK to me.

spec/index.md
```diff
@@ -1284,6 +1290,7 @@ The following sections describe various System API functions, also referred to a
     ic0.msg_reject_code : () -> i32;                                            // Ry Rt CRy CRt
     ic0.msg_reject_msg_size : () -> i32;                                        // Rt CRt
     ic0.msg_reject_msg_copy : (dst : i32, offset : i32, size : i32) -> ();      // Rt CRt
+    ic0.msg_deadline : () -> i64;                                               // U Ry Rt
```
mraszyk, line 1293: Would we want to add (trivial) support for composite query calls so that code can be reused for composite query calls and update calls?

spec/index.md
```diff
@@ -3702,6 +3738,77 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+#### Spontaneous request rejection {#request-rejection}
+
+The system can reject a request at any point in time, e.g., because it is overloaded. The following specification is an overapproximation of the rejection behavior. In particular, the system is guaranteed not to emit `DESTINATION_INVALID` in some cases, e.g., if the callee exists, and if the caller successfully calls the caller once, subsequent calls will not fail with `DESTINATION_INVALID`.
+
+Condition:
+```html
+S.messages = Older_messages · CallMessage CM · Younger_messages
+(CM.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
+reject_code ∈ { SYS_FATAL, SYS_TRANSIENT, DESTINATION_INVALID }
+∃reject_msg: True
```
mraszyk, line 3750: I guess we could also bind it in the textual description above to avoid this trivial predicate.

spec/index.md
```diff
@@ -3702,6 +3738,77 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+S.messages = Older_messages · M { deadline = Expired } · Younger_messages ·
+    ResponseMessage {
+        origin = CM.origin;
+        response = Reject (SYS_UNKNOWN, reject_msg);
+        refunded_cycles = 0;
+        deadline = NoDeadline;
+    }
+```
+
+The next transition defines the expiry of calls that are being processed by the callee.
```
mraszyk, line 3791: Wouldn't we have two reject messages with SYS_UNKNOWN at the end?

spec/index.md
```diff
@@ -3744,10 +3852,10 @@ Conditions
 ```html
 
 S.call_contexts[Ctxt_id].needs_to_respond = false
-∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
-∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
-∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ≠ Ctxt_id
-∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ≠ Ctxt_id
+∀ CallMessage {origin = FromCanister O, deadline, …} ∈ S.messages. O.calling_context ≠ Ctxt_id ∨ deadline = Expired
```
mraszyk, line 3855: Do we need ` ∨ deadline = Expired` or can we assume that the (expired) call context will be removed before applying this transition?

spec/index.md
```diff
@@ -5102,6 +5234,7 @@ RM.origin = FromCanister {
   }
 S.call_contexts[Ctxt_id].deleted
 S.call_contexts[Ctxt_id].canister ∈ dom(S.balances)
+RM.deadline ≠ Expired
```
mraszyk, line 5237: I guess we need an analogous transition for expired messages that just drops them without any refund.

spec/index.md
```diff
@@ -1322,6 +1329,7 @@ The following sections describe various System API functions, also referred to a
       ) -> ();
     ic0.call_on_cleanup : (fun : i32, env : i32) -> ();                         // U CQ Ry Rt CRy CRt T
     ic0.call_data_append : (src : i32, size : i32) -> ();                       // U CQ Ry Rt CRy CRt T
+    ic0.call_with_best_effort_response : (timeout_seconds : i32) -> ();         // U Ry Rt T
```
mraszyk, line 1332: Let's please also model those functions formally in the "Abstract Canisters to System API" section.

spec/index.md
```diff
@@ -1424,6 +1432,12 @@ The canister can access an argument. For `canister_init`, `canister_post_upgrade
 
     The reject message. Traps if there is no reject message (i.e. if `reject_code` is `0`).
 
+-   `ic0.msg_deadline : () -> i64`
+
+    The deadline, in nanoseconds since 1970-01-01, after which the caller will stop waiting for a response.
+
+    For calls with best-effort responses, the deadline is computed based on the time the call was made, and the `timeout_seconds` parameter provided by the caller. For other calls, a deadline of 0 will be returned.
```
Dfinity-Bjoern, line 1439: Did we consider the effects of unsynchronized clocks between subnets?

spec/index.md
oggy-dfin: Done

spec/index.md
oggy-dfin: As discussed in the meeting last week, the call life cycle ensures that only one `CallMessage`, `CallContext` or a `ResponseMessage` exists for a given call at a given point in time. With response timeouts, this still holds as long as we ignore those values whose deadline is `Expired` (even if they get converted to `ResponseMessage` eventually, the response will not be delivered but eventually garbage collected).

spec/index.md
oggy-dfin: We can't assume that the expired message will be removed. Here we ensure that the appropriate reject response (with `SYS_UNKNOWN`) has been delivered; this is because such responses all use `NoDeadline`, so they can't be ignored in this precondition. But downstream processing of responses can go on.

spec/index.md
oggy-dfin: As discussed in the meeting, monotonicity of time would be orthogonal to this PR (canisters can in theory observe time going back even without timeouts and best-effort responses).

Also as discussed, what would be relevant for timeouts would be large clock skews between subnets, because these would cause spurious request rejection (if the receiver subnet was >5 minutes ahead). Such rejection is technically still compliant with the spec, but canister communication would be effectively halted until the clocks are in reasonable sync again. We had an internal discussion on this and the conclusion was that we aren't worried about such a scenario. First, the canisters need to handle this situation anyways for other reasons (e.g., one of the subnets being under high load, or stalled), so such a situation shouldn't be a correctness problem, only an availability one. Second, given the existing time-related checks in our consensus mechanism, data about average quartz clock drifts, and the fact that we use NTP, we couldn't come up with a realistic scenario where we get large clock drift between subnets.

spec/index.md
```diff
@@ -1550,6 +1564,16 @@ If this traps (e.g. runs out of cycles), the state changes from the `cleanup` fu
...
+
+    Any value for `timeout_seconds` is permitted, but is silently bounded from above by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The implementation may add a specific [error code](#error-codes) to a reject message to indicate the cause, in particular whether the timeout expired. Note that the reject callback may be executed (possibly significantly) later than the specified time (e.g., if the caller is under high load), or before timeout expiration (e.g., if the system is under load).
+
+    A caller that receives a `SYS_UNKNOWN` node, yet needs to learn the call outcome, must find an out-of-band way of doing so. For example, if the callee provides idempotent function calls, the caller can simply retry the call. Sample causes of `SYS_UNKNOWN` include the call not being delivered in time, call processing not completing in time, reply delivery taking too long, and the system shedding load.
```
mraszyk, line 1573:
```suggestion
    A caller that receives a `SYS_UNKNOWN` code, yet needs to learn the call outcome, must find an out-of-band way of doing so. For example, if the callee provides idempotent function calls, the caller can simply retry the call. Sample causes of `SYS_UNKNOWN` include the call not being delivered in time, call processing not completing in time, reply delivery taking too long, and the system shedding load.
```

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+reject_code ∈ { SYS_FATAL, SYS_TRANSIENT, DESTINATION_INVALID }
+```
+
+State after (given some `reject_msg`):
+```html
+S.messages =
+    Older_messages ·
+    ResponseMessage {
+        origin = CM.origin;
+        response = Reject (reject_code, reject_msg);
```
mraszyk, line 3752:
```suggestion
        response = Reject (reject_code, <implementation-specific>);
```
that we don't need to bind `reject_msg` above and it's consistent with what we do elsewhere

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
 
 Note that by construction, a query function will either trap or return with a response; it will never send calls, and it will never change the state of the canister.
 
+#### Spontaneous request rejection {#request-rejection}
+
+The system can reject a request at any point in time, e.g., because it is overloaded. The following specification is an overapproximation of the rejection behavior. In particular, the system is guaranteed not to emit `DESTINATION_INVALID` in some cases, e.g., if the callee exists, and if the caller successfully calls the caller once, subsequent calls will not fail with `DESTINATION_INVALID`.
```
mraszyk, line 3737:
```suggestion
The system can reject a request at any point in time, e.g., because it is overloaded.
```

mraszyk: I'd drop this as the spec generally overapproximates the actual behavior and the example is quite hard to follow.

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+        origin = CM.origin;
+        response = Reject (reject_code, reject_msg);
+        refunded_cycles = CM.transferred_cycles;
+    } ·
+    Younger_messages
+```
+
+#### Call expiry {#call-expiry}
+
+These transitions expire calls with timeouts. To account for SYS_UNKNOWN being issued early (e.g., due to high system load), we ignore the caller time in these transitions. We define two variants of the transition, one that expires messages, and one that expires calls that are in progress (i.e., have open downstream call contexts).
```
mraszyk, line 3760:
```suggestion
These transitions expire calls with timeouts. To account for the rejection before the actual deadline (e.g., due to high system load), we ignore the caller time in these transitions. We define two variants of the transition, one that expires messages, and one that expires calls that are in progress (i.e., have open downstream call contexts).
```

spec/index.md
```diff
@@ -6154,6 +6271,12 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       if es.context ∉ {Rt, CRt} then Trap {cycles_used = es.cycles_used;}
       copy_to_canister<es>(dst, offset, size, es.params.reject_msg)
 
+    ic0.msg_deadline<es>() : i64 =
+        if es.context ∉ {U Q CQ Ry Rt CRy CRt} then Trap {cycles_used = es.cycles_used;}
```
mraszyk, line 6275:
```suggestion
        if es.context ∉ {U, Q, CQ, Ry, Rt, CRy, CRt} then Trap {cycles_used = es.cycles_used;}
```

spec/index.md
```diff
@@ -6306,8 +6429,17 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
...
+        timeout_seconds = NoTimeout
       }
 
+    ic0.call_with_best_effort_response<es>(timeout_seconds : i32) =
+      if
+          es.context ∉ {U, CQ, Ry, Rt, CRy, CRt, T}
+          or es.pending_call = NoPendingCall
+          or es.pending_call.timeout ≠ NoTimeout
+      then Trap {cycles_used = es.cycles_used;}
+      es.pending_call.timeout_seconds := timeout_seconds
```
mraszyk, line 6441:
```suggestion
      es.pending_call.timeout_seconds := max(timeout_seconds, MAX_CALL_TIMEOUT)
```

spec/index.md
```diff
@@ -5116,6 +5216,22 @@ S with
 
 ```
 
+#### Dropping expired messages {#message-timeout}
```
mraszyk, line 5219: Why not drop expired messages in their expiry transition instead of setting the deadline to expired and dropping them in a separate transition?

spec/index.md
mraszyk: > the call life cycle ensures that only one CallMessage, CallContext or a ResponseMessage exists for a given call at a given point in time

There can a CallContext and ResponseMessage at the same time.

spec/index.md
mraszyk: Or we do need to keep expired messages around since they can still have side-effects on the callee?

spec/index.md
```diff
@@ -5089,7 +5187,7 @@ S with
 
 ```
 
-If the responded call context does not exist anymore, because the canister has been uninstalled since, the refunded cycles are still added to the canister balance, but no function invocation is enqueued:
+If the responded call context does not exist anymore, because the canister has been uninstalled since, the refunded cycles are still added to the canister balance, but no function invocation is enqueued. No refunds are issued for expired calls.
```
mraszyk, line 5190:
```suggestion
If the responded call context does not exist anymore, because the canister has been uninstalled since, the refunded cycles are still added to the canister balance, but no function invocation is enqueued.
```
This follows from setting the refunded cycles to zero when expiring the corresponding message. I wouldn't point it out here since it's confusing (there's no special treatment of expired cycles here in terms of refunded cycles).

spec/index.md
```diff
@@ -5679,6 +5795,7 @@ We can model the execution of WebAssembly functions as stateful functions that h
       sysenv : Env;
       cycles_refunded : Nat;
       method_name : NoText | Text;
+      deadline : NoDeadline | Timestamp;
```
mraszyk, line 5798: This field is not initialized - I'd rather put it into sysenv : Env and initialize it in the abstract state machine above whenever Env is created.

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+
+These transitions expire calls with timeouts. To account for SYS_UNKNOWN being issued early (e.g., due to high system load), we ignore the caller time in these transitions. We define two variants of the transition, one that expires messages, and one that expires calls that are in progress (i.e., have open downstream call contexts).
+
+The first transition defines the expiry of messages, where `reject_msg` is some textual message describing the rejection reason.
+
+```html
+S.messages = Older_messages · M · Younger_messages
+M = CallMessage _ ∨ M = ResponseMessage _
+M.origin = FromCanister O
+O.deadline ∉ { NoDeadline, Expired }
```
mraszyk, line 3768: How can a ResponseMessage have a deadline in this spec?

spec/index.md
mraszyk: > canisters can in theory observe time going back even without timeouts and best-effort responses

how?

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+
+#### Call expiry {#call-expiry}
+
+These transitions expire calls with timeouts. To account for SYS_UNKNOWN being issued early (e.g., due to high system load), we ignore the caller time in these transitions. We define two variants of the transition, one that expires messages, and one that expires calls that are in progress (i.e., have open downstream call contexts).
+
+The first transition defines the expiry of messages, where `reject_msg` is some textual message describing the rejection reason.
+
+```html
+S.messages = Older_messages · M · Younger_messages
+M = CallMessage _ ∨ M = ResponseMessage _
```
mraszyk, line 3766:
```suggestion
M = CallMessage _ 
```

mraszyk: also see my comment a few lines below

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+```html
+ctxt_id ∈ S.call_contexts
+S.call_contexts[ctxt_id].origin = FromCanister O
+O.deadline ∉ { NoDeadline, Expired }
+```
+
+State after
+
+```html
+S.call_contexts[ctxt_id].origin = FromCanister O with deadline = Expired
```
mraszyk, line 3793: I guess we should set needs_to_respond to false since the call context effectively replied with a reject message

spec/index.md
```diff
@@ -3711,10 +3807,10 @@ Conditions
...
 
 S.call_contexts[Ctxt_id].needs_to_respond = true
-∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
-∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
-∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ≠ Ctxt_id
-∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ≠ Ctxt_id
+∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
+∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
+∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
+∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
```
mraszyk, line 3813: Management canister call contexts for stopping canisters have their own timeout flow so this is not necessary.

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+```
+
+State after (given some `reject_msg`):
+```html
+S.messages =
+    Older_messages ·
+    ResponseMessage {
+        origin = CM.origin;
+        response = Reject (reject_code, reject_msg);
+        refunded_cycles = CM.transferred_cycles;
```
derlerd-dfinity, line 3753: Can the system really refund the cycles here in all cases, e.g., couldn't there be a spontaneous reject corresponding to a request that has already made it into a state where it is unknown whether it has been processed or not?

mraszyk: I guess it's fine to refund cycles (everywhere) as long as we only accept at most a single refund for every downstream call.

derlerd-dfinity: Not sure I follow. If the receiver gets to see the request then it could technically accept the cycles. If we also refund them then we have created new cycles, no?

spec/index.md
```diff
@@ -3702,6 +3732,72 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+```
+
+State after
+
+```html
+S.call_contexts[ctxt_id].origin = FromCanister O with deadline = Expired
+S.messages = S.messages · ResponseMessage {
+        origin = FromCanister O with deadline = NoDeadline;
+        response = Reject (SYS_UNKNOWN, reject_msg);
+        refunded_cycles = 0;
```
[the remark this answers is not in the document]
alin-at-dfinity: Eventually and with a lot of extra logic we could do best-effort refunds. The key word(s) being "best-effort".

Here, we're producing a reject response while the call is still running (and may continue to do so for arbitrarily long). So the best we can do is not refund anything.

spec/index.md
oggy-dfin: The question was about unsynchronized subnet clocks and time along the call graph; there, we don't guarantee time monotonicity across canisters on the same call graph.

spec/index.md
oggy-dfin: The call can also be rejected with other codes (e.g., SYS_TRANSIENT) before the expiry; this transition is specific to SYS_UNKNOWN.

spec/index.md
oggy-dfin: The callee may still process a `CallMessage` even though it's expired. Furthermore, the callee may produce a response message even though the caller has already received a `SYS_UNKNOWN`.

spec/index.md
oggy-dfin: OK, I'll move the remark to the expiry transition.

spec/index.md
oggy-dfin: For the specific error codes above (SYS_FATAL, SYS_TRANSIENT, DESTINATION_INVALID), the receiver must not have received the request.

spec/index.md
oggy-dfin: That would be a weird spec; it would force the system to wait with producing a synthetic response until the stopping request that the canister issued times out, even though the canister already received a SYS_UNKNOWN response for that stopping request.

spec/index.md
derlerd-dfinity: I see makes sense - overlooked the restriction to certain error cases. 

Do we want to specify things this way? It might be a simplification worth considering to initially never refund cycles when we see errors at all and only start refunding in a later iteration.

spec/index.md
oggy-dfin: Intuitively, I'd expect this to be passed the same way that `arg` is, because it's defined by the caller (or by the system in case of ingress messages). But maybe I'm missing something?

spec/index.md
```diff
@@ -1550,6 +1564,16 @@ If this traps (e.g. runs out of cycles), the state changes from the `cleanup` fu
 
 There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` and `ic0.call_perform`.
 
+-   `ic0.call_with_best_effort_response : (timeout_seconds : i32) -> ()`
+
+    Relaxes the response delivery guarantee to be best effort, asking the system to respond at the latest after `timeout_seconds` have elapsed. Best effort means the system may also respond with a `SYS_UNKNOWN` reject code, signifying that the call **may or may not** have been processed by the callee. Then, even if the callee produces a response, it will not be delivered to the caller.
+
+    Any value for `timeout_seconds` is permitted, but is silently bounded from above by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The implementation may add a specific [error code](#error-codes) to a reject message to indicate the cause, in particular whether the timeout expired. Note that the reject callback may be executed (possibly significantly) later than the specified time (e.g., if the caller is under high load), or before timeout expiration (e.g., if the system is under load).
```
rsundar01, line 1571: > or before timeout expiration (e.g., if the system is under load)

Wondering why the reject callback would execute before the timeout when under load.

spec/index.md
```diff
@@ -3608,6 +3634,10 @@ then
           origin = FromCanister {
             call_context = M.call_context;
             callback = call.callback;
+            deadline = if call.timeout_seconds ≠ NoTimeout
+                        then S.time[M.receiver] + call.timeout_seconds
```
mraszyk, line 3638:
```suggestion
                        then S.time[M.receiver] + call.timeout_seconds * 1000000000
```

spec/index.md
mraszyk: > if the caller successfully calls the caller once, subsequent calls will not fail with `DESTINATION_INVALID`.

this might not be true if the callee canister is deleted

spec/index.md
```diff
@@ -3702,6 +3732,73 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
 Note that by construction, a query function will either trap or return with a response; it will never send calls, and it will never change the state of the canister.
 
+#### Spontaneous request rejection {#request-rejection}
+
+The system can reject a request at any point in time, e.g., because it is overloaded. The following specification is an overapproximation of the rejection behavior. In particular, the system is guaranteed not to emit `DESTINATION_INVALID` in some cases, e.g., if the callee exists, and if the caller successfully calls the caller once, subsequent calls will not fail with `DESTINATION_INVALID`.
+
+Condition:
+```html
+S.messages = Older_messages · CallMessage CM · Younger_messages
+(CM.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ M.queue)
```
mraszyk, line 3742:
```suggestion
(CM.queue = Unordered) or (∀ msg ∈ Older_messages. msg.queue ≠ CM.queue)
```

spec/index.md
mraszyk: > I bundled the fix to this with the best-effort PR since I was too lazy to handle merge conflicts :)

Could you please create a separate PR after all?

spec/index.md
```diff
@@ -3702,6 +3732,73 @@ The functions `query_as_update` and `system_task_as_update` turns a query functi
...
+        origin = CM.origin;
+        response = Reject (reject_code, reject_msg);
+        refunded_cycles = CM.transferred_cycles;
+    } ·
+    Younger_messages
+```
+
+#### Call expiry {#call-expiry}
+
+These transitions expire calls with timeouts. To account for SYS_UNKNOWN being issued early (e.g., due to high system load), we ignore the caller time in these transitions. We define two variants of the transition, one that expires messages, and one that expires calls that are in progress (i.e., have open downstream call contexts).
```
mraszyk, line 3760:
```suggestion
These transitions expire calls with best-effort responses. To account for SYS_UNKNOWN being issued early (e.g., due to high system load), we ignore the caller time in these transitions. We define two variants of the transition, one that expires messages, and one that expires calls that are in progress (i.e., have open downstream call contexts).
```

spec/index.md
```diff
@@ -3744,10 +3841,10 @@ Conditions
...
 
 S.call_contexts[Ctxt_id].needs_to_respond = false
-∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
-∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id
-∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ≠ Ctxt_id
-∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ≠ Ctxt_id
+∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
+∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
+∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
+∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ≠ Ctxt_id ∨ O.deadline = Expired
```
mraszyk, line 3847: Can't this lead to a call context being removed although some of its children are still alive (e.g., because they don't use best-effort responses)?

spec/index.md
mraszyk: > Intuitively, I'd expect this to be passed the same way that arg is, because it's defined by the caller (or by the system in case of ingress messages).

Sure, that's an option, too, but then the types of canister methods must be adjusted by one additional argument. If the deadline is put into Env, no changes to canister method types are necessary.

spec/index.md
mraszyk: also below

spec/index.md
oggy-dfin: Now that the deadline has been moved to origin, there won't be a large merge conflict if we do this in a separate PR, so I can do that.

spec/index.md
```diff
@@ -5064,9 +5161,11 @@ S.messages = Older_messages · ResponseMessage RM · Younger_messages
 RM.origin = FromCanister {
     call_context = Ctxt_id
     callback = Callback
+    deadline = D
```
oggy-dfin, line 5164: TODO: Add a condition here (and the transition below) to say that `ctxt_id \in dom(S.call_contexts)`; and also remove the invariant that we don't have "holes" in our call context graph

spec/index.md
```diff
@@ -127,6 +127,10 @@ This specification may refer to certain constants and limits without specifying
 
     Maximum wall clock time spent on evaluation of a query call.
 
+-   `MAX_CALL_TIMEOUT`
+
+    The maximum timeout (in seconds) for an inter-canister call.
```
dragoljub-duric, line 132: What is the reason we are introducing MAX_CALL_TIMEOUT?
I believe that implicitly bounding timeout can be problematic from a developer experience standpoint. Even more, since the current implementation of the message broadcast is AFAIK equivalent to having an infinite timeout, why we are disallowing that for best-effort messages?

dragoljub-duric: What do you think about returning an error if the timeout is larger than `MAX_CALL_TIMEOUT`, instead of implicitly bounding it?

dragoljub-duric: TBH, If I were a developer, and if I set a timeout on X seconds, and I received `SYS_UNKNOWN` after X/10 seconds (=`MAX_CALL_TIMEOUT`) I do not believe the first place I would look for a mistake would be interface spec of call_with_best_effort_response.

oggy-dfin: We could return an error on larger timeouts (IMO it would only make sense to return an error on `call_perform` rather than `call_with_best_effort_response`). But, for load shedding, we also want to allow the system to respond with a `SYS_UNKNOWN` even before the timeout elapses, no matter what the timeout is (even if it's below `MAX_CALL_TIMEOUT`). So the canister authors will have to deal with the case of a `SYS_UNKNOWN` being returned sooner, regardless of what we do with bounding the timeout parameter. This obviously needs to be documented well, not just in the spec, but I don't see it changing, otherwise the replica can't be elastic in how it handles load. Given this behavior, we opted not to return an error on exceeding `MAX_CALL_TIMEOUT`.

Another option was to add support for larger timeouts on a best-effort basis, i.e., we accept any timeout, but we reserve the right to invoke a callback earlier with SYS_UNKNOWN if we're under load (which we'd do anyways). But IMO that would complicate the reasoning (1) about keeping the replica state bounded (for example, we plan to support 10s of millions of messages, so we need to make sure that callbacks etc. don't take too much space) while (2) being fair among the canisters (e.g., if one canister has 9 million open calls, we should shed them even if their timeout is years in the feature). And I don't think it changes the situation much for the developers, since they have to deal with the "early reject" anyways.

dsarlis: I would like to point out that we have seen such decisions to come back and bite us -- essentially we build something that technically works fine and is well documented but people are surprised by unexpected behaviour or trapped in situations where they cannot differentiate between clearly different error cases that warrant different handling.

I think returning an error when passing what the system considers an invalid parameter (i.e. timeout larger than max allowed) is clearly something different than getting `SYS_UNKNOWN` because the system had to load shed. And people might want to know the difference between the two (similar situation we have with the reject codes where we have people come back to ask how can they differentiate between errors that clearly need different handling). In this case, the former (passing invalid parameter) can be clearly fixed by using an accepted timeout, the latter might well be outside your control (so not much you can do).

oggy-dfin: I'm not dead set against adding an error, but right now I don't think it's the better option.
1. You bring up distinguishing the two relevant error cases from the user perspective. I do actually agree that this is desirable, but I don't see how returning an error on overly large timeouts would solve this robustly. I think the only robust way to distinguish is by looking at the error message (a much better way, as we recently discussed, would be error codes, but that's another topic). Otherwise, what could the user do? They could store the expected deadline when issuing a request, and compare the time of the `SYS_UNKNOWN` reject with the deadline. But this is not robust. First, it could be that the `SYS_UNKNOWN` reject is issued before the deadline because of high load, but stuck in the input queue for a long time. So the user would wrongly conclude that the reject is due to timing out. Conversely (though less likely), with a time drift of even a few seconds, a remote subnet could reject the message early (compared to the local subnet's clock) with a timeout; unless we implement some more complicated machinery, the user could receive this reject before the deadline, breaking their logic again.
2. We have no established way for canisters to query any of the system limits, so I suppose the hard limit on timeouts would be no exception. So the user would have to again dig through the doc to find what the current limit is and hard code it in their canister.
3. In general, but especially given the previous point, we could never lower the max timeout without breaking backwards compatibility. Now sure, the hope is that we designed the whole thing well enough that it stays robust with a 5 minute max timeout, but it seems prudent to be careful.

dsarlis: I was thinking that the error when setting a higher than max timeout would be synchronous vs the asynchronous cases you mention (it's kinda similar to my output queue is full vs the receiving canister's input queue is full). However, I now remembered that actually users cannot differentiate between synchronous and asynchronous errors (which yet another topic in itself) so indeed there's probably not much the user can do and as such let's stick to what we have.

Maybe if we get traction on the other effort of distinguishing error codes we can revisit this one too.

alin-at-dfinity: Since the actual value of `MAX_CALL_TIMEOUT` is not specified, I could see `call_with_best_effort_response()` returning the actual, clamped timeout to give canister developers a hint of the actual outcome.

But returning an error and forcing everyone to deal with it seems to me like a lot more trouble than it's worth. Especially since, as Ogi pointed out, without a specified upper bound we might be breaking canisters if we adjust the bound downward.

In my mind, best-effort calls are just that: best-effort. Anything can go wrong and even if the system makes a good faith effort to differentiate between e.g. locally rejected requests (which are known to not have been delivered, so `SYS_TRANSIENT`) and timed out or shed ones, there is no guarantee that you'll ever get a vague `SYS_UNKNOWN` for your troubles.

Looking at this specific issue, even if the canister can reliably differentiate between a time out and an unceremoniously dropped request, there is no way for it to reliably act on it. For now, when we shed load, we simply drop the largest message. So the canister "dealing with it" might mean breaking down the payload into smaller chunks (assuming that's even a possibility). But in the future we may entirely change the load shedding logic and say "we prefer dropping messages that would be routed into heavily backlogged streams"; or whatever else. There is no way for a canister to deal with all the possibilities.

So I'm thinking we'll do a good faith attempt to differentiate between one and the other, but it's mostly going to be useful for metrics/stats ("X% of my calls time out and Y% are shed, so maybe I should focus on better sharding") as opposed to  "this specific call failed because it was load shed, let me try to resubmit it in smaller chunks".

oggy-dfin: Returning the accepted timeout sounds like a good idea. I don't immediately see how the Rust CDK/Motoko would export this given their current call return types, but I also don't really see anything speaking against it. There's still potential for confusion with interpreting the result (with all the previous caveats applying) but no worse than the current proposal. Wdyt @dragoljub-duric @dsarlis ? 

Re: error handling, I was thinking that the difference in error handling load shedding rejects vs timeout rejects could be "I'll back off from sending messages for a while until the system load subsides" vs "let's either resubmit or query the target whether the message arrived". But yeah, it might be more of a stretch, especially given the current lack of a `sleep` call.

oggy-dfin:
> @oggy-dfin Should we rephrase the title of this PR to something more descriptive?

Good point, changed now.

lwshang:
Can we consider renaming `call_with_best_effort_response: (timeout_seconds : i32) -> ()` to something like `call_timeout_set : (seconds : i32)`.

IMO, this will be more consistent with other `call_*` API's naming convention:
1. `call_` prefix indicates that this is for making inter-canister calls.
2. a noun specifies which perspective of the call is being set/modified, e.g. `cycles` in `call_cycles_add`.
3. a verb indicates what action is done to the none. e.g. `add` in `call_cycles_add`.

I believe "Best-Effort Response" is a suitable name for an abstract feature. However, for the specific API, I think `call_timeout_set` would be a clearer and more descriptive name.

spec/index.md
```diff
@@ -3088,12 +3114,12 @@ The following is an incomplete list of invariants that should hold for the abstr
...
+-   Referenced call contexts exist, unless the origins have expired deadlines:
 
-        ∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ∈ dom(S.call_contexts)
-        ∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.calling_context ∈ dom(S.call_contexts)
-        ∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.calling_context ∈ dom(S.call_contexts)
-        ∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.calling_context ∈ dom(S.call_contexts)
+        ∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.deadline ≠ Expired ⇒ O.calling_context ∈ dom(S.call_contexts)
+        ∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.deadline ≠ Expired ⇒ O.calling_context ∈ dom(S.call_contexts)
+        ∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.deadline ≠ Expired ⇒ O.calling_context ∈ dom(S.call_contexts)
+        ∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.deadline ≠ Expired ⇒ O.calling_context ∈ dom(S.call_contexts)
```
mraszyk, line 3122:
```suggestion
        ∀ CallMessage {origin = FromCanister O, …} ∈ S.messages. O.deadline ≠ Expired _ ⇒ O.calling_context ∈ dom(S.call_contexts)
        ∀ ResponseMessage {origin = FromCanister O, …} ∈ S.messages. O.deadline ≠ Expired _ ⇒ O.calling_context ∈ dom(S.call_contexts)
        ∀ (_ ↦ {needs_to_respond = true, origin = FromCanister O, …}) ∈ S.call_contexts: O.deadline ≠ Expired _ ⇒ O.calling_context ∈ dom(S.call_contexts)
        ∀ (_ ↦ Stopping Origins) ∈ S.canister_status: ∀(FromCanister O, _) ∈ Origins. O.deadline ≠ Expired _ ⇒ O.calling_context ∈ dom(S.call_contexts)
```

spec/index.md
```diff
@@ -6306,8 +6409,17 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
...
+        timeout_seconds = NoTimeout
       }
 
+    ic0.call_with_best_effort_response<es>(timeout_seconds : i32) =
+      if
+          es.context ∉ {U, CQ, Ry, Rt, CRy, CRt, T}
+          or es.pending_call = NoPendingCall
+          or es.pending_call.timeout ≠ NoTimeout
+      then Trap {cycles_used = es.cycles_used;}
+      es.pending_call.timeout_seconds := max(timeout_seconds, MAX_CALL_TIMEOUT)
```
mraszyk, line 6421:
```suggestion
      es.pending_call.timeout_seconds := min(timeout_seconds, MAX_CALL_TIMEOUT)
```

spec/index.md
```diff
@@ -1436,7 +1436,7 @@ The canister can access an argument. For `canister_init`, `canister_post_upgrade
 
     The deadline, in nanoseconds since 1970-01-01, after which the caller might stop waiting for a response.
 
-    For calls with best-effort responses, the deadline is computed based on the time the call was made, and the `timeout_seconds` parameter provided by the caller. For other calls, a deadline of 0 will be returned.
+    For calls to update methods with best-effort responses, the deadline is computed based on the time the call was made, and the `timeout_seconds` parameter provided by the caller. For other calls (ingress messages and all calls to query and composite query methods, including calls in replicated mode) a deadline of 0 is returned.
```
mraszyk, line 1439:
```suggestion
    For calls to update methods with best-effort responses and their callbacks, the deadline is computed based on the time the call was made, and the `timeout_seconds` parameter provided by the caller. For other calls (ingress messages and all calls to query and composite query methods, including calls in replicated mode) a deadline of 0 is returned.
```

mraszyk:
@oggy-dfin Could you please also resolve the conflicts?

spec/index.md
```diff
@@ -2781,7 +2781,7 @@ The [WebAssembly System API](#system-api) is relatively low-level, and some of i
             new_global_timer : NoGlobalTimer | Nat;
             cycles_used : Nat;
           }
-          update_methods : MethodName ↦ ((Arg, CallerId, Env, AvailableCycles) -> UpdateFunc)
+          update_methods : MethodName ↦ ((Arg, CallerId, Deadline, Env, AvailableCycles) -> UpdateFunc)
```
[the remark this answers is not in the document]
oggy-dfin: Done, thanks for spotting this

spec/index.md
```diff
@@ -1676,6 +1690,16 @@ If this traps (e.g. runs out of cycles), the state changes from the `cleanup` fu
 
 There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` and `ic0.call_perform`.
 
+-   `ic0.call_with_best_effort_response : (timeout_seconds : i32) -> ()`
+
+    Relaxes the response delivery guarantee to be best effort, asking the system to respond at the latest after `timeout_seconds` have elapsed. Best effort means the system may also respond with a `SYS_UNKNOWN` reject code, signifying that the call **may or may not** have been processed by the callee. Then, even if the callee produces a response, it will not be delivered to the caller.
+
+    Any value for `timeout_seconds` is permitted, but is silently bounded from above by the `MAX_CALL_TIMEOUT` system constant; i.e., larger timeouts are treated as equivalent to `MAX_CALL_TIMEOUT` and do not cause an error. The implementation may add a specific [error code](#error-codes) to a reject message to indicate the cause, in particular whether the timeout expired. Note that the reject callback may be executed (possibly significantly) later than the specified time (e.g., if the caller is under high load), or before timeout expiration (e.g., if the system is under load).
```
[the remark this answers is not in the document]
mraszyk: The callee can learn the timeout via `ic0.msg_deadline`.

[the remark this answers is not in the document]
oggy-dfin: I think Bjoern is referring to the `MAX_CALL_TIMEOUT`?
I agree that it's unfortunate, but it's no less unfortunate than all the other system constants, such as instruction limits, message size limits, (stable) memory size limits, the various costs etc etc. I think it doesn't make sense to do a one-off for this constant. But I'm not sure what a good general solution is. We have so far (for good reason, IMO) steered clear from returning structured data in API calls. Maybe with [WIT](https://component-model.bytecodealliance.org/design/wit.html) it's time to reconsider? Otherwise, having a `get_constant_X` system API call per constant is super yucky. I suppose we could have a single `get_constant(constant_name)`, but that would require a uniform return type.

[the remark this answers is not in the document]
mraszyk: @luc-blaeser wants to design a system API returning parameters such as remaining instructions, wasm heap etc. Maybe this would be an opportunity to reveal `MAX_CALL_TIMEOUT`.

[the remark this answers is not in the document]
Dfinity-Bjoern: Makes sense, let's include it in the limits.

Dfinity-Bjoern (approved):
Looks good!

mraszyk:
Superseded by https://github.com/dfinity/portal/pull/3764

PR #271, opened 2024-02-01
chore: add note on composite queries in HTTP gateway protocol

mraszyk:
This MR adds a note that composite query methods can be used instead of query methods in the HTTP gateway protocol and fixes a few typos/ambiguous formulations.

### DIFF
--- a/spec/http-gateway-protocol-spec.md
+++ b/spec/http-gateway-protocol-spec.md
@@ -14,8 +14,8 @@ An HTTP request by an HTTP client is handled by these steps:
 2. The HTTP Gateway intercepts the request.
 3. The HTTP Gateway resolves the canister ID that the request is intended for.
 4. The HTTP Gateway Candid encodes the HTTP request.
-5. The HTTP Gateway invokes the canister via a query call to the `http_request` interface.
-6. The canister handles the request and returns an HTTP response, encoded in Candid, together with additional metadata.
+5. The HTTP Gateway invokes the canister via a query call to the `http_request` canister method.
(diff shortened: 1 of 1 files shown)


spec/http-gateway-protocol-spec.md
```diff
@@ -448,6 +448,19 @@ You can also [download the file](./_attachments/http-gateway.did).
 
 Not all of this interface is required. The following sections detail what can be optionally omitted depending on the requirements of the canister in question.
 
+Note. Composite query methods can be used instead of query methods to allow for calling composite query methods of other canisters on the same subnet
+when processing an HTTP request, e.g., the canister can export
+
+```
+http_request: (request: HttpRequest) -> (HttpResponse) composite_query;
```
nathanosdev, line 455: Does this require any changes in the HTTP Gateway implementations to leverage this?

mraszyk: As long as the HTTP gateway implementations make HTTP requests to `/api/v2/.../query` to call this method, no changes are necessary. But note that HTTP requests to `/api/v2/.../call` for composite queries will be rejected (I'd expect the HTTP gateway to use the other method http_request_update if an update call is to be made).

nathanosdev: Thanks for the clarification. Yes, the HTTP Gateway won't make an update call to `http_request`, only to `http_request_update`.

nathanosdev (approved):

Dfinity-Bjoern (approved):

PR #272, opened 2024-02-01
fix: wrap chunk hash in a record and rename storage_canister to store_canister

mraszyk:
This PR wraps a chunk hash in the management canister API into a record with a single field containing the hash and renames `storage_canister` to `store_canister`.

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -201,7 +201,7 @@ type install_chunked_code_args = record {
         };
     };
     target_canister : canister_id;
-    storage_canister : opt canister_id;
+    store_canister : opt canister_id;
     chunk_hashes_list : vec chunk_hash;
     wasm_module_hash : blob;
     arg : blob;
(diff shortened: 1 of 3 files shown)


Dfinity-Bjoern:
I guess this is in order to reflect the interface that is actually implemented, but: Why?

adambratschikaye:
When dealing with the API, it's not so intuitive what the `blob` returned by `upload_chunk` represents. So putting it in a labeled record makes it more user friendly.

adambratschikaye:
^ merging that MR now.

lwshang:
Since the IC MR has been merged, we can merge this PR now, right?

lwshang:
There is still an inconsistent behavior between this interface-spec and the actual implementation.
The `chunk_hashes_list` field of `install_chunked_code`'s argument is different:
* In this spec: it is `vec chunk_hash` in which `type chunk_hash = record { hash : blob; };`
* In implementation: it is `vec vec nat8`

The implementation behavior was observed in Rust CDK CI using `ic-test-state-machine`. A sample failure can be found here:
https://github.com/dfinity/cdk-rs/actions/runs/8052621991/job/21993175798#step:5:1264

bogwar (approved):
thanks Martin

PR #273, opened 2024-02-01
fix: ic0.data_certificate system API access restrictions

mraszyk:
This MR fixes the access restrictions specified for the system API calls `ic0.data_certificate_{size,copy}` to be query and composite query methods. Since these calls always trap in any other method's context. Note that they also trap if `ic0.data_certificate_present()` returns 0, but this can't be encoded with the current access restriction matrix.


dfinity-berestovskyy (approved):
Thanks, Martin!

Dfinity-Bjoern (approved):
Makes sense to specify it in this way, thanks @mraszyk !

PR #275, opened 2024-02-08
add convenience types for canister management methods

### LINKED ISSUE #269: adapt canister management interface to include request and response types
the btc interface is nice and clean, each method has request and response types defined in the `did`
```
    bitcoin_get_balance: (get_balance_request) -> (satoshi);
    bitcoin_get_utxos: (get_utxos_request) -> (get_utxos_response);
```

the canister management methods are lacking those extra types
```
    create_canister : (record {
    settings : opt canister_settings;
    }) -> (record {canister_id : canister_id});
    update_settings : (record {
    canister_id : principal;
    settings : canister_settings;
    }) -> ();
    install_code : (record {
    mode : variant {install; reinstall; upgrade};
    canister_id : canister_id;
    wasm_module : wasm_module;
    arg : blob;
    }) -> ();
    uninstall_code : (record {
    canister_id : canister_id;
    }) -> ();
    start_canister : (record {canister_id : canister_id}) -> ();
    stop_canister : (record {canister_id : canister_id}) -> ();
    canister_status : (record {canister_id : canister_id}) -> (record {
        status : variant { running; stopping; stopped };
        settings: definite_canister_settings;
        module_hash: opt blob;
        memory_size: nat;
        cycles: nat;
    });
    canister_info : (record {
        canister_id : canister_id;
        num_requested_changes : opt nat64;
    }) -> (record {
        total_num_changes : nat64;
        recent_changes : vec change;
        module_hash : opt blob;
        controllers : vec principal;
    });
    delete_canister : (record {canister_id : canister_id}) -> ();
```

this means when i import the management canister into my motoko project like this
```
import ic "ic:aaaaa-aa";
```
i tend to create an extra file for the request and response types like this
```
	public type CanisterSettings = {
		freezing_threshold : Nat;
		controllers : [Principal];
		memory_allocation : Nat;
		compute_allocation : Nat;
	};
	public type CanisterStatus = {
		status : { #stopped; #stopping; #running };
		memory_size : Nat;
		cycles : Nat;
		settings : CanisterSettings;
		idle_cycles_burned_per_day : Nat;
		module_hash : ?Blob;
		reserved_cycles : Nat;
	};
```

this seems redundant. can we instead include those types in the `did` directly?

letmejustputthishere:
i formatted the document for a consistent style and added the missing convenience types to 
close #269

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -6,6 +6,7 @@ type canister_settings = record {
 	compute_allocation : opt nat;
 	memory_allocation : opt nat;
 	freezing_threshold : opt nat;
+	reserved_cycles_limit : opt nat;
 };
 
 type definite_canister_settings = record {
@@ -13,6 +14,7 @@ type definite_canister_settings = record {
 	compute_allocation : nat;
 	memory_allocation : nat;
 	freezing_threshold : nat;
+	reserved_cycles_limit : nat;
 };
 
 type change_origin = variant {
@@ -45,6 +47,8 @@ type change = record {
 	origin : change_origin;
 	details : change_details;
 };
+
+type chunk_hash = blob;
 
 type http_header = record { name : text; value : text };
 
@@ -111,12 +115,20 @@ type send_transaction_request = record {
 
 type millisatoshi_per_byte = nat64;
 
+type node_metrics = record {
+	node_id : principal;
+	num_blocks_total : nat64;
+	num_block_failures_total : nat64;
+};
+
 type create_canister_request = record {
 	settings : opt canister_settings;
 	sender_canister_version : opt nat64;
 };
 
-type create_canister_response = record { canister_id : canister_id }
+type create_canister_response = record {
+	canister_id : canister_id;
+};
 
 type update_settings_request = record {
 	canister_id : principal;
@@ -124,10 +136,45 @@ type update_settings_request = record {
 	sender_canister_version : opt nat64;
 };
 
+type upload_chunk_request = record {
+	canister_id : principal;
+	chunk : blob;
+};
+
+type clear_chunk_store_request = record {
+	canister_id : canister_id;
+};
+
+type stored_chunks_request = record {
+	canister_id : canister_id;
+};
+
 type install_code_request = record {
-	mode : variant { install; reinstall; upgrade };
+	mode : variant {
+		install;
+		reinstall;
+		upgrade : opt record {
+			skip_pre_upgrade : opt bool;
+		};
+	};
 	canister_id : canister_id;
 	wasm_module : wasm_module;
+	arg : blob;
+	sender_canister_version : opt nat64;
+};
+
+type install_chunked_code_request = record {
+	mode : variant {
+		install;
+		reinstall;
+		upgrade : opt record {
+			skip_pre_upgrade : opt bool;
+		};
+	};
+	target_canister : canister_id;
+	storage_canister : opt canister_id;
+	chunk_hashes_list : vec chunk_hash;
+	wasm_module_hash : blob;
 	arg : blob;
 	sender_canister_version : opt nat64;
 };
@@ -137,11 +184,17 @@ type uninstall_code_request = record {
 	sender_canister_version : opt nat64;
 };
 
-type start_canister_request = record { canister_id : canister_id }
+type start_canister_request = record {
+	canister_id : canister_id;
+};
 
-type stop_canister_request = record { canister_id : canister_id }
+type stop_canister_request = record {
+	canister_id : canister_id;
+};
 
-type canister_status_request = record { canister_id : canister_id }
+type canister_status_request = record {
+	canister_id : canister_id;
+};
 
 type canister_status_response = record {
 	status : variant { running; stopping; stopped };
@@ -149,6 +202,7 @@ type canister_status_response = record {
 	module_hash : opt blob;
 	memory_size : nat;
 	cycles : nat;
+	reserved_cycles : nat;
 	idle_cycles_burned_per_day : nat;
 };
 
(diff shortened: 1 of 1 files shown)


letmejustputthishere:
i formatted the document for a consistent style and added the missing convenience types to 
close #269

spec/_attachments/ic.did
```diff
@@ -2,244 +2,305 @@ type canister_id = principal;
...
+	uninstall_code : (uninstall_code_request) -> ();
+	start_canister : (start_canister_request) -> ();
+	stop_canister : (stop_canister_request) -> ();
+	canister_status : (canister_status_request) -> (canister_status_response);
+	canister_info : (canister_info_request) -> (canister_info_response);
+	delete_canister : (delete_canister_request) -> ();
+	deposit_cycles : (deposit_cycles_request) -> ();
+	raw_rand : () -> (blob);
+	http_request : (
+		record {
```
mraszyk, line 246: should we have a type synonym for this and some other records that are still inlined in the service definition? my preference would be to  introduce type synonyms for all records in the service

spec/_attachments/ic.did
```diff
@@ -2,244 +2,305 @@ type canister_id = principal;
 type wasm_module = blob;
 
 type canister_settings = record {
-  controllers : opt vec principal;
-  compute_allocation : opt nat;
-  memory_allocation : opt nat;
-  freezing_threshold : opt nat;
-  reserved_cycles_limit : opt nat;
+	controllers : opt vec principal;
```
mraszyk, line 5: Could we please stick with spaces instead of tabs (e.g., with 4 spaces instead of 2 spaces, but avoiding tabs)?

mraszyk (changes requested):
Thank you for polishing the ic.did file! I've left some comments.

letmejustputthishere:
thank you for reviewing! i implemented the requested changes

spec/_attachments/ic.did
```diff
@@ -2,115 +2,142 @@ type canister_id = principal;
...
+    network : bitcoin_network;
+    filter : opt variant {
+        min_confirmations : nat32;
+        page : blob;
+    };
 };
 
-type get_current_fee_percentiles_request = record {
-  network: bitcoin_network;
+type bitcoin_get_utxos_query_args = record {
```
mraszyk, line 99: Do we need to duplicate the type for query methods since they're exactly the same as for update methods?

spec/_attachments/ic.did
```diff
@@ -121,125 +148,216 @@ type node_metrics = record {
...
+    module_hash : opt blob;
+    memory_size : nat;
+    cycles : nat;
+    reserved_cycles : nat;
+    idle_cycles_burned_per_day : nat;
+};
+
+type canister_info_args = record {
+    canister_id : canister_id;
+    num_argsed_changes : opt nat64;
```
mraszyk, line 238:
```suggestion
    num_requested_changes : opt nat64;
```

mraszyk:
Thank you for implementing my comments. Left two more, but I'm happy to approve after the typo comment is resolved (i.e., the bitcoin types can be kept duplicated if you prefer).

spec/_attachments/ic.did
letmejustputthishere: i think  you're right and the chance that those two types diverge is probably 0, but for the sake of consistency i'd just keep this duplicate

letmejustputthishere:
this is ready to merge :)

mraszyk (approved):
LGTM Thank you! @letmejustputthishere

peterpeterparker:
While synchronizing the DID types in ic-js, I noticed unexpected format differences in `@dfinity/ic-management`, given that we sync the IC management types from this repo.

May I ask why the format was changed, what tool was used to format the DID files, and whether the format will be guaranteed to remain consistent for future iterations?

Capture d’écran 2024-02-26 à 11 57 33

peterpeterparker:
Thanks for the feedback. In the future, it would be great if we can remain consistent. It makes integration in ic-js a bit easier.

spec/_attachments/ic.did
peterpeterparker: What's the consistency reference? Is the source code of those really duplicated?

letmejustputthishere: > What's the consistency reference? Is the source code of those really duplicated?

that the parameters of each method have their own type definition named `<method_name>_args`.

peterpeterparker: Yeah ok but, is it so in the source code?

letmejustputthishere: I don't know, why do you think that matters?

letmejustputthishere: I case you haven't seen it, [this](https://github.com/dfinity/interface-spec/issues/269) was the motivation.

peterpeterparker: Because the did files should reflect the implementation not a pattern. So if in the source code both parameters are actually inheriting the same type, those should not be duplicated in the did file regarless of the "consistency" pattern. Therefore the source code matters.

Ultimately I'm also curious about this because it implies that I should duplicate code in ic-js.

peterpeterparker: It's the same type in the source code. See [implementation](https://github.com/dfinity/ic/blob/5fe907da2193c6051634b29133280f53f2490d52/rs/tests/src/execution/queries.rs#L136).

Therefore I think it's incorrect to duplicate the types in the specification. Agree?

letmejustputthishere: > Because the did files should reflect the implementation not a pattern.

i'd argue that the interface and the implementation can be viewed separately. e.g. i could have a public method that i don't want to add to my `did` file. but yes, in any case they should be in sync but i don't think that means they have to mirror each other regarding their structure. (this is a lightly held opinion :D)

peterpeterparker: I've implemented those two in this ic-js PR now awyway. https://github.com/dfinity/ic-js/pull/559
Thanks for the feedback.

PR #276, opened 2024-02-08
feat: add ICRC-21 API to management canister interface

frederikrothenberger:
Initially, (to keep the scope small) the interface will be restricted to calls related to canister management that can be called through ingress.

If there is demand from the community, more calls can be supported.

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -121,6 +121,39 @@ type node_metrics = record {
     num_block_failures_total : nat64;
 };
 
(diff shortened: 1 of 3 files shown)


Dfinity-Bjoern:
Current state from the interface spec meeting: Likely the best path forward is wait for the replicated query work to mature, and switch the ICRC-21 methods to query mode.

Dfinity-Bjoern:
Closing this for now since we want to wait for a more secure query mechanism – this should not go through update.

PR #277, opened 2024-02-09
feat: canister version guaranteed to increase if running status changes

mraszyk:
This PR specifies the canister version to be incremented if the running status (Running, Stopping, Stopped) of a canister changes. The motivation for this change is that a canister can observe its running status through ic0.canister_status system API and thus query caching in the replica implementation based on canister version is unsound if the canister version does not reflect changes to the canister's running status.


dfinity-berestovskyy (approved):
The replica side changes: https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/17590

Dfinity-Bjoern (approved):
Makes sense to me, thanks @mraszyk

PR #278, opened 2024-02-26
[FINAL] feat: WASM heap memory limit for canisters

mraszyk:
This PR introduces a configurable WASM heap memory limit for canisters.

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -7,6 +7,7 @@ type canister_settings = record {
     memory_allocation : opt nat;
     freezing_threshold : opt nat;
     reserved_cycles_limit : opt nat;
+    wasm_heap_memory_limit : nat;
 };
 
 type definite_canister_settings = record {
@@ -15,6 +16,7 @@ type definite_canister_settings = record {
     memory_allocation : nat;
     freezing_threshold : nat;
     reserved_cycles_limit : nat;
+    wasm_heap_memory_limit : nat;
 };
 
 type change_origin = variant {
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -7,7 +7,7 @@ type canister_settings = record {
     memory_allocation : opt nat;
     freezing_threshold : opt nat;
     reserved_cycles_limit : opt nat;
-    wasm_memory_limit : nat;
+    wasm_memory_limit : opt nat;
 };
 
 type definite_canister_settings = record {
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -7,7 +7,7 @@ type canister_settings = record {
     memory_allocation : opt nat;
     freezing_threshold : opt nat;
     reserved_cycles_limit : opt nat;
-    wasm_heap_memory_limit : nat;
+    wasm_memory_limit : nat;
 };
 
 type definite_canister_settings = record {
@@ -16,7 +16,7 @@ type definite_canister_settings = record {
     memory_allocation : nat;
     freezing_threshold : nat;
     reserved_cycles_limit : nat;
-    wasm_heap_memory_limit : nat;
+    wasm_memory_limit : nat;
 };
 
 type change_origin = variant {
--- a/spec/index.md
+++ b/spec/index.md
@@ -3553,13 +3553,13 @@ or
 ( M.entry_point = PublicMethod Name Caller Arg
   F = query_as_update(Mod.query_methods[Name], Arg, Caller, Env)
   New_canister_version = S.canister_version[M.receiver]
-  Wasm_memory_limit = 3_221_225_472
+  Wasm_memory_limit = 4_294_967_296
 )
 or
 ( M.entry_point = Callback Callback Response RefundedCycles
   F = Mod.callbacks(Callback, Response, RefundedCycles, Env, Available)
   New_canister_version = S.canister_version[M.receiver] + 1
-  Wasm_memory_limit = 3_221_225_472
+  Wasm_memory_limit = 4_294_967_296
 )
 or
 ( M.entry_point = Heartbeat
--- a/spec/index.md
+++ b/spec/index.md
@@ -1956,7 +1956,7 @@ The optional `settings` parameter can be used to set the following settings:
 
 -   `wasm_memory_limit` (`nat`)
 
-    Must be a number between 0 and 2<sup>32</sup>-1, inclusively, and indicates the upper limit on the WASM heap memory consumption of the canister.
+    Must be a number between 0 and 2<sup>48</sup>-1 (i.e., 256TB), inclusively, and indicates the upper limit on the WASM heap memory consumption of the canister.
 
     An operation (update method, global timer, heartbeat, canister init, canister post_upgrade) that causes the WASM heap memory consumption to exceed this limit will trap.
     The WASM heap memory limit is ignored for query methods, response callback handlers, and canister pre_upgrade.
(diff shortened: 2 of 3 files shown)


spec/index.md
```diff
@@ -3801,6 +3820,10 @@ if A.settings.reserved_cycles_limit is not null:
   New_reserved_balance_limit = A.settings.reserved_cycles_limit
 else:
   New_reserved_balance_limit = 5_000_000_000_000
+if A.settings.wasm_heap_memory_limit is not null:
+  New_wasm_heap_memory_limit = A.settings.wasm_heap_memory_limit
+else:
+  New_wasm_heap_memory_limit = 4_294_967_296
```
mraszyk, line 3826: What should be the default value for new canisters?

spec/index.md
```diff
@@ -1954,6 +1954,15 @@ The optional `settings` parameter can be used to set the following settings:
...
     Default value: 5_000_000_000_000 (5 trillion cycles).
 
+-   `wasm_heap_memory_limit` (`nat`)
+
+    Must be a number between 0 and 2<sup>32</sup>-1, inclusively, and indicates the upper limit on the WASM heap memory consumption of the canister.
+
+    An operation (update method, global timer, heartbeat, canister init, canister post_upgrade) that causes the WASM heap memory consumption to exceed this limit will trap.
+    The WASM heap memory limit is ignored for query methods, response callback handlers, and canister pre_upgrade.
+
+    Default value: 4_294_967_296 (4 GiB).
```
mraszyk, line 1964: It seems in line with reserved_cycles_limit to always have a limit and just set it to a high value (e.g., 4GiB) to effectively disable the limit.

spec/_attachments/ic.did
```diff
@@ -7,6 +7,7 @@ type canister_settings = record {
     memory_allocation : opt nat;
     freezing_threshold : opt nat;
     reserved_cycles_limit : opt nat;
+    wasm_heap_memory_limit : nat;
```
ulan, line 10: Let's go with `wasm_memory_limit` which is shorted and would allow us to add `stable_memory_limit` in the future. (Sorry if it was not clear from the ticket).

spec/index.md
ulan: Let's set the default value to 3GiB to keep the spec simple.

The actual rollout will be more smart about it:
- if the canister memory usage is below 2.5GiB, then we set the value to 3GiB.
- otherwise, we set the value to min(current_memory + 0.5GiB, 4GiB).
^ Does it make sense?

ulan (approved):
LGTM, thanks a lot @mraszyk

spec/index.md
mraszyk: Yeah, that makes sense. I didn't specify it though since it's a one-off action.

Dfinity-Bjoern (approved):

spec/index.md
```diff
@@ -1954,6 +1954,15 @@ The optional `settings` parameter can be used to set the following settings:
 
     Default value: 5_000_000_000_000 (5 trillion cycles).
 
+-   `wasm_memory_limit` (`nat`)
+
+    Must be a number between 0 and 2<sup>32</sup>-1, inclusively, and indicates the upper limit on the WASM heap memory consumption of the canister.
```
ulan, line 1959: @mraszyk: would you mind adjusting this limit to 2^48 (256TB) similar to the exiting `memory_allocation` field. The reason: there is an upcoming wasm64 feature which will allow the wasm memory to be larger than 4GiB.

ulan:
@mraszyk: could you please merge this PR? The replica supports the feature now. Rust CDK and agent changes are blocked by this spec PR being merged according to the SDK team.

spec/index.md
mraszyk: How about the default value (currently 3GiB)? And how about canisters that do not declare a 64-bit memory - there using limits and defaults for the 64-bit case doesn't make much sense, right?

ulan: For 64-bit canisters we will still have the limit on the replica side (initially it will be 6 or 8 GiB). We can set the default value based on that. Instead of hardcoding the default in the spec, would it make sense to use a named parameter?

mraszyk: Discussed offline.

spec/_attachments/ic.did
```diff
@@ -7,6 +7,7 @@ type canister_settings = record {
     memory_allocation : opt nat;
     freezing_threshold : opt nat;
     reserved_cycles_limit : opt nat;
+    wasm_memory_limit : nat;
```
mraszyk, line 10:
```suggestion
    wasm_memory_limit : opt nat;
```
let's double check

spec/_attachments/ic.did
```diff
@@ -15,6 +16,7 @@ type definite_canister_settings = record {
     memory_allocation : nat;
     freezing_threshold : nat;
     reserved_cycles_limit : nat;
+    wasm_memory_limit : nat;
```
mraszyk, line 19:
```suggestion
    wasm_memory_limit : opt nat;
```
???

spec/index.md
```diff
@@ -1954,6 +1954,15 @@ The optional `settings` parameter can be used to set the following settings:
...
     Default value: 5_000_000_000_000 (5 trillion cycles).
 
+-   `wasm_memory_limit` (`nat`)
+
+    Must be a number between 0 and 2<sup>48</sup>-1 (i.e., 256TB), inclusively, and indicates the upper limit on the WASM heap memory consumption of the canister.
+
+    An operation (update method, global timer, heartbeat, canister init, canister post_upgrade) that causes the WASM heap memory consumption to exceed this limit will trap.
+    The WASM heap memory limit is ignored for query methods, response callback handlers, and canister pre_upgrade.
+
+    Default value: 3_221_225_472 (3 GiB).
```
mraszyk, line 1964: There doesn't seem to be a default value in the implementation and an unspecified limit is returned as 0 in the canister status.

spec/index.md
```diff
@@ -1954,6 +1954,15 @@ The optional `settings` parameter can be used to set the following settings:
 
     Default value: 5_000_000_000_000 (5 trillion cycles).
 
+-   `wasm_memory_limit` (`nat`)
+
+    Must be a number between 0 and 2<sup>48</sup>-1 (i.e., 256TB), inclusively, and indicates the upper limit on the WASM heap memory consumption of the canister.
```
mraszyk, line 1959: Is the range validated somewhere in the code?

mraszyk:
@ulan There seem to be some mismatches between the spec and the implementation - could you please take a look at my above comments?

spec/index.md
```diff
@@ -1954,6 +1954,15 @@ The optional `settings` parameter can be used to set the following settings:
...
+
+    An operation (update method, global timer, heartbeat, canister init, canister post_upgrade) that causes the WASM heap memory consumption to exceed this limit will trap.
+    The WASM heap memory limit is ignored for query methods, response callback handlers, and canister pre_upgrade.
```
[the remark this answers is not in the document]
ulan: That will be implemented once canister logging ships:
                // The Wasm memory limit is not enforced in system tasks (timers
                // and heartbeats) until canister logging is implemented.
                // Without canister logging developers do not get error messages
                // from system tasks.
                // TODO(RUN-957): Enforce the limit after canister logging ships.

spec/index.md
ulan: The default is not set yet. The will be the final MR on the implementation side.

spec/index.md
ulan: Good catch! The implementation currently allows 2^64. I followed the memory_allocation validation, but that enforces the 2^48 implicitly via the subnet available memory.

ulan: I'll fix it.

PR #279, opened 2024-02-28
fix: RUN-926: Re-classify `canister-out-of-cycles` errors as `transient`

dfinity-berestovskyy:
When canister balance goes below the freezing threshold, any attempt to execute an update or a query call returns the `canister-out-of-cycles` error. At the moment this error is classified as a `canister error`, just like the `canister-trap` error. The CDK timers library does not retry the timer handler execution on `canister error`, and hence the timer handler execution might be skipped when canister goes below the freezing threshold.

This MR re-classifies the `canister-out-of-cycles` error as a `transient error`, just like the `canister-queue-full` error. The timers CDK library retries the timer handler execution on transient errors. So with this change, the timer handler execution should be retried, once the canister balance is above the freezing threshold.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -285,7 +285,7 @@ This status is orthogonal to whether a canister is empty or not: an empty canist
 
 :::note
 
-This status is orthogonal to whether a canister is frozen or not: a frozen canister can be in status `running`. Calls to such a canister are still rejected by the IC, but because the canister is frozen.
+This status is orthogonal to whether a canister is frozen or not: a frozen canister can be in status `running`. Calls to such a canister are still rejected by the IC, but because the canister is frozen, the returned reject code is `SYS_TRANSIENT`.
 
 :::
 
@@ -1942,7 +1942,7 @@ The optional `settings` parameter can be used to set the following settings:
 
     A canister is considered frozen whenever the IC estimates that the canister would be depleted of cycles before `freezing_threshold` seconds pass, given the canister's current size and the IC's current cost for storage.
 
-    Calls to a frozen canister will be rejected (like for a stopping canister). Additionally, a canister cannot perform calls if that would, due the cost of the call and transferred cycles, would push the balance into frozen territory; these calls fail with `ic0.call_perform` returning a non-zero error code.
+    Calls to a frozen canister will be rejected with `SYS_TRANSIENT` reject code. Additionally, a canister cannot perform calls if that would, due the cost of the call and transferred cycles, would push the balance into frozen territory; these calls fail with `ic0.call_perform` returning a non-zero error code.
 
     Default value: 2592000 (approximately 30 days).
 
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -3327,6 +3327,19 @@ messages = Older_messages · Younger_messages  ·
 
 ```
 
+But a call to a frozen canister is rejected with `SYS_TRANSIENT` reject code:
```
mraszyk, line 3330: Could we please split up the transition into two transitions tearing
```
S.canister_status[CM.callee] = Stopped or S.canister_status[CM.callee] = Stopping _ or liquid_balance(
```
into
```
S.canister_status[CM.callee] = Stopped or S.canister_status[CM.callee] = Stopping _
```
in one transition and
```
liquid_balance(...
```
in the other one?

spec/index.md
```diff
@@ -5410,21 +5423,23 @@ We define an auxiliary method that handles calls from composite query methods by
...
+             S.freezing_threshold[Canister_id],
+             memory_usage_wasm_state(S.canisters[Canister_id].wasm_state) +
+               memory_usage_raw_module(S.canisters[Canister_id].raw_module) +
+               memory_usage_canister_history(S.canister_history[Canister_id]),
+             S.canister_subnet[Canister_id].subnet_size,
+           )
+         ) < 0
+      then
+         Return (Reject (SYS_TRANSIENT, <implementation-specific>), Cycles)
       if S.canisters[Canister_id] ≠ EmptyCanister and
```
mraszyk, line 5441:
```suggestion
      else if S.canisters[Canister_id] ≠ EmptyCanister and
```

spec/index.md
```diff
@@ -5410,21 +5423,23 @@ We define an auxiliary method that handles calls from composite query methods by
                   status = simple_status(S.canister_status[Canister_id]);
                   canister_version = S.canister_version[Canister_id];
                 }
+      if liquid_balance(
```
mraszyk, line 5426: Does this condition really take precedence over checking if the canister is running etc.?

dfinity-berestovskyy: The implementation is completely different, see `fn evaluate_query_call_graph` I've seen this and raised this with the team. We've decided that the exact error order is hard to express and not that important.

dfinity-berestovskyy: But I'll move the `liquid_balance` closer to it's actual location in the code...

mraszyk: > We've decided that the exact error order is hard to express and not that important.

In that case, we should consider leaving it out from the spec.

dfinity-berestovskyy: We've discussed this option as well. This high-level algorithm might be still useful, but I don't have a strong opinion...

mraszyk: Let's keep this as is.

mraszyk:
One more ask: could you please update the changelog?

mraszyk (approved):

PR #281, opened 2024-03-12
[FINAL] feat: wasm_memory_persistence option for canister upgrades

mraszyk:
This PR adds an option (called `wasm_memory_persistence`) for canister upgrades. If this option is set, then the WebAssembly memory of the old canister is preserved across the upgrade.

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -182,7 +182,10 @@ type install_code_args = record {
         reinstall;
         upgrade : opt record {
             skip_pre_upgrade : opt bool;
-            keep_main_memory : opt bool;
+            wasm_memory_persistence : opt variant {
+                keep;
+                drop;
+            };
         };
     };
     canister_id : canister_id;
@@ -197,7 +200,10 @@ type install_chunked_code_args = record {
         reinstall;
         upgrade : opt record {
             skip_pre_upgrade : opt bool;
-            keep_main_memory : opt bool;
+            wasm_memory_persistence : opt variant {
+                keep;
+                drop;
+            };
         };
     };
     target_canister : canister_id;
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -182,6 +182,7 @@ type install_code_args = record {
         reinstall;
         upgrade : opt record {
             skip_pre_upgrade : opt bool;
+            keep_main_memory : opt bool;
         };
     };
     canister_id : canister_id;
@@ -196,6 +197,7 @@ type install_chunked_code_args = record {
         reinstall;
         upgrade : opt record {
             skip_pre_upgrade : opt bool;
+            keep_main_memory : opt bool;
         };
     };
     target_canister : canister_id;
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -185,7 +185,7 @@ type canister_install_mode = variant {
         skip_pre_upgrade : opt bool;
         wasm_memory_persistence : opt variant {
             keep;
-            drop;
+            replace;
         };
     };
 };
--- a/spec/index.md
+++ b/spec/index.md
@@ -4355,7 +4355,7 @@ Only the controllers of the given canister can install new code. This changes th
 
 In the following, the `initial_wasm_store` is the store of the WebAssembly module after instantiation (as per WebAssembly spec) of the WebAssembly module contained in `A.wasm_module`, before executing a potential `(start)` function. The store `initialize_store(State, A.wasm_module)` is the store of the WebAssembly module after instantiation (as per WebAssembly spec) of the WebAssembly module contained in `A.wasm_module` while reusing the WebAssembly memory of `State`.
 
-If the canister exports a private custom section with the name "enhanced-orthogonal-persistence", then the `wasm_memory_persistence` option must be set to `opt keep`.
+If the old canister module exports a private custom section with the name "enhanced-orthogonal-persistence", then the `wasm_memory_persistence` option must be set to `opt keep`.
 
 Conditions  
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -1235,7 +1235,7 @@ When a canister is upgraded to a new WebAssembly module, the IC:
 
 3.  Invokes `canister_post_upgrade` (if present) on the new instance, passing the `arg` provided in the `install_code` call ([IC method](#ic-install_code)).
 
-The stable memory is preserved throughout the process; any other WebAssembly state is discarded unless `wasm_memory_persistence` is `opt keep`.
+The stable memory is preserved throughout the process; the WebAssembly heap memory is discarded unless `wasm_memory_persistence` is `opt keep`.
 
 During these steps, no other entry point of the old or new canister is invoked. The `canister_init` function of the new canister is *not* invoked.
 
@@ -2012,7 +2012,7 @@ Only controllers of the canister can install code.
 
     Note that this is different from `uninstall_code` followed by `install_code`, as `uninstall_code` generates a synthetic reject response to all callers of the uninstalled canister that the uninstalled canister did not yet reply to and ensures that callbacks to outstanding calls made by the uninstalled canister won't be executed (i.e., upon receiving a response from a downstream call made by the uninstalled canister, the cycles attached to the response are refunded, but no callbacks are executed).
 
--   If `mode = variant { upgrade }` or `mode = variant { upgrade = opt record { skip_pre_upgrade = .., wasm_memory_persistence = .. } }`, this will perform an upgrade of a non-empty canister as described in [Canister upgrades](#system-api-upgrades), passing `arg` to the `canister_post_upgrade` method of the new instance. If `skip_pre_upgrade = opt true`, then the `canister_pre_upgrade` method on the old instance is not executed. If `wasm_memory_persistence = opt keep`, then the WebAssembly state is preserved.
+-   If `mode = variant { upgrade }` or `mode = variant { upgrade = opt record { skip_pre_upgrade = .., wasm_memory_persistence = .. } }`, this will perform an upgrade of a non-empty canister as described in [Canister upgrades](#system-api-upgrades), passing `arg` to the `canister_post_upgrade` method of the new instance. If `skip_pre_upgrade = opt true`, then the `canister_pre_upgrade` method on the old instance is not executed. If `wasm_memory_persistence = opt keep`, then the WebAssembly heap memory is preserved.
 
 This is atomic: If the response to this request is a `reject`, then this call had no effect.
 
@@ -4351,7 +4351,7 @@ S with
 
 #### IC Management Canister: Code upgrade
 
-Only the controllers of the given canister can install new code. This changes the code of an *existing* canister, preserving the state in the stable memory. This involves invoking the `canister_pre_upgrade` method, if the `skip_pre_upgrade` flag is not set to `opt true`, on the old and `canister_post_upgrade` method on the new canister, which must succeed and must not invoke other methods. If the `wasm_memory_persistence` flag is set to `opt keep`, then the WebAssembly state is preserved.
+Only the controllers of the given canister can install new code. This changes the code of an *existing* canister, preserving the state in the stable memory. This involves invoking the `canister_pre_upgrade` method, if the `skip_pre_upgrade` flag is not set to `opt true`, on the old and `canister_post_upgrade` method on the new canister, which must succeed and must not invoke other methods. If the `wasm_memory_persistence` flag is set to `opt keep`, then the WebAssembly heap memory is preserved.
 
 In the following, the `initial_wasm_store` is the store of the WebAssembly module after instantiation (as per WebAssembly spec) of the WebAssembly module contained in `A.wasm_module`, before executing a potential `(start)` function. The store `initialize_store(State, A.wasm_module)` is the store of the WebAssembly module after instantiation (as per WebAssembly spec) of the WebAssembly module contained in `A.wasm_module` while reusing the WebAssembly memory of `State`.
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -1229,13 +1229,13 @@ The IC assumes the canister to be fully instantiated if the `canister_init` meth
 
 When a canister is upgraded to a new WebAssembly module, the IC:
 
-1.  Invokes `canister_pre_upgrade` (if exported by the current canister code and  `skip_pre_upgrade` is not `opt true` ) on the old instance, to give the canister a chance to clean up (e.g. move data to [stable memory](#system-api-stable-memory)).
+1.  Invokes `canister_pre_upgrade` (if exported by the current canister code and `skip_pre_upgrade` is not `opt true`) on the old instance, to give the canister a chance to clean up (e.g. move data to [stable memory](#system-api-stable-memory)).
 
 2.  Instantiates the new module, including the execution of `(start)`, with a fresh WebAssembly state.
 
 3.  Invokes `canister_post_upgrade` (if present) on the new instance, passing the `arg` provided in the `install_code` call ([IC method](#ic-install_code)).
 
-The stable memory is preserved throughout the process; any other WebAssembly state is discarded.
+The stable memory is preserved throughout the process; any other WebAssembly state is discarded unless `keep_main_memory` is `opt true` in which case only WebAssembly data segments are discarded.
 
 During these steps, no other entry point of the old or new canister is invoked. The `canister_init` function of the new canister is *not* invoked.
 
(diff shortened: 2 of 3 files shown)


spec/_attachments/ic.did
```diff
@@ -182,6 +182,7 @@ type install_code_args = record {
         reinstall;
         upgrade : opt record {
             skip_pre_upgrade : opt bool;
+            keep_main_memory : opt bool;
```
mraszyk, line 185:
```suggestion
            keep_wasm_memory : opt bool;
```
to be consistent with https://github.com/dfinity/interface-spec/pull/278

spec/index.md
```diff
@@ -1229,13 +1229,13 @@ The IC assumes the canister to be fully instantiated if the `canister_init` meth
...
 
-1.  Invokes `canister_pre_upgrade` (if exported by the current canister code and  `skip_pre_upgrade` is not `opt true` ) on the old instance, to give the canister a chance to clean up (e.g. move data to [stable memory](#system-api-stable-memory)).
+1.  Invokes `canister_pre_upgrade` (if exported by the current canister code and `skip_pre_upgrade` is not `opt true`) on the old instance, to give the canister a chance to clean up (e.g. move data to [stable memory](#system-api-stable-memory)).
 
 2.  Instantiates the new module, including the execution of `(start)`, with a fresh WebAssembly state.
 
 3.  Invokes `canister_post_upgrade` (if present) on the new instance, passing the `arg` provided in the `install_code` call ([IC method](#ic-install_code)).
 
-The stable memory is preserved throughout the process; any other WebAssembly state is discarded.
+The stable memory is preserved throughout the process; any other WebAssembly state is discarded unless `wasm_memory_persistence` is `opt keep` in which case only WebAssembly data segments are discarded.
```
luc-blaeser, line 1238: A detail aspect: The active data segments of the initial program version would still be accessible as the entire main memory is retained. For the new program version, the active data segments are not loaded. This is why passive data segments are currently used for the orthogonal persistence.
Maybe one could phrase it as: "in which case the main Wasm memory is additionally preserved.". 
Maybe with additional information: "If `wasm_memory_peristence is `opt keep`, active data segments of the new program version are not loaded."

spec/index.md
```diff
@@ -4353,7 +4351,9 @@ S with
 
 #### IC Management Canister: Code upgrade
 
-Only the controllers of the given canister can install new code. This changes the code of an *existing* canister, preserving the state in the stable memory. This involves invoking the `canister_pre_upgrade` method, if the `skip_pre_upgrade` flag is not set to `opt true`, on the old and `canister_post_upgrade` method on the new canister, which must succeed and must not invoke other methods.
+Only the controllers of the given canister can install new code. This changes the code of an *existing* canister, preserving the state in the stable memory. This involves invoking the `canister_pre_upgrade` method, if the `skip_pre_upgrade` flag is not set to `opt true`, on the old and `canister_post_upgrade` method on the new canister, which must succeed and must not invoke other methods. If the `wasm_memory_persistence` flag is set to `opt keep`, then the WebAssembly state is preserved except for WebAssembly data segments.
```
luc-blaeser, line 4354: Maybe: 
```suggestion
Only the controllers of the given canister can install new code. This changes the code of an *existing* canister, preserving the state in the stable memory. This involves invoking the `canister_pre_upgrade` method, if the `skip_pre_upgrade` flag is not set to `opt true`, on the old and `canister_post_upgrade` method on the new canister, which must succeed and must not invoke other methods. If the `wasm_memory_persistence` flag is set to `opt keep`, then the WebAssembly memory state is preserved and data segments of the new WebAssembly code need to be explicitly loaded by using passive data segments.
```

spec/index.md
```diff
@@ -4400,19 +4400,30 @@ Env = {
...
 )
 
+(
+  (A.mode = upgrade U and U.wasm_memory_persistence ≠ keep)
+  Intermediate_state = {store = initial_wasm_store; self_id = A.canister_id; stable_mem = Stable_memory}
+)
+or
+(
+  (A.mode = upgrade U and U.wasm_memory_persistence = keep)
+  Intermediate_state = {store = load_data_segments(Old_state.store, A.wasm_module); self_id = A.canister_id; stable_mem = Stable_memory}
```
luc-blaeser, line 4416: For simplicity, we thought not to load the new active data segments and not to patch the memory (as this would need a reserved space and is also more complicated for the IC execution layer). Instead, we can use passive data segments that can be lazily loaded at a dynamic address.

It would be more like:
```suggestion
  Intermediate_state = {store = Old_state.store; self_id = A.canister_id; stable_mem = Stable_memory}
```

spec/index.md
```diff
@@ -4353,7 +4351,9 @@ S with
...
+Only the controllers of the given canister can install new code. This changes the code of an *existing* canister, preserving the state in the stable memory. This involves invoking the `canister_pre_upgrade` method, if the `skip_pre_upgrade` flag is not set to `opt true`, on the old and `canister_post_upgrade` method on the new canister, which must succeed and must not invoke other methods. If the `wasm_memory_persistence` flag is set to `opt keep`, then the WebAssembly state is preserved except for WebAssembly data segments.
+
+In the following, the `initial_wasm_store` is the store of the WebAssembly module after *instantiation* (as per WebAssembly spec) of the WebAssembly module contained in the [canister module](#canister-module-format), before executing a potential `(start)` function. The function `load_data_segments(S, wasm_module)` returns the WebAssembly store `S` after loading the data segments from the given WebAssembly module.
```
luc-blaeser, line 4356: We could probably skip the `load_data_segments()`, please cf. below.

spec/index.md
mraszyk: In that case, I'd go ahead and simplify the spec to not talk about data segments at all. Does that sound good?

spec/_attachments/ic.did
```diff
@@ -182,6 +182,10 @@ type install_code_args = record {
         reinstall;
         upgrade : opt record {
             skip_pre_upgrade : opt bool;
+            wasm_memory_persistence : opt variant {
+                keep;
+                drop;
+            };
```
crusso, line 188: Good idea to have a variant. Then we can add future options (eg. to handle multiple memories selectively)

crusso: Would it make sense to add type
```
type Mode = variant {...}
```
and use it twice, rather than duplicating. Or maybe that breaks existing (Rust) clients.

crusso (commented):
LGTM

spec/_attachments/ic.did
luc-blaeser: Maybe we could even sub-summarize the entire upgrade options as an intermediate type:

```
type CanisterUpgradeOptions = record {
     skip_pre_upgrade : opt bool;
     wasm_memory_persistence : opt variant {
        keep;
        drop;
     };
};
```

mraszyk: The duplication is due to having install_code and install_chunked_code: I plan to address it separately.

mraszyk: https://github.com/dfinity/interface-spec/pull/282

mraszyk:
@luc-blaeser Is it important to distinguish between "WebAssembly state" and "WebAssembly memory state"? I saw you suggested inserting "memory" and I wonder why you made the suggestion.

luc-blaeser:
> @luc-blaeser Is it important to distinguish between "WebAssembly state" and "WebAssembly memory state"? I saw you suggested inserting "memory" and I wonder why you made the suggestion.

It is not important. I only thought that the binary will be replaced (with its contained data segments).

luc-blaeser (approved):
Thanks a lot Martin. This looks good.

ulan:
Redirecting the review to @eust-dfinity and @alexandru-uta

spec/index.md
```diff
@@ -4391,28 +4391,39 @@ Env = {
...
 )
 
+(
+  (A.mode = upgrade U and U.wasm_memory_persistence ≠ keep)
+  Persisted_state = {store = initial_wasm_store; self_id = A.canister_id; stable_mem = Intermediate_state.stable_memory}
+)
+or
+(
+  (A.mode = upgrade U and U.wasm_memory_persistence = keep)
+  Persisted_state = Intermediate_state
```
mraszyk, line 4416:
```suggestion
  Persisted_state = initialize_stote(Intermediate_state, A.wasm_module)
```
where `initialize_store(State, Wasm_module)` is the store of the WebAssembly module after *instantiation* (as per WebAssembly spec) of the WebAssembly module `Wasm_module` reusing the WebAssembly memory of `State`.

eust-dfinity (approved):

spec/index.md
mraszyk: Done without commenting on active data segments. @crusso Could you please comment on whether special treatment of active data segments is needed after all?

crusso: @luc-blaeser and I discussed this offline and perhaps the simplest to specify is to disallow active data segments for EOP binaries - i.e, when keep is true. The motoko compiler already only emits passive data segments, rewriting the active data segment of the RTS to a passive one off-line.

Once could relax the restriction of no active data segments later  on in the spec and implementation.

mraszyk: > disallow active data segments for (replacement) EOP (only) binaries - i.e, when keep is true

Why would it be simpler to disallow them rather than not commenting on them at all (meaning if the new module provides active data segments, then they'll overwrite parts of the persisted Wasm memory)?

luc-blaeser: I had a corresponding prototype implementation, but there was a concern on complexity (and performance) of the upgrade that we have to selectively patch the active data segments. I believe a reason is that for the initial memory setup, the data segments can be applied in a simple page-based granularity (`Segments::as_pages()` extending the data segments to pages with zero initialization of the unused parts), while for the upgrade, the patching would need to be happen on the existing pages in the more fine-granular byte-wise approach.

The main reason was however that it is not very useful because the runtime system would need to reserve a fix section for active data segments in order not accidentally overwrite the heap. Choosing the size of the reserved space would be very difficult, it is often too wasteful or too limited. This is why we only use passive data segments, as it gives us the flexibility to choose the addresses dynamically in a simple way.

spec/index.md
```diff
@@ -4391,28 +4391,39 @@ Env = {
 }
 
 (
-  (A.mode = upgrade or A.mode = upgrade {skip_pre_upgrade = false})
+  (A.mode = upgrade U and U.skip_pre_upgrade ≠ true)
```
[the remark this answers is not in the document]
mraszyk: > Are there some Candid options missing here?

What do you mean by "missing"? In this condition, only skip_pre_uprade is relevant and that's why the other options do not appear explicitly (they aren't read from the record U).

spec/index.md
```diff
@@ -4391,28 +4393,43 @@ Env = {
...
```
crusso, line 4396: What I mean is, assuming A is the textual candid argument, isn't the correct condition (taking account of the defaults) something like
```suggestion
  (A.mode = variant { upgrade = 0 } and 0 = null or (0 = opt U and  U.skip_pre_upgrade ≠ opt true))
```
But that is really ugly...

crusso: Or is there some convention that let's you finesse this?

spec/index.md
```diff
@@ -5872,7 +5889,7 @@ Finally we can specify the abstract `CanisterModule` that models a concrete WebA
 
     If the WebAssembly module does not export a function called under the name `canister_pre_upgrade`, then it simply returns the stable memory:
```
crusso, line 5890:
```suggestion
    If the WebAssembly module does not export a function called under the name `canister_pre_upgrade`, then it simply returns the current state:
```

spec/index.md
```diff
@@ -4391,28 +4393,43 @@ Env = {
...
     canister_version = S.canister_version[A.canister_id];
   }
-  Old_module.pre_upgrade(Old_State, M.caller, Env1) = Return {stable_memory = Stable_memory; new_certified_data = New_certified_data; cycles_used = Cycles_used;}
+  Old_module.pre_upgrade(Old_State, M.caller, Env1) = Return {new_state = Intermediate_state; new_certified_data = New_certified_data; cycles_used = Cycles_used;}
 )
 or
 (
-  A.mode = upgrade {skip_pre_upgrade = true}
-  Stable_memory = Old_State.stable_mem
+  (A.mode = upgrade U and U.skip_pre_upgrade = true)
```
crusso, line 4405:
```suggestion
  (A.mode = variant { upgrade = opt U } and U.skip_pre_upgrade = opt true)
```
Perhaps?

spec/index.md
```diff
@@ -4391,28 +4393,43 @@ Env = {
...
+  Intermediate_state = Old_state
   New_certified_data = NoCertifiedData
   Cycles_used = 0
 )
 
+(
+  (A.mode = upgrade U and U.wasm_memory_persistence ≠ keep)
```
crusso, line 4412:
```suggestion
  (A.mode = variant { upgrade = O } and O = null or O = opt U and U.wasm_memory_persistence ≠ opt variant { keep })
```
Yuck

spec/index.md
```diff
@@ -4391,28 +4393,43 @@ Env = {
...
+  Persisted_state = {store = initial_wasm_store; self_id = A.canister_id; stable_mem = Intermediate_state.stable_memory}
+)
+or
+(
+  (A.mode = upgrade U and U.wasm_memory_persistence = keep)
```
crusso, line 4417:
```suggestion
  (A.mode = variant { upgrade = opt U } and U.wasm_memory_persistence = opt keep)
```

spec/index.md
```diff
@@ -4391,28 +4393,43 @@ Env = {
...
+  Persisted_state = initialize_store(Intermediate_state, A.wasm_module)
+)
+
+(A.mode = upgrade U and U.wasm_memory_persistence = keep)
```
crusso, line 4421:
```suggestion
(A.mode = variant { upgrade = opt U } and U.wasm_memory_persistence = opt variant { keep } )
```

spec/index.md
```diff
@@ -4353,7 +4351,11 @@ S with
...
+Only the controllers of the given canister can install new code. This changes the code of an *existing* canister, preserving the state in the stable memory. This involves invoking the `canister_pre_upgrade` method, if the `skip_pre_upgrade` flag is not set to `opt true`, on the old and `canister_post_upgrade` method on the new canister, which must succeed and must not invoke other methods. If the `wasm_memory_persistence` flag is set to `opt keep`, then the WebAssembly memory is preserved.
+
+In the following, the `initial_wasm_store` is the store of the WebAssembly module after instantiation (as per WebAssembly spec) of the WebAssembly module contained in `A.wasm_module`, before executing a potential `(start)` function. The store `initialize_store(State, A.wasm_module)` is the store of the WebAssembly module after instantiation (as per WebAssembly spec) of the WebAssembly module contained in `A.wasm_module` while reusing the WebAssembly memory of `State`.
+
+If the canister exports a private custom section with the name "enhanced-orthogonal-persistence", then the `wasm_memory_persistence` option must be set to `opt keep`.
```
Dfinity-Bjoern, line 4358: I first misunderstood the mechanism as looking at the custom section of the *new* module. Re-reading this sentence, it is already unambiguous, but maybe we want to still emphasize a bit more explicitly that the custom section is in the *already running* wasm module.

spec/index.md
mraszyk: I assume an informal dot notation on `A` since `A` is turned into a binary payload using the function `candid(A)`.

crusso (commented):
LGTM

Dfinity-Bjoern (approved):
Makes sense thanks for the update @mraszyk

spec/index.md
```diff
@@ -4391,28 +4395,47 @@ Env = {
...
+  Persisted_state = initialize_store(Intermediate_state, A.wasm_module)
+)
+
+(A.mode = upgrade U and U.wasm_memory_persistence = keep)
+or
+(A.mode = upgrade U and U.wasm_memory_persistence = replace)
+or
+(S.canisters[A.canister_id].private_custom_sections["enhanced-orthogonal-persistence"] = null)
+
+not (A.mode = upgrade U and U.wasm_memory_persistence = keep and Private_custom_sections["enhanced-orthogonal-persistence"] = null)
```
mraszyk, line 4429: @luc-blaeser Should we add an analogous check that the old module must declare the custom section "enhanced-orthogonal-persistence" if we use the "keep" option?

crusso: I just discussed this with @luc-blaeser and I think his implementation implements that check, so the spec is probably missing it. But best to get confirmation from @luc-blaeser himself...

luc-blaeser: Yes, I agree, it would make sense to check that both at least the **new** version has the EOP custom section, when `Keep` is specified. Otherwise, a classical version would misinterpret the non-zero main memory and not see any serialized state in stable memory. This is currently implemented in the MR (https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/17893)

We could also additionally require that the **old** version carries the EOP custom section, when `Keep` is specified. However, this is less critical, as the EOP runtime should detect this case and simply deserialize state from stable memory and reset the main memory state. This check is currently not implemented in the MR.

mraszyk: >  This is currently implemented in the MR

and also in the spec at the line

```
not (A.mode = upgrade U and U.wasm_memory_persistence = keep and Private_custom_sections["enhanced-orthogonal-persistence"] = null)
```

> This check is currently not implemented in the MR and in my opinion, not necessarily needed.

This is what I was asking.

>  as the EOP runtime should detect this case and simply deserialize state from stable memory and reset the main memory state

Then it sounds safe, indeed.

PR #282, opened 2024-03-12
chore: avoid canister install mode candid duplication

mraszyk:
This PR factors out the canister install mode from install_code and install_chunked_code argument definitions in the candid definitions.


Dfinity-Bjoern (approved):
Thanks @mraszyk !

PR #283, opened 2024-03-13
[FINAL] feat: 64-bit system API

mraszyk:
This PR introduces system API functions taking 64-bit WASM memory addresses.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1372,12 +1372,12 @@ defaulting to `I = i32` if the canister declares no memory.
     ic0.call_new :
       ( callee_src  : I,
         callee_size : I,
-        name_src  : I,
-        name_size : I,
-        reply_fun : i32,
-        reply_env : i32,
-        reject_fun : i32,
-        reject_env : i32
+        name_src    : I,
+        name_size   : I,
+        reply_fun   : i32,
+        reply_env   : i32,
+        reject_fun  : i32,
+        reject_env  : i32
       ) -> ();                                                                            // U CQ Ry Rt CRy CRt T
     ic0.call_on_cleanup : (fun : i32, env : i32) -> ();                                   // U CQ Ry Rt CRy CRt T
     ic0.call_data_append : (src : I, size : I) -> ();                                     // U CQ Ry Rt CRy CRt T
@@ -1389,15 +1389,15 @@ defaulting to `I = i32` if the canister declares no memory.
     ic0.stable64_write : (offset : i64, src : i64, size : i64) -> ();                     // * s
     ic0.stable64_read : (dst : i64, offset : i64, size : i64) -> ();                      // * s
 
-    ic0.certified_data_set : (src: I, size : I) -> ();                                    // I G U Ry Rt T
+    ic0.certified_data_set : (src : I, size : I) -> ();                                   // I G U Ry Rt T
     ic0.data_certificate_present : () -> i32;                                             // *
     ic0.data_certificate_size : () -> I;                                                  // NRQ CQ
-    ic0.data_certificate_copy : (dst: I, offset : I, size : I) -> ();                     // NRQ CQ
+    ic0.data_certificate_copy : (dst : I, offset : I, size : I) -> ();                    // NRQ CQ
 
     ic0.time : () -> (timestamp : i64);                                                   // *
     ic0.global_timer_set : (timestamp : i64) -> i64;                                      // I G U Ry Rt C T
     ic0.performance_counter : (counter_type : i32) -> (counter : i64);                    // * s
-    ic0.is_controller: (src: I, size : I) -> ( result: i32);                              // * s
+    ic0.is_controller: (src : I, size : I) -> ( result: i32);                             // * s
     ic0.in_replicated_execution: () -> (result: i32);                                     // * s
 
     ic0.debug_print : (src : I, size : I) -> ();                                          // * s
@@ -1587,14 +1587,14 @@ During the canister upgrade process, `canister_pre_upgrade` sees the old counter
 When handling an update call (or a callback), a canister can do further calls to another canister. Calls are assembled in a builder-like fashion, starting with `ic0.call_new`, adding more attributes using the `ic0.call_*` functions, and eventually performing the call with `ic0.call_perform`.
 
 -   `ic0.call_new :
-    ( callee_src : I,
+    ( callee_src  : I,
       callee_size : I,
-      name_src  : I,
-      name_size : I,
-      reply_fun : i32,
-      reply_env : i32,
-      reject_fun : i32,
-      reject_env : i32,
+      name_src    : I,
+      name_size   : I,
+      reply_fun   : i32,
+      reply_env   : i32,
+      reject_fun  : i32,
+      reject_env  : i32,
     ) → ()`; I ∈ {i32, i64}
 
 Begins assembling a call to the canister specified by `callee_src/_size` at method `name_src/_size`.
@@ -6576,7 +6576,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       return |es.params.arg|
 
     I ∈ {i32, i64}
-    ic0.msg_arg_data_copy<es>(dst:I, offset : I, size : I) =
+    ic0.msg_arg_data_copy<es>(dst : I, offset : I, size : I) =
       if es.context ∉ {I, U, RQ, NRQ, CQ, Ry, CRy, F} then Trap {cycles_used = es.cycles_used;}
       copy_to_canister<es>(dst, offset, size, es.params.arg)
 
@@ -6586,7 +6586,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       return |es.params.caller|
 
     I ∈ {i32, i64}
-    ic0.msg_caller_copy(dst:I, offset : I, size : I) =
+    ic0.msg_caller_copy(dst : I, offset : I, size : I) =
       if es.context = s then Trap {cycles_used = es.cycles_used;}
       copy_to_canister<es>(dst, offset, size, es.params.caller)
 
@@ -6600,7 +6600,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       return |es.params.reject_msg|
 
     I ∈ {i32, i64}
-    ic0.msg_reject_msg_copy<es>(dst:I, offset : I, size : I) =
+    ic0.msg_reject_msg_copy<es>(dst : I, offset : I, size : I) =
       if es.context ∉ {Rt, CRt} then Trap {cycles_used = es.cycles_used;}
       copy_to_canister<es>(dst, offset, size, es.params.reject_msg)
 
@@ -6690,7 +6690,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       return |es.wasm_state.self_id|
 
     I ∈ {i32, i64}
-    ic0.canister_self_copy<es>(dst:I, offset : I, size : I) =
+    ic0.canister_self_copy<es>(dst : I, offset : I, size : I) =
       if es.context = s then Trap {cycles_used = es.cycles_used;}
       copy_to_canister<es>(dst, offset, size, es.wasm_state.self_id)
 
@@ -6905,7 +6905,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       es.wasm_state.store.mem[offset..offset+size] := es.wasm_state.stable.mem[src..src+size]
 
     I ∈ {i32, i64}
-    ic0.certified_data_set<es>(src: I, size : I) =
+    ic0.certified_data_set<es>(src : I, size : I) =
       if es.context ∉ {I, G, U, Ry, Rt, T} then Trap {cycles_used = es.cycles_used;}
       es.new_certified_data := es.wasm_state[src..src+size]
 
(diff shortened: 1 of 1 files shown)


crusso (commented):
Nice, That was quick!

I know that wasm lets you overload imports with different types, but I wonder if that might be a problem for certain hosts and other tools. 

The system API as it stands, is designed to work with memory 0, right?
Instead of having two versions of some functions,
Would it instead be possible to just have one function whose signature depends on the size of memory 0? I don't really see any canister requiring both functions at the same time, right?

Then a canister will only require one 32 or 64-bit system API, and never see both at the same time and the definition of I will just depend on its memory 0.

In the spec, the side conditions in I in {i32, i64} would disappear and we specify  I = i32 or I=i64 elsehwere.

spec/index.md
```diff
@@ -1298,78 +1298,80 @@ In the reply callback of a [inter-canister method call](#system-api-call), the a
...
-    ic0.canister_status : () -> i32;                                            // *
-    ic0.canister_version : () -> i64;                                           // *
-
-    ic0.msg_method_name_size : () -> i32;                                       // F
-    ic0.msg_method_name_copy : (dst : i32, offset : i32, size : i32) -> ();     // F
-    ic0.accept_message : () -> ();                                              // F
-
-    ic0.call_new :                                                              // U CQ Ry Rt CRy CRt T
-      ( callee_src  : i32,
+The notation `I ∈ {i32, i64}` means that there are actually two corresponding System API functions for the two integer types `I`: `i32` and `i64`.
```
luc-blaeser, line 1301: Maybe the overloading could lead to difficulties in Rust and other languages that do not support overloading. This can probably be solved by specifying a Wasm binding attribute but I am not sure how easy this is.

spec/index.md
```diff
@@ -1298,78 +1298,80 @@ In the reply callback of a [inter-canister method call](#system-api-call), the a
...
+      ) -> (); I ∈ {i32, i64}                                                             // U CQ Ry Rt CRy CRt T
+    ic0.call_on_cleanup : (fun : i32, env : i32) -> ();                                   // U CQ Ry Rt CRy CRt T
+    ic0.call_data_append : (src : I, size : i32) -> (); I ∈ {i32, i64}                    // U CQ Ry Rt CRy CRt T
+    ic0.call_cycles_add : (amount : i64) -> ();                                           // U Ry Rt T
+    ic0.call_cycles_add128 : (amount_high : i64, amount_low: i64) -> ();                  // U Ry Rt T
+    ic0.call_perform : () -> ( err_code : i32 );                                          // U CQ Ry Rt CRy CRt T
+
+    ic0.stable_size : () -> (page_count : i32);                                           // * s
+    ic0.stable_grow : (new_pages : i32) -> (old_page_count : i32);                        // * s
+    ic0.stable_write : (offset : i32, src : i32, size : i32) -> ();                       // * s
```
luc-blaeser, line 1355: Conceptually, `src` would also become `I` for `stable_write` and `stable64_write` (distinguishing the address space of main memory from stable memory). Analogously for `dst` in `stable_read` and `stable64_read`. However, I also see good rationale to simply keep the current two explicit versions for historical reasons.

spec/index.md
```diff
@@ -6297,7 +6312,8 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       if es.balance >= 2^64 then Trap {cycles_used = es.cycles_used;}
       return es.balance
 
-    ic0.canister_cycles_balance128<es>(dst : i32) =
+    I ∈ {i32, i64}
+    ic0.canister_cycles_balance128<es>(dst : i) =
```
luc-blaeser, line 6316:
```suggestion
    ic0.canister_cycles_balance128<es>(dst : I) =
```

spec/index.md
mraszyk: I kept the types here since we already have the two versions with separate names and I don't think we could ever deprecate one of the two versions.

spec/index.md
mraszyk: I guess you shouldn't need to use the two versions at the same time so conditional compilation should do the job.

luc-blaeser:
Thanks a lot Martin for this PR. Very nice.
As mentioned by Claudio, the only question I have is related to overloading whether this could cause any difficulties with Rust or other non-overloading-capable languages. A kind of a generic API with a parametric main memory pointer type determined based on the bit-width of the main memory seems like an interesting solution approach, provided that it can be easily realized in the IC implementation.

spec/index.md
luc-blaeser: I agree, importing the API functions in a canister is probably easy as the compiler can selectively import only one version of the overloaded functions as needed for the corresponding memory address space. The only thing I am not sure is related to the IC implementation as it would have to export the overloaded functions, because a canister may decide to mix both 32-bit and 64-bit IC calls. Probably, wasm bindings can solve this mapping in Rust.

mraszyk: Actually, the replica implementation wouldn't allow overloading system API functions (adding an overloaded function [here](https://github.com/dfinity/ic/blob/d51f0f58f899e404be9bc0b60a0d35df53cb2553/rs/embedders/src/wasmtime_embedder/system_api.rs#L434) crashes the replica).

luc-blaeser: Thanks for the analysis, Martin. 

Maybe, the other idea would work where the replica would only expose one version of the function with the pointer type dependent on the bit-width of the canister's main memory?

mraszyk: > with the pointer type dependent on the bit-width of the canister's main memory?

How about letting the canister choose the bit-width as an optional argument (with 32-bit default) to install_code?

luc-blaeser: A nice design property of Wasm Memory64 is that it is automatically backwards-compatible to Wasm32. No canister option is needed on the IC to enable it and it can be dynamically detected in the runtime system. Ideally, we could select the IC system API bit-width also automatically, without having an explicit install option. I believe this would be simpler for the users and the implementation and it would also be a little bit more consistent (the API bit-width would always match the memory bit-width). If I am not mistaken, introducing a install option for the system API bit-width would be more complicated as it needs to be stored as additional configuration in the canister state (to be available for subsequent IC calls on that canister). This would probably require more effort than the `wasm_memory_persistence` upgrade option that is only temporarily needed during the upgrade and does not need to be stored beyond the install message.

adambratschikaye: I don't really see why a canister would be using a mix of the two types. So we could probably infer which types should be used from the type of the memory defined in the Wasm module?

crusso: > I don't really see why a canister would be using a mix of the two types. So we could probably infer which types should be used from the type of the memory defined in the Wasm module?

That's get's my vote.

crusso, line 1301:
```suggestion
The notation `I ∈ {i32, i64}` means that there are actually two corresponding System API functions for the two integer types `I`: `i32` and `i64`.
Since the System API functions target memory 0 of a particular canister, `I` is determined uniformly by the Wasm type of the canister's memory 0.

```

alexandru-uta: +1 on this as well.

mraszyk:
> I see that for many function calls only `dst` has type `I`, but where it semantically makes sense, should we also have `offset` and `size` also type `I`?

Do you foresee that canisters will be dealing with more than 4GB of data in one system API call? In that case, using I everywhere would be necessary.

alexandru-uta:
> > I see that for many function calls only `dst` has type `I`, but where it semantically makes sense, should we also have `offset` and `size` also type `I`?
> 
> Do you foresee that canisters will be dealing with more than 4GB of data in one system API call? In that case, using I everywhere would be necessary.

Not necessarily, but on the implementation side it's usually much more cleaner to have all args the same type.

spec/index.md
mraszyk: Sounds good. I've updated this PR to derive the value of `I` from the bit-width of the memory.

eust-dfinity:
> > > I see that for many function calls only `dst` has type `I`, but where it semantically makes sense, should we also have `offset` and `size` also type `I`?
> > 
> > 
> > Do you foresee that canisters will be dealing with more than 4GB of data in one system API call? In that case, using I everywhere would be necessary.
> 
> Not necessarily, but on the implementation side it's usually much more cleaner to have all args the same type. @eust-dfinity and @luc-blaeser, what's you take on this?

Most probably we won't need offsets and sizes bigger than 32bit. However it makes sense from the api perspective to have them of the same type. It is more convenient to call (src/dst, buf, buf.size()) without having to cast buf.size() to a smaller type. All the other 64bit functions involved have consistent u64 arguments (e.g. our stable64 api have offsets and lengths as u64 even though it used to copy to a 32bit destination).

I guess the question is what is more expected (unexpected) from a typical developer perspective. (for me the more expected is 64bit everywhere)

luc-blaeser:
> 
> I guess the question is what is more expected (unexpected) from a typical developer perspective. (for me the more expected is 64bit everywhere)

I agree, it would somewhat ease the integration as it is uniform with the address bit-width and probably save some casts (by the CDK and maybe also in the IC runtime).

spec/index.md
```diff
@@ -1299,77 +1303,82 @@ In the reply callback of a [inter-canister method call](#system-api-call), the a
...
-    ic0.time : () -> (timestamp : i64);                                         // *
-    ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U Ry Rt C T
-    ic0.performance_counter : (counter_type : i32) -> (counter : i64);          // * s
-    ic0.is_controller: (src: i32, size: i32) -> ( result: i32);                 // * s
-    ic0.in_replicated_execution: () -> (result: i32);                           // * s
-
-    ic0.debug_print : (src : i32, size : i32) -> ();                            // * s
-    ic0.trap : (src : i32, size : i32) -> ();                                   // * s
+      ) -> ();                                                                            // U CQ Ry Rt CRy CRt T
+    ic0.call_on_cleanup : (fun : i32, env : i32) -> ();                                   // U CQ Ry Rt CRy CRt T
```
Dfinity-Bjoern, line 1354: From interface spec meeting: it may be worth generalizing the `env` to `I` as well since it could legitimately be a pointer to memory.

eust-dfinity: Ohh, good point. cdk seems to be using it by putting raw pointers there...

eust-dfinity: seems like tables are indexed by u32 in both wasm32 and wasm64 modes

crusso:
I just double-checked with @luc-blaeser and he is fine (and already assumes) 32-bit `env` values for any callbacks.

spec/index.md
lwshang: I'm working on experimenting wasm64 support in Rust CDK.
And I just noticed that it is not only `env` but also the `fun`/`reply_fun`/`reject_fun` will be 32 bit in wasm64.

The `env: i32` can be handled by introduced an [indirection](https://github.com/dfinity/cdk-rs/pull/492).
When it comes to `fun`, I don't know how to do it. In [current Rust CDK](https://github.com/dfinity/cdk-rs/blob/05ca5719c3baf7fb4c4982e31097ff8b07fd8a25/src/ic-cdk/src/api/call.rs#L106-L114), we are passing function pointers.

spec/index.md
```diff
@@ -1401,78 +1407,89 @@ The 32-bit stable memory System API (`ic0.stable_size`, `ic0.stable_grow`, `ic0.
...
+    ic0.performance_counter : (counter_type : i32) -> (counter : i64);                    // * s
+    ic0.is_controller: (src : I, size : I) -> ( result: i32);                             // * s
+    ic0.in_replicated_execution: () -> (result: i32);                                     // * s
+
+    ic0.debug_print : (src : I, size : I) -> ();                                          // * s
+    ic0.trap : (src : I, size : I) -> ();                                                 // * s
+```
+
+The following System API functions are only available if `I = i32`, i.e., if the bit-width of the declared memory is 32
+or if the canister declares no memory.
```
mraszyk, line 1481: Is this really the case in the replica?

mraszyk: @adambratschikaye Could you please confirm?

berestovskyy: Yes, the System APIs use i32 as a default, and switch to i64 if the module declares any memory and it is 64-bit. It's scattered across the code, but you can see an example in `main_memory_type()`.

mraszyk: On the other hand, [here](https://github.com/dfinity/ic/blob/7f4cfdf6e469c37c14fc9cd8bc4a5a61cb7ed759/rs/embedders/src/wasmtime_embedder/system_api.rs#L302) it doesn't seem to make difference, e.g., the 64-bit cycles API is supposed to be disabled for 64-bit memories, but [here](https://github.com/dfinity/ic/blob/7f4cfdf6e469c37c14fc9cd8bc4a5a61cb7ed759/rs/embedders/src/wasmtime_embedder/system_api.rs#L967) it seems to be enabled.

berestovskyy: Ah, I see. You're asking to verify that the 32/64-bit APIs in replica align with the specification. I did a similar check while reviewing some older MRs, but I agree we should do another pass now. cc @alexandru-uta

alexandru-uta: thanks, will have another look.

alexandru-uta: @mraszyk and @berestovskyy , I think these are checked in [validation](https://github.com/dfinity/ic/blob/7f4cfdf6e469c37c14fc9cd8bc4a5a61cb7ed759/rs/embedders/src/wasm_utils/validation.rs#L773) and there's a difference on what imports are allowed in 32 or 64 bit modes.

mraszyk:
Superseded by https://github.com/dfinity/portal/pull/3762

PR #284, opened 2024-03-15
fix: update subnet read state conditions to match the implementation

mraszyk:
This PR updates the subnet read state conditions to match the replica implementation.

### DIFF
--- a/spec/_attachments/interface-spec-changelog.md
+++ b/spec/_attachments/interface-spec-changelog.md
@@ -1,6 +1,7 @@
 ## Changelog {#changelog}
 
 ### ∞ (unreleased)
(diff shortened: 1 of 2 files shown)


Dfinity-Bjoern (commented):
Discussed in the interface spec meeting that it seems more logical to allow access to the data except metrics such that it can always be requested. (De facto that's anyway encoded in the current scheme since the effective canister id is not checked.)

mraszyk:
> Discussed in the interface spec meeting that it seems more logical to allow access to the data except metrics such that it can always be requested. (De facto that's anyway encoded in the current scheme since the effective canister id is not checked.)

Done both in the spec and replica code.

PR #286, opened 2024-03-22
feat: on low wasm memory hook

ielashi:
This proposal is based on this [forum post](https://forum.dfinity.org/t/canister-lifecycle-hooks/17089) and has already been approved by [motion proposal 106146](https://dashboard.internetcomputer.org/proposal/106146).

Canister developers have to actively monitor their canisters to know if they are low on wasm memory. If detected too late, a canister can be completely stuck and forever un-upgradable.

To address this, we introduce a hook called `on_low_wasm_memory`. When defined, it is triggered whenever the canister's memory usage exceeds some developer-defined threshold.

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -14,7 +14,7 @@ type canister_settings = record {
     reserved_cycles_limit : opt nat;
     log_visibility : opt log_visibility;
     wasm_memory_limit : opt nat;
-    low_wasm_memory_threshold : opt nat;
+    wasm_memory_threshold : opt nat;
 };
 
 type definite_canister_settings = record {
@@ -25,7 +25,7 @@ type definite_canister_settings = record {
     reserved_cycles_limit : nat;
     log_visibility : log_visibility;
     wasm_memory_limit : nat;
-    low_wasm_memory_threshold: nat;
+    wasm_memory_threshold: nat;
 };
 
 type change_origin = variant {
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -7,6 +7,7 @@ type canister_settings = record {
     memory_allocation : opt nat;
     freezing_threshold : opt nat;
     reserved_cycles_limit : opt nat;
+    low_wasm_memory_threshold : opt nat;
 };
 
 type definite_canister_settings = record {
@@ -15,6 +16,7 @@ type definite_canister_settings = record {
     memory_allocation : nat;
     freezing_threshold : nat;
     reserved_cycles_limit : nat;
+    low_wasm_memory_threshold: nat;
 };
 
 type change_origin = variant {
--- a/spec/index.md
+++ b/spec/index.md
@@ -1295,7 +1295,7 @@ While an implementation will likely try to keep the interval between the value o
 
 #### On Low Wasm Memory {#on-low-wasm-memory}
 
-A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's wasm memory size in bytes grows from below a threshold `t` to >= t.
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's wasm memory size in bytes grows from below a threshold `t` to >= `t`.
 The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 2<sup>64</sup> − 1.
 
 :::note
@@ -2093,7 +2093,7 @@ Indicates various information about the canister. It contains:
 
     -   The reserved cycles limit of the canister, i.e., the maximum number of cycles that can be in the canister's reserved balance after increasing the canister's memory allocation and/or actual memory usage.
 
-    -   The "low wasm memory" threshold, which is used to determine when the `on_low_wasm_memory` function is executed. See {#on-low-wasm-memory}.
+    -   The "low wasm memory" threshold, which is used to determine when the [canister_on_low_wasm_memory](#on-low-wasm-memory) function is executed.
 
 -   A SHA256 hash of the module installed on the canister. This is `null` if the canister is empty.
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -1354,7 +1354,7 @@ While an implementation will likely try to keep the interval between the value o
 
 #### On Low Wasm Memory {#on-low-wasm-memory}
 
-A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's wasm memory size in bytes grows from below a threshold `t` to >= `t`.
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from above a threshold `t` to < `t`.
 The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 2<sup>64</sup> − 1.
 
 :::note
--- a/spec/index.md
+++ b/spec/index.md
@@ -1373,12 +1373,12 @@ While an implementation will likely try to keep the interval between the value o
 
 #### On Low Wasm Memory {#on-low-wasm-memory}
 
-A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from above a threshold `t` to strictly less than `t`.
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from at least a threshold `t` to strictly less than `t`.
 The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 0.
 
 :::note
 
-While the above function is scheduled immediately once the condition above is triggered, it may not necessarily be executed immediately if it does not have enough cycles.
+While the above function is scheduled immediately once the condition above is triggered, it may not necessarily be executed immediately if the canister does not have enough cycles.
 If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory remains above the threshold `t`.
 
 :::
@@ -3188,6 +3188,7 @@ The [WebAssembly System API](#system-api) is relatively low-level, and some of i
           composite_query_methods : MethodName ↦ ((Arg, CallerId, Env) -> CompositeQueryFunc)
           heartbeat : (Env) -> SystemTaskFunc
           global_timer : (Env) -> SystemTaskFunc
+          on_low_wasm_memory : (Env) -> SystemTaskFunc
           callbacks : (Callback, Response, RefundedCycles, Env, AvailableCycles) -> UpdateFunc
           composite_callbacks : (Callback, Response, Env) -> UpdateFunc
           inspect_message : (MethodName, WasmState, Arg, CallerId, Env) -> Trap | Return {
@@ -3254,6 +3255,7 @@ EntryPoint
   | Callback Callback Response RefundedCycles
   | Heartbeat
   | GlobalTimer
+  | OnLowWasmMemory
 Message
   = CallMessage {
       origin : CallOrigin;
@@ -3410,6 +3412,10 @@ CanisterLog = {
   timestamp_nanos : Nat;
   content : Blob;
 }
+OnLowWasmMemoryHookStatus
+  = ConditionNotSatisfied
+  | Ready
+  | Executed
 QueryStats = {
   timestamp : Timestamp;
   num_instructions : Nat;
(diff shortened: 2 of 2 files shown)


ielashi:
@mraszyk I didn't make any changes to the formal model in this PR yet, as I first wanted to get your feedback that the proposal as-is looks good.

spec/index.md
```diff
@@ -1289,6 +1293,18 @@ While an implementation will likely try to keep the interval between the value o
...
 
+#### On Low Wasm Memory {#on-low-wasm-memory}
+
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's wasm memory size in bytes grows from below a threshold `t` to >= `t`.
+The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 2<sup>64</sup> − 1.
+
+:::note
+
+While the above function is scheduled immediately once the condition above is triggered, it may not necessarily be executed immediately if it does not have enough cycles.
+If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory remains above the threshold `t`.
```
mraszyk, line 1304: So you don't plan to treat it analogously to call_on_cleanup after all?

ielashi: We discussed it and preferred to move forward with the scheduling approach. The reason for that is that `call_on_cleanup`, unlike this hook, cannot make outbound calls. Making outbound calls in the hook while also treating it like `call_on_cleanup` causes unnecessary complications in the implementation.

mraszyk: > The reason for that is that call_on_cleanup, unlike this hook, cannot make outbound calls.

As long as it can schedule a one-off timer, I don't see a major issue with this limitation.

dsarlis:
> @mraszyk I didn't make any changes to the formal model in this PR yet, as I first wanted to get your feedback that the proposal as-is looks good.

@ielashi @mraszyk I assume that we're fine with the current status of this PR so can the formal model be updated so we can move this to "FINAL" and be ready to start implementation at our convenience?

ielashi:
@mraszyk Is the formal model something that you'd update yourself or should we look into that on our side?

mraszyk:
> I assume that we're fine with the current status of this PR

How about the call_on_cleanup analogy that I raised in the unresolved comment above?

ielashi:
> > I assume that we're fine with the current status of this PR
> 
> How about the call_on_cleanup analogy that I raised in the unresolved comment above?

@mraszyk Doesn't [this](https://github.com/dfinity/interface-spec/pull/286#discussion_r1537213212) answer your question?

mraszyk:
@ielashi I replied

> The reason for that is that call_on_cleanup, unlike this hook, cannot make outbound calls.

As long as it can schedule a one-off timer, I don't see a major issue with this limitation.

on that thread and you put thumbs up so I thought you saw that.

spec/index.md
```diff
@@ -1399,7 +1415,7 @@ The comment after each function lists from where these functions may be invoked:
 
 -   `F`: from `canister_inspect_message`
 
--   `T`: from *system task* (`canister_heartbeat` or `canister_global_timer`)
+-   `T`: from *system task* (`canister_heartbeat` or `canister_global_timer` `canister_on_low_wasm_memory`)
```
dragoljub-duric, line 1418:
```suggestion
-   `T`: from *system task* (`canister_heartbeat` or `canister_global_timer` or `canister_on_low_wasm_memory`)
```

spec/index.md
```diff
@@ -1289,6 +1293,18 @@ While an implementation will likely try to keep the interval between the value o
 
 :::
 
+#### On Low Wasm Memory {#on-low-wasm-memory}
+
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's wasm memory size in bytes grows from below a threshold `t` to >= `t`.
```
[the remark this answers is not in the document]
dragoljub-duric: @mraszyk WDYT? I think that more important information is how much memory is left. Furthermore, if we decide to check the replica's remaining memory, we cannot use the upper bound on the wasm memory size for that.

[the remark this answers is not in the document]
mraszyk: Makes sense to me. Should the inequality be <= or < ?

spec/index.md
```diff
@@ -1297,6 +1301,18 @@ While an implementation will likely try to keep the interval between the value o
...
+The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 2<sup>64</sup> − 1.
```
mraszyk, line 1307: The default value should be 0 with your above change, right?

dragoljub-duric: That is correct, thank you!

spec/index.md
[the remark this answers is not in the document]
dragoljub-duric: "<" makes more sense, since the default is 0, hence if we have a default value of the treshold the hook will never be invoked. I updated sugesstion.

spec/index.md
dragoljub-duric, line 1307:
```suggestion
The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 0.
```

dragoljub-duric:
@mraszyk Should we add a new API type as well?

dragoljub-duric:
@mraszyk @ielashi @dsarlis 
**Question 1:**
          WDYT when this hook should be executed:
          1. In every round canister is scheduled - this approach seems non-optimal because when we do not have anything to execute there is no risk of lack of memory
          2. In every round canister executes a task or message - this approach is the safest since we will ensure that canister has enough memory before every execution, but we will have a bigger overhead when we execute hook with for example every heartbeat 
          3. In every round canister executes a message - this approach seems like reasonable tradeoff, but I am not sure that decreasing overhead from previous point is worth the risk

          
**Question 2:**
         Should the hook be executed before or after the execution of the message (or task)? 
         
         Since the number of hooks executed will remain the same, I propose that the hook is always executed prior to the other execution, reason: 
         Suppose we decide to take into the account remaining memory on the subnet, after the previous execution of canister it had enough memory, and subnet capacity was sufficient as well. But in meantime, the subnet capacity becomes filled and the canister attempts new execution expecting that it has enough memory which is not the case. 
         
         On the other side executing the hook after other executions can alert users earlier that memory is running low, but I think the other argument is stronger.
         
         
**Question 3:**
            If canister executes multiple messages (and/or tasks) in a single round should the hook be invoked before every execution?

spec/index.md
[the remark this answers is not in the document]
Dfinity-Bjoern: Discussion in the interface spec meeting:
- The semantics here are unclear, for instance in the case that the canister has a certain memory allocation and uses both heap and stable memory. What does "the canister's remaining wasm memory size" then refer to, exactly?
- The formulation seems to indicate a guarantee that actually cannot be provided (without an explicit memory allocation): Suppose some canister makes a big memory allocation, which brings some 10s of thousands of canisters on the same subnet over/under the threshold. We would have to call the hook on all of those 10s of thousands of canisters, but it's very unlikely that all of those would be able to satisfy their memory needs.

Thus, we are not convinced that this change of semantics is beneficial, given that the only way for a canister to guarantee that it can still allocate memory (from the shared resource of overall subnet memory) is via memory allocation, which however has ambiguous semantics as described above.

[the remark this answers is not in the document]
Dfinity-Bjoern: Another complication here is that the canister may not be able to _pay for_ allocating additional memory due to the cycles prepayment when the subnet memory becomes scarce. Should that trigger the condition as well?

Dfinity-Bjoern:
As described above, we suggest keeping the old semantics (specifying a limit on allocated heap memory rather than "remaining" memory), in which case the scheduling is also much clearer (just execute when an execution of the canister exceeds the limit).

spec/index.md
```diff
@@ -1485,7 +1501,7 @@ The comment after each function lists from where these functions may be invoked:
 
 -   `F`: from `canister_inspect_message`
 
--   `T`: from *system task* (`canister_heartbeat` or `canister_global_timer`)
+-   `T`: from *system task* (`canister_heartbeat` or `canister_global_timer` or `canister_on_low_wasm_memory`)
```
Dfinity-Bjoern, line 1504: If we make this of type `T`, then a canister could update the threshold in a setting and allocate a new page, which should make the hook trigger again. So if we guarantee *immediate* execution, one can build a chain of calls to the hook. So we need to make sure that *immediate* may mean that it executes as part of the next execution round (just that nothing else runs in between).

spec/index.md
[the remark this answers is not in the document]
dragoljub-duric: I agree it is better to look only at the `canister's memory`, not the `subnet available memory`. In that case:

1. If the certain application uses both heap and stable memory then the `remaining wasm memory` will be:

min(`memory_allocation` - `used_stable_memory`, 4 GB) - `used_heap_memory`

2. If we look only at the canister's memory, the allocation of the other canister cannot trigger the hook on the current canister, hence that will not be a problem. Furthermore, by checking if the `remaining wasm memory` is below the `threshold` we will ensure enough `wasm memory` in case of usage of `memory allocation` (first part in the formula above), which cannot be done if the hook is checking if the `wasm memory` if above the `threshold`.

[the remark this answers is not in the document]
Dfinity-Bjoern: The formula would likely have to include a few more terms that are accounted for in `memory_allocation`, such as the wasm module size or the size of the canister history. When that is done, it seems an acceptable API.

I guess in my opinion there is no clearly "better" API. The "old" one is a bit easier to understand in that it uses more basic values, but has the disadvantage of probably requiring the canister to change the threshold at least upon significant allocation of additional stable memory. This "new" variant may be a bit easier to use, but is bit more complicated to understand and would need to be adapted when new features (e.g. multiple wasm memories) are supported.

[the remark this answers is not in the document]
dragoljub-duric: In that case, I will go forward and update the specs to use “remaining memory drops to below a threshold”.

dragoljub-duric:
@Dfinity-Bjoern @mraszyk If we start keeping information about the hook as {"Condition is not satisfied", "Executed", "Ready to be executed"}, and based on that we use to determine if we will run hook, do you think we should persist that information in snapshots.
Screenshot 2024-07-30 at 15 15 20

spec/index.md
```diff
@@ -1271,6 +1273,8 @@ The canister provides entry points which are invoked by the IC under various cir
 
 -   The canister may export functions with name `canister_composite_query <name>` and type `() -> ()`.
 
+-   The canister may export functions with name `canister_on_low_wasm_memory` and type `() -> ()`.
```
mraszyk, line 1276:
```suggestion
-   The canister may export a function with the name `canister_on_low_wasm_memory` and type `() -> ()`.
```

spec/index.md
```diff
@@ -1348,6 +1352,18 @@ While an implementation will likely try to keep the interval between the value o
 
 :::
 
+#### On Low Wasm Memory {#on-low-wasm-memory}
+
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from above a threshold `t` to < `t`.
```
mraszyk, line 1357:
```suggestion
A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from above a threshold `t` to strictly less than `t`.
```

spec/index.md
```diff
@@ -1367,6 +1371,18 @@ While an implementation will likely try to keep the interval between the value o
...
 
+#### On Low Wasm Memory {#on-low-wasm-memory}
+
+A canister can export a function with the name `canister_on_low_wasm_memory`, which is scheduled whenever the canister's remaining wasm memory size in bytes falls from above a threshold `t` to strictly less than `t`.
+The threshold `t` can be defined in the [canister's settings](#ic-canister_status) and by default it is set to 0.
+
+:::note
+
+While the above function is scheduled immediately once the condition above is triggered, it may not necessarily be executed immediately if it does not have enough cycles.
+If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory remains above the threshold `t`.
```
mraszyk, line 1382: > If the canister gets frozen immediately after the function is scheduled for execution, the function will run once the canister's unfrozen _if_ the canister's wasm memory remains above the threshold `t`.

This property is not established by the implementation.

spec/_attachments/ic.did
```diff
@@ -25,6 +26,7 @@ type definite_canister_settings = record {
     reserved_cycles_limit : nat;
     log_visibility : log_visibility;
     wasm_memory_limit : nat;
+    wasm_memory_threshold: nat;
```
mraszyk, line 29: This field is not exported by the implementation.

mraszyk:
The implementation only considers stable memory and ignores, e.g., the chunk store and snapshot size, when deriving the wasm memory capacity.

mraszyk:
The hook should not run after an upgrade if the condition is not satisfied after the upgrade (although the condition was satisfied before the upgrade and the hook did not execute yet).

spec/index.md
```diff
@@ -3949,6 +3977,47 @@ S with
 
 ```
 
+*Call context creation: On low wasm memory*
+
+If canister `C` exports a method with name `canister_on_low_wasm_memory` and `S.on_low_wasm_memory_hook_status[C]` is `Ready`, the IC will create the corresponding call context and set `S.on_low_wasm_memory_hook_status[C]` to `Executed`.
```
mraszyk, line 3982:
```suggestion
If `S.on_low_wasm_memory_hook_status[C]` is `Ready` for a canister `C`, the IC will create the corresponding call context and set `S.on_low_wasm_memory_hook_status[C]` to `Executed`.
```

mraszyk:
Superseded by https://github.com/dfinity/portal/pull/3761

PR #287, opened 2024-03-22
fix: effective canister id of bitcoin query calls

mraszyk:
This PR fixes the effective canister id of bitcoin query calls so that it is consistent with query call response verification.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -866,11 +866,19 @@ It must be contained in the canister ranges of a subnet, otherwise the correspon
 
     -   Otherwise, the call is rejected by the system independently of the effective canister id.
 
--   If the request is an update call to a canister that is not the Management Canister (`aaaaa-aa`) or if the request is a query call, then the effective canister id must be the `canister_id` in the request.
+-   If the request is a query call to the Management Canister (`aaaaa-aa`), then:
+
+    -   If the call is to the `bitcoin_get_balance_query` or `bitcoin_get_utxos_query` method, then the effective canister id for this call must be the bitcoin canister id which is left unspecified (in particular, it depends on the bitcoin network).
+
+    -   Otherwise, the call is rejected by the system independently of the effective canister id.
+
+-   If the request is an update or query call to a canister that is not the Management Canister (`aaaaa-aa`), then the effective canister id must be the `canister_id` in the request.
 
 :::note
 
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -866,12 +866,20 @@ It must be contained in the canister ranges of a subnet, otherwise the correspon
...
+
+    -   Otherwise, the call is rejected by the system independently of the effective canister id.
+
+-   If the request is an update or query call to a canister that is not the Management Canister (`aaaaa-aa`), then the effective canister id must be the `canister_id` in the request.
 
 :::note
 
 The expectation is that user-side agent code shields users and developers from the notion of effective canister id, in analogy to how the System API interface shields canister developers from worrying about routing.
 
+The bitcoin mainnet canister ID is `ghsi2-tqaaa-aaaan-aaaca-cai`. The bitcoin testnet canister ID is `g4xu7-jiaaa-aaaan-aaaaq-cai` (also used in local testing for the bitcoin regtest network).
```
lwshang, line 881: I believe we should not mention these particular IDs in the spec.
The line of "the effective canister id for this call must be the bitcoin canister id which is left unspecified" is sufficient.
Agent authors (e.g. we, the SDK team) need to figure these ID out by themselves. If we make such information clear in some other materials, it should not be a difficult problem. 
Once the agents use these IDs as default value, the end users will not need to worry about for the most use cases.

spec/index.md
```diff
@@ -866,12 +866,20 @@ It must be contained in the canister ranges of a subnet, otherwise the correspon
 
     -   Otherwise, the call is rejected by the system independently of the effective canister id.
 
--   If the request is an update call to a canister that is not the Management Canister (`aaaaa-aa`) or if the request is a query call, then the effective canister id must be the `canister_id` in the request.
+-   If the request is a query call to the Management Canister (`aaaaa-aa`), then:
+
+    -   If the call is to the `bitcoin_get_balance_query` or `bitcoin_get_utxos_query` method, then the effective canister id for this call must be the bitcoin canister id which is left unspecified (in particular, it depends on the bitcoin network).
```
THLO, line 871: This line hints at the fact that there is a Bitcoin canister.
Note that insinuating that this protocol-level feature is backed by canisters in the interface spec would require the approval of upper management.
It would be better to avoid this path in my opinion.

mraszyk: The existence of the Bitcoin canister is publicly revealed in the [docs](https://internetcomputer.org/docs/current/references/bitcoin-how-it-works#components).

PR #290, opened 2024-03-28
Dsharifi/call v3

DSharifi:


Dfinity-Bjoern:
If we now go for a separate `sync_call` endpoint, why do we even introduce a `v3`? We can just add an endpoint under the existing `v2` path; what's speaking against that?

rumenov:
> If we now go for a separate `sync_call` endpoint, why do we even introduce a `v3`? We can just add an endpoint under the existing `v2` path; what's speaking against that?

I think the idea is to optimize for clarity long term. Maybe we should meet to discuss the pros and cons of the approaches (add v2/sync_call, add v3/ endpoints for all existing, add v3 only for call).

I am also curious if the certified queries will have different arguments than the existing query endpoint and what is the long term plan for them.

PR #292, opened 2024-04-02
fix: add note on cycles for sign_with_ecdsa

mraszyk:
This PR adds a note that cycles must be attached to calls to `sign_with_ecdsa`.


bogwar (approved):
lgtm

PR #293, opened 2024-04-02
fix: add note on cycles for create_canister

mraszyk:
This PR adds a note that cycles must be attached to calls to create_canister.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -2163,7 +2163,7 @@ The signatures are encoded as the concatenation of the [SEC1](https://www.secg.o
 
 This call requires that the ECDSA feature is enabled, the caller is a canister, and `message_hash` is 32 bytes long. Otherwise it will be rejected.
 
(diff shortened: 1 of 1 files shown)


robin-kunzler (approved):
Thanks a lot @mraszyk , LGTM!

spec/index.md
```diff
@@ -2161,6 +2163,8 @@ The signatures are encoded as the concatenation of the [SEC1](https://www.secg.o
 
 This call requires that the ECDSA feature is enabled, the caller is a canister, and `message_hash` is 32 bytes long. Otherwise it will be rejected.
 
+Cycles to pay for the call must be explicitly transferred with the call, i.e., they are not deducted from the caller's balance implicitly (e.g., as for inter-canister calls).
```
bogwar, line 2166:
```suggestion
Cycles to pay for the call must be explicitly transferred with the call, i.e., they are not automatically deducted from the caller's balance implicitly (e.g., as for inter-canister calls).
```

bogwar (approved):
lgtm (made a small suggestion)

PR #294, opened 2024-04-02
fix: refund cycles attached to provisional_top_up_canister

mraszyk:
This MR specifies that cycles attached to `provisional_top_up_canister` are refunded to the caller.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -2274,7 +2274,7 @@ This method is only available in local development instances.
 
 As a provisional method on development instances, the `provisional_top_up_canister` method is provided. It adds `amount` cycles to the balance of canister identified by `amount`.
 
-Cycles added to this call via `ic0.call_cycles_add128` are returned to the caller.
+Cycles added to this call via `ic0.call_cycles_add` and `ic0.call_cycles_add128` are returned to the caller.
 
 Any user can top-up any canister this way.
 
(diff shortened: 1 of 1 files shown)


robin-kunzler (approved):
Thanks a lot @mraszyk , LGTM! 👍

spec/index.md
```diff
@@ -2161,6 +2163,8 @@ The signatures are encoded as the concatenation of the [SEC1](https://www.secg.o
 
 This call requires that the ECDSA feature is enabled, the caller is a canister, and `message_hash` is 32 bytes long. Otherwise it will be rejected.
 
+Cycles to pay for the call must be explicitly transferred with the call, i.e., they are not deducted from the caller's balance implicitly (e.g., as for inter-canister calls).
```
bogwar, line 2166:
```suggestion
Cycles to pay for the call must be explicitly transferred with the call, i.e., they are not automatically deducted from the caller's balance implicitly (e.g., as for inter-canister calls).
```

Dfinity-Bjoern:
Is this behavior what happens in the replica?

mraszyk:
> Is this behavior what happens in the replica?

Yes. See this [line](https://github.com/dfinity/ic/blob/27656c1abce70da733682699af14bfa8c20bba92/rs/execution_environment/src/execution_environment.rs#L1118) or the discussion in the [ticket](https://dfinity.atlassian.net/browse/FOLLOW-671).

PR #296, opened 2024-04-04
fix: various (type-level) fixes

mraszyk:
This MR makes various (in particular, type-level) fixes.


spec/index.md
```diff
@@ -2871,12 +2871,20 @@ Therefore, a message can have different shapes:
           origin : CallOrigin;
           response : Response;
           refunded_cycles : Nat;
+          queue : Queue;
         }
 
-The `queue` field is used to describe the message ordering behavior. Its concrete value is only used to determine when the relative order of two messages must be preserved, and is otherwise not interpreted. Response messages are not ordered, as explained above, so they have no `queue` field.
```
mraszyk, line 2876: > Response messages are not ordered, as explained above, so they have no `queue` field.

@Dfinity-Bjoern Do you know why we have this (wrong) sentence here?

Dfinity-Bjoern: Why is this wrong? Do we guarantee ordering of responses in canister-to-canister messages (that seems difficult)? There is no ordering in the ingress history for sure.

mraszyk: It is wrong because if canister A calls canister B two times and the canister B responds to A without an async call, then the responses from B to A are ordered. Responses to ingress messages are unordered just like ingress messages themselves which is modeled by the queue being "Unordered" as opposed to a pair of canisters.

derlerd-dfinity: I'm a bit worried if we commit to a response ordering. The current implementation may coincidentally provide it but I don't think the spec should guarantee it.

derlerd-dfinity: Correction: we don't even provide response ordering guarantees between canister currently. For example, subnet splitting may break it.

mraszyk: Ok. I've undone the changes.

PR #298, opened 2024-04-04
[FINAL] Adding an endpoint to fetch bitcoin block headers

THLO:
The current Bitcoin integration API does not expose any information about block headers.
This information would be useful to verify blocks (e.g., obtained using HTTPS outcalls) by comparing the Merkle root hash with the root hash in the block header.

This PR introduces a new endpoint `bitcoin_get_block_headers`.

The PR also reorganizes some Bitcoin integration related definitions to make things easier to read from top to bottom.


spec/index.md
```diff
@@ -2376,6 +2376,17 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
 
 The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method), inclusive, with the addition of a 0th percentile is used. Concretely, for any i from 1 to 100, the ith percentile is the fee with rank `⌈i * 100⌉`. The 0th percentile is defined as the smallest fee (excluding coinbase transactions).
 
+### IC method `bitcoin_get_block_headers` {#ic-bitcoin_get_block_headers}
```
mraszyk, line 2379: Could you please mark this method as "experimental": "The bitcoin get block headers management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way."

spec/_attachments/ic.did
```diff
@@ -131,18 +131,36 @@ type bitcoin_get_balance_args = record {
...
+
 type bitcoin_send_transaction_args = record {
     transaction : blob;
     network : bitcoin_network;
 };
 
-type millisatoshi_per_byte = nat64;
+type bitcoin_get_block_headers_args = record {
+	start_height : nat32;
+	end_height : opt nat32;
```
ielashi, line 157: nit: the rest of this file is using spaces rather than tabs.

ielashi (commented):
API LGTM, leaving final approval from execution to @dsarlis.

spec/_attachments/ic.did
```diff
@@ -131,18 +131,36 @@ type bitcoin_get_balance_args = record {
...
+type bitcoin_get_current_fee_percentiles_result = vec millisatoshi_per_byte;
+
 type bitcoin_send_transaction_args = record {
     transaction : blob;
     network : bitcoin_network;
 };
 
-type millisatoshi_per_byte = nat64;
+type bitcoin_get_block_headers_args = record {
+	start_height : nat32;
```
ielashi, line 156: nit: we're using `nat32` for height in a number of places. Would it make sense to define this as a type? (e.g. `type bitcoin_block_height = nat32`)

spec/_attachments/ic.did
```diff
@@ -131,18 +131,36 @@ type bitcoin_get_balance_args = record {
...
+
+type bitcoin_get_current_fee_percentiles_result = vec millisatoshi_per_byte;
+
 type bitcoin_send_transaction_args = record {
     transaction : blob;
     network : bitcoin_network;
 };
 
-type millisatoshi_per_byte = nat64;
+type bitcoin_get_block_headers_args = record {
```
Dfinity-Bjoern, line 155: How do we see this work in terms of confirmations: the canister retrieves the block header with `end_height = null` and then discards as many unconfirmed blocks as it needs to have the level of stability needed for the application?

spec/_attachments/ic.did
THLO: Done! Can you confirm that it has been fixed?

spec/index.md
THLO: Done!

spec/_attachments/ic.did
THLO: Good idea! I added the type definition. I also changed `block_hash` to `bitcoin_block_hash` for the sake of consistency.
Do you agree with this change?

spec/index.md
```diff
@@ -2378,6 +2378,12 @@ The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Perc
 
 ### IC method `bitcoin_get_block_headers` {#ic-bitcoin_get_block_headers}
 
+:::note
+
+The node metrics management canister API is considered EXPERIMENTAL. Canister developers must be aware that the API may evolve in a non-backward-compatible way.
```
mraszyk, line 2383: node metrics -> bitcoin get block headers

spec/_attachments/ic.did
ielashi: Yes, looks good!

spec/_attachments/ic.did
```diff
@@ -131,18 +133,36 @@ type bitcoin_get_balance_args = record {
...
+    start_height : bitcoin_block_height;
+    end_height : opt bitcoin_block_height;
+};
+
+type bitcoin_get_block_headers_result = record {
+    tip_height : bitcoin_block_height;
+    block_headers : vec bitcoin_block_header;
```
ielashi, line 164: I'm wondering if it makes sense to wrap each `block_header` in a record. That way, we'd leave the door open in the future to add additional to each block header (e.g. number of confirmations).

THLO: I don't have a strong opinion about this.
Since we also return a record, an alternative would be to add more information to that record, right?
For example, if we add the `start_height` and `end_height`, then the number of confirmations for each block header is clear (given the `tip_height`). What do you think?

ielashi: SGTM, I also don't have a strong opinion about this, and as you say we can always add additional information by adding new fields to this record.

spec/index.md
THLO: Ha! Good catch, thanks!
I fixed it.

dsarlis (approved):
Approving on behalf of execution. The API seems reasonable to me and the experimental marker allows us to tweak things if needed in the future.

Dfinity-Bjoern (approved):
Looks good, please synchronize merging with implementation effort.

spec/index.md
```diff
@@ -2376,6 +2376,23 @@ This function returns fee percentiles, measured in millisatoshi/vbyte (1000 mill
...
+
+:::note
+
+The `bitcoin_get_block_headers` endpoint is considered EXPERIMENTAL. Canister developers must be aware that this endpoint may evolve in a non-backward-compatible way.
+
+:::
+
+Given a start height and an optional end height, the function returns the block headers in the provided range. The range is inclusive, i.e., the block headers at the start and end heights are returned as well.
+
+ If no end height is specified, all blocks until the tip height, i.e., the largest available height, are returned. However, if the range from the start height to the end height or the tip height is large, only a prefix of the requested block headers may be returned in order to bound the size of the response.
```
dragoljub-duric, line 2389: What should happen if the end height is specified and it is larger than the height of the main chain?

THLO: Good question. In my opinion, the most natural approach is to return block headers until the tip height in this case.
Do you agree?

I think the text should be updated to make this clear.

dragoljub-duric: In my opinion, it is clearer to return an error, if the user wants all blocks until the end it has an option to specify the end of the range as `None`. Furthermore, since we can limit the number of returned `block_headers` if the response is too large, it may happen following case, user request headers `[request_start, request_end]`, and the receivers `[request_start, response_end]` where `response_end < request_end` at that moment user cannot know if we have limited the size of response or bitcoin canister does not have rest of the block headers. So to distinguish two cases I will prefer to return an error.

THLO: Yeah, that makes sense.
@ielashi @dsarlis, do you agree with @dragoljub-duric's proposal?

ielashi: The proposal SGTM

spec/_attachments/ic.did
```diff
@@ -131,18 +133,36 @@ type bitcoin_get_balance_args = record {
...
+
 type bitcoin_send_transaction_args = record {
     transaction : blob;
     network : bitcoin_network;
 };
 
-type millisatoshi_per_byte = nat64;
+type bitcoin_get_block_headers_args = record {
+    start_height : bitcoin_block_height;
+    end_height : opt bitcoin_block_height;
```
dragoljub-duric, line 159: @THLO Should we add a network field in order to know where to route the request?

mraszyk: > this endpoint needs one as well

@THLO Should the candid file be updated? The PR is marked as final already so could you please also follow up with the SDK team?

THLO: Updated!

dsarlis (approved):

PR #302, opened 2024-04-16
fix: Correct the naming of the trustworthy metric: node num_blocks_total

sasa-tomic:
The current name is confusing not only for me but also for the community member(s). I propose we change it to make it more specific.

image

link: https://matrix.to/#/!jUsknEqNdJobySpXuN:matrix.org/$wmamJVHU4QTXBWvNx-Rbq8NubEfHyAymhfdjMDsrWfU?via=matrix.org&via=greensteps.cn&via=rudd-o.com

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -146,7 +146,7 @@ type millisatoshi_per_byte = nat64;
 
 type node_metrics = record {
     node_id : principal;
-    num_blocks_total : nat64;
+    num_blocks_proposed_total : nat64;
     num_block_failures_total : nat64;
 };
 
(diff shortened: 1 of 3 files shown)


Dfinity-Bjoern:
Is this already in production? Then this would be a breaking change.

sasa-tomic:
> Is this already in production? Then this would be a breaking change.

@Dfinity-Bjoern correct, but this feature is marked as experimental, so there is no better time to fix it than now.
Also, the only user of these metrics (that we know of) is the DRE tool.

Dfinity-Bjoern:
Your right, thanks for pointing this out!

Dfinity-Bjoern (approved):

mraszyk:
@sasa-tomic Please do not merge until the feature is rolled out in production.

sasa-tomic:
The replica code change is in https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/19153

mraszyk:
This feature is contained in the [upcoming](https://dashboard.internetcomputer.org/proposal/129746) IC release so merging this MR now.

PR #304, opened 2024-04-29
fix: ic0.call_new can't trap due to insufficient cycles

mraszyk:
This PR fixes the specification of `ic0.call_new` that can't trap due to insufficient cycles (cycles for response execution are only checked in `ic0.call_perform`). Moreover, it makes the error code returned by `ic0.call_perform` implementation-specific to align with the actual replica implementation.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1594,7 +1594,7 @@ There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` a
 
     This concludes assembling the call. It queues the call message to the given destination, but does not actually act on it until the current WebAssembly function returns without trapping.
 
-    This deducts `MAX_CYCLES_PER_RESPONSE` cycles from the canister balance and sets them aside for response processing. This will trap if not sufficient cycles are available.
+    This deducts `MAX_CYCLES_PER_RESPONSE` cycles from the canister balance and sets them aside for response processing.
 
     If the function returns `0` as the `err_code`, the IC was able to enqueue the call. In this case, the call will either be delivered, returned because the destination canister does not exist or returned because of an out of cycles condition. This also means that exactly one of the reply or reject callbacks will be executed.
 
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -1596,6 +1594,8 @@ There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` a
 
     This concludes assembling the call. It queues the call message to the given destination, but does not actually act on it until the current WebAssembly function returns without trapping.
 
+    This deducts `MAX_CYCLES_PER_RESPONSE` cycles from the canister balance and sets them aside for response processing. This will trap if not sufficient cycles are available.
```
dsarlis, line 1597:
```suggestion
    This deducts `MAX_CYCLES_PER_RESPONSE` cycles from the canister balance and sets them aside for response processing.
```

dsarlis: This is not happening in the implementation and I think it's also wrong to do that. If we can't make the reservation we should return a non-zero error code to signal that there were not enough cycles available.

mraszyk: Good point! I've fixed that now (also in the formal specs).

dsarlis (approved):

spec/index.md
```diff
@@ -6477,7 +6476,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       ) < 0 or system_cannot_do_this_call_now()
       then
         discard_pending_call<es>()
-        return 1
+        return <implementation-specific>
```
oggy-dfin, line 6479: I guess this should be non-zero, but I understand it's difficult to specify given the convention we adopted elsewhere.

Dfinity-Bjoern: I don't think the error can be implementation specific as this happens in the deterministic execution?

mraszyk: sure, but we don't want to specify it (just like the amount of cycles consumed which also happens in the deterministic execution, but we mark it as implementation-specific since we don't cover it in this specification)

Dfinity-Bjoern (approved):
Thanks for the correction!

PR #306, opened 2024-05-13
feat: Introduce replicated and non-replicated execution modes

dsarlis:
Replicated and non-replicated execution modes are already discussed in the spec but they are not formally specified. The two modes are already mentioned to distinguish the results of `ic0.data_certificate_*` calls and also in the case of the bitcoin query APIs on the management canister.

This PR formally introduces the concept of the two modes and updates the system API to be more explicit about when certain APIs are available in replicated vs non-replicated queries.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1321,11 +1321,11 @@ The following sections describe various System API functions, also referred to a
     ic0.msg_cycles_available128 : (dst : i32) -> ();                            // U Rt Ry
     ic0.msg_cycles_refunded : () -> i64;                                        // Rt Ry
     ic0.msg_cycles_refunded128 : (dst : i32) -> ();                             // Rt Ry
-    ic0.msg_cycles_accept : (max_amount : i64) -> (amount : i64);               // U RQ Rt Ry
+    ic0.msg_cycles_accept : (max_amount : i64) -> (amount : i64);               // U Rt Ry
     ic0.msg_cycles_accept128 : (max_amount_high : i64, max_amount_low: i64, dst : i32)
-                           -> ();                                               // U RQ Rt Ry
+                           -> ();                                               // U Rt Ry
 
-    ic0.cycles_burn128 : (amount_high : i64, amount_low : i64, dst : i32) -> ();               // I G U RQ Ry Rt C T
+    ic0.cycles_burn128 : (amount_high : i64, amount_low : i64, dst : i32) -> ();               // I G U Ry Rt C T
 
     ic0.canister_self_size : () -> i32;                                         // *
     ic0.canister_self_copy : (dst : i32, offset : i32, size : i32) -> ();       // *
@@ -6298,7 +6298,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       copy_cycles_to_canister<es>(dst, amount.to_little_endian_bytes())
 
     ic0.msg_cycles_accept<es>(max_amount : i64) : i64 =
-      if es.context ∉ {U, RQ, Rt, Ry} then Trap {cycles_used = es.cycles_used;}
+      if es.context ∉ {U, Rt, Ry} then Trap {cycles_used = es.cycles_used;}
       let amount = min(max_amount, es.cycles_available)
       es.cycles_available := es.cycles_available - amount
       es.cycles_accepted := es.cycles_accepted + amount
@@ -6306,7 +6306,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       return amount
 
     ic0.msg_cycles_accept128<es>(max_amount_high : i64, max_amount_low : i64, dst : i32) =
-      if es.context ∉ {U, RQ, Rt, Ry} then Trap {cycles_used = es.cycles_used;}
+      if es.context ∉ {U, Rt, Ry} then Trap {cycles_used = es.cycles_used;}
       let max_amount = max_amount_high * 2^64 + max_amount_low
       let amount = min(max_amount, es.cycles_available)
       es.cycles_available := es.cycles_available - amount
@@ -6315,7 +6315,7 @@ The pseudo-code below does *not* explicitly enforce the restrictions of which im
       copy_cycles_to_canister<es>(dst, amount.to_little_endian_bytes())
 
     ic0.cycles_burn128<es>(amount_high : i64, amount_low : i64, dst : i32) =
-      if es.context ∉ {I, G, U, RQ, Ry, Rt, C, T} then Trap {cycles_used = es.cycles_used;}
+      if es.context ∉ {I, G, U, Ry, Rt, C, T} then Trap {cycles_used = es.cycles_used;}
       let amount = amount_high * 2^64 + amount_low
       let burned_amount = min(
         amount,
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -1297,31 +1297,35 @@ Callbacks are addressed by their table index (as a proxy for a Wasm `funcref`).
 
 In the reply callback of a [inter-canister method call](#system-api-call), the argument refers to the response to that call. In reject callbacks, no argument is available.
 
+### Replicated and Non-Replicated execution mode
+
+Canister methods can be executed either in *replicated mode* where the method runs on all subnet nodes and the results go through consensus or in *non-replicated mode* where the method runs on a single node and the result does not go through consensus. Update methods can only be run in replicated mode (as they need to preserve state changes) while composite query methods can only be executed in non-replicated mode. Query methods can be executed in either mode.
```
mraszyk, line 1302:
```suggestion
Canister methods can be executed either in *replicated mode* where the method runs on all subnet nodes and the results go through consensus or in *non-replicated mode* where the method runs on a single node and the result does not go through consensus. Update methods can only be run in replicated mode (as they need to preserve state changes) while composite query and inspect message methods can only be executed in non-replicated mode. Query methods can be executed in either mode.
```

mraszyk: On the other hand, we aren't pointing out all replicated mode methods either, so I'm not sure about this suggestion after all.

mraszyk:
We should also update this line
```
`*` = `I G U Q CQ Ry Rt CRy CRt C CC F T` (NB: Not `(start)`)
```

mraszyk:
Some of the conditions like
```
 ic0.msg_reply_data_append<es>(src : i32, size : i32) =
      if es.context ∉ {U, Q, CQ, Ry, Rt, CRy, CRt} then Trap {cycles_used = es.cycles_used;}
```
are still not fixed.

spec/index.md
```diff
@@ -1361,11 +1365,11 @@ The following sections describe various System API functions, also referred to a
...
     ic0.certified_data_set : (src: i32, size: i32) -> ();                       // I G U Ry Rt T
     ic0.data_certificate_present : () -> i32;                                   // *
-    ic0.data_certificate_size : () -> i32;                                      // Q CQ
-    ic0.data_certificate_copy : (dst: i32, offset: i32, size: i32) -> ();       // Q CQ
+    ic0.data_certificate_size : () -> i32;                                      // NRQ CQ
+    ic0.data_certificate_copy : (dst: i32, offset: i32, size: i32) -> ();       // NRQ CQ
 
     ic0.time : () -> (timestamp : i64);                                         // *
-    ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U Ry Rt C T
+    ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U RQ Ry Rt C T
```
mraszyk, line 1372: This isn't implemented yet, right?

mraszyk: Also it might have unexpected consequences as e.g., the Rust timers CDK needs to alter state when registering a new timer and these state changes will be rolled back in replicated query mode.

spec/index.md
dsarlis: I think we should figure out what we want to call out here. We can list all methods and whether they can be run in replicated or non-replicated mode. I started with something to have a discussion.

spec/index.md
```diff
@@ -1297,31 +1297,35 @@ Callbacks are addressed by their table index (as a proxy for a Wasm `funcref`).
...
     ic0.msg_cycles_refunded : () -> i64;                                        // Rt Ry
     ic0.msg_cycles_refunded128 : (dst : i32) -> ();                             // Rt Ry
-    ic0.msg_cycles_accept : (max_amount : i64) -> (amount : i64);               // U Rt Ry
+    ic0.msg_cycles_accept : (max_amount : i64) -> (amount : i64);               // U RQ Rt Ry
     ic0.msg_cycles_accept128 : (max_amount_high : i64, max_amount_low: i64, dst : i32)
-                           -> ();                                               // U Rt Ry
+                           -> ();                                               // U RQ Rt Ry
 
-    ic0.cycles_burn128 : (amount_high : i64, amount_low : i64, dst : i32) -> ();               // I G U Ry Rt C T
+    ic0.cycles_burn128 : (amount_high : i64, amount_low : i64, dst : i32) -> ();               // I G U RQ Ry Rt C T
```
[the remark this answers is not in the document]
dsarlis: No, not yet. First we agree on the spec PR and then we make the necessary changes.

spec/index.md
dsarlis: No, and I was actually unsure whether we should do it. It just felt that there's no strong reason to disallow using this API in replicated queries but indeed it's not going to be usable by the Rust CDK so maybe we I revert this for now.

dsarlis: Reverted this change.

spec/index.md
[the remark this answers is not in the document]
dsarlis: My main goal was to allow canisters to accept cycles in replicated queries (which we discussed in the interface spec meeting last week). Then it felt like `ic0.cycles_burn128` could also be allowed in replicated queries and it would allow you to burn cycles, i.e. the cycles changes would indeed be committed to the replicated state.

Why do you think that would be confusing? It's certainly a change in behaviour but especially accepting cycles in replicated queries allows people to build reasonable business models for their callers by requiring payments for accessing endpoints.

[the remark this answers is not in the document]
mraszyk: ok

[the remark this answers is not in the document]
dsarlis: As discussed during the meeting yesterday, I removed the changes to the cycles API to keep this PR only focused in formally introducing the two modes without changes in current behaviour.

spec/index.md
dsarlis: I added a table with all canister methods and the mode that they can be executed in. Was there anything else we said we wanted to include?

spec/index.md
```diff
@@ -1297,21 +1297,39 @@ Callbacks are addressed by their table index (as a proxy for a Wasm `funcref`).
 
 In the reply callback of a [inter-canister method call](#system-api-call), the argument refers to the response to that call. In reject callbacks, no argument is available.
 
+### Replicated and Non-Replicated execution mode
+
+Canister methods can be executed either in *replicated* mode where the method runs on all subnet nodes and the results go through consensus or in *non-replicated* mode where the method runs on a single node and the result does not go through consensus. The trade-off between replicated and non-replicated mode is therefore one between speed of getting a result and trustworthiness of the result.
```
mraszyk, line 1302:
```suggestion
Canister methods can be executed either in *replicated* mode where the method runs on all subnet nodes and the results go through consensus or in *non-replicated* mode where the method runs on a single node and the result does not go through consensus. The trade-off between replicated and non-replicated mode is therefore one between the result's latency and trustworthiness.
```

mraszyk (approved):

spec/index.md
mraszyk: We also discussed to include an overview of what changes get persisted in the individual cases, but we can also do that in a follow up MR to not block the improvement this MR already brings.

dsarlis: Thanks for the reminder. I suggest we cover this in a follow up MR to unblock merging here.

PR #307, opened 2024-05-29
Introduce `snapshot_id` in canister change details

AlexandraZapuc:


spec/index.md
```diff
@@ -5494,6 +5496,7 @@ New_canister_history = {
     canister_version = S.canister_version[A.canister_id] + 1
     origin = change_origin(M.caller, A.sender_canister_version, M.origin);
     details = LoadSnapshot {
+      snapshot_id = A.snapshot_id
```
mraszyk, line 5499:
```suggestion
      snapshot_id = S.snapshots[A.canister_id].snapshot_id
```
for consistency with the two lines that follow (I know we check for equality with what you had before)

PR #312, opened 2024-06-12
feat: remove bitcoin_get_balance_query and bitcoin_get_utxos_query

mraszyk:
This PR removes the management canister query calls `bitcoin_get_balance_query` and `bitcoin_get_utxos_query`. According to the metrics, they are not being used and the query call signature verification cannot be performed according to the specification (the read state request to fetch the replica public keys is gonna fail).


Dfinity-Bjoern (approved):
What a beautiful PR!

dsarlis (approved):

PR #313, opened 2024-06-12
[FINAL] feat: Allow accepting and burning cycles in replicated queries

dsarlis:
Given a previous change that introduced formally the concepts of replicated and non-replicated queries, we can now extend the spec to allow accepting and burning cycles in replicated queries. This allows for creating reasonable business models for canisters by requiring callers in replicated mode to always make payments for accessing their endpoints.


spec/index.md
```diff
@@ -6242,16 +6243,18 @@ Finally we can specify the abstract `CanisterModule` that models a concrete WebA
...
               wasm_state = wasm_state;
               params = empty_params with { arg = arg; caller = caller; sysenv }
               balance = sysenv.balance
+              cycles_available = available;
               context = Q
             }
           try func<es>() with Trap then Trap {cycles_used = es.cycles_used;}
           Return {
             response = es.response;
+            cycles_accepted = es.cycles_accepted;
```
mraszyk, line 6257: This field should be read in the Message execution transition:
```
query_as_update(f, arg, env) = λ wasm_state →
  match f(arg, env)(wasm_state) with
    Trap trap → Trap trap
    Return res → Return {
      new_state = wasm_state;
      new_calls = [];
      new_certified_data = NoCertifiedData;
      new_global_timer = NoGlobalTimer;
      response = res.response;
      cycles_accepted = res.cycles_accepted;
      cycles_used = res.cycles_used;
    }
```

spec/index.md
```diff
@@ -3689,7 +3690,7 @@ Available = S.call_contexts[M.call_contexts].available_cycles
 )
 or
 ( M.entry_point = PublicMethod Name Caller Arg
-  F = query_as_update(Mod.query_methods[Name], Arg, Caller, Env)
+  F = query_as_update(Mod.query_methods[Name], Arg, Caller, Env, Available)
```
mraszyk, line 3693: We also need to change this:
```
    query_as_update(f, arg, caller, env, available) = λ wasm_state →
      match f(arg, caller, env, available)(wasm_state) with
```
the missing "caller" is a bug from before this PR.

dsarlis: Done, although I'm not sure I understand the notation really. It feels like something is missing after "with".

spec/index.md
dsarlis: Done.

spec/index.md
mraszyk: Oh, sorry for the confusion, we need to perform the change at the place where the function `query_as_update` is defined (further below), not at this line where I attached the comment (I can't open a comment on a line that is far from something that is changed by this PR).

mraszyk: the change must happen a few lines above your change of `cycles_accepted = res.cycles_accepted;`

dsarlis: Oh, of course, this makes sense now and I see what you mean about the missing "caller".

dsarlis:
@mraszyk I realized while working on the replica side of things that since certain parts of the canister's state, more specifically its system state. will change during query execution, the canister version will be bumped. I suppose this is fine but do you think we should make any amendments in the spec? Or is the current section on canister version "vague" enough to allow for this?

mraszyk:
> @mraszyk I realized while working on the replica side of things that since certain parts of the canister's state, more specifically its system state, will change during replicated query execution, the canister version will be bumped. I suppose this is fine but do you think we should make any amendments in the spec? Or is the current section on canister version "vague" enough to allow for this?

The spec says that "The system can arbitrarily increment the canister version also if the canister's code, settings, running status, and memory do not change." so we're technically fine, but we might want to update the section on "Canister version" to mention the case of replicated queries explicitly.

dsarlis:
> > @mraszyk I realized while working on the replica side of things that since certain parts of the canister's state, more specifically its system state, will change during replicated query execution, the canister version will be bumped. I suppose this is fine but do you think we should make any amendments in the spec? Or is the current section on canister version "vague" enough to allow for this?
> 
> The spec says that "The system can arbitrarily increment the canister version also if the canister's code, settings, running status, and memory do not change." so we're technically fine, but we might want to update the section on "Canister version" to mention the case of replicated queries explicitly.

Made an update. See my last commit and let me know if it's ok.

spec/index.md
```diff
@@ -1618,7 +1618,7 @@ This function allows a canister to find out if it is running, stopping or stoppe
...
+For each canister, the system maintains a *canister version*. Upon canister creation, it is set to 0, and it is **guaranteed** to be incremented upon every change of the canister's code, settings, running status (Running, Stopping, Stopped), and memory (WASM and stable memory), i.e., upon every successful management canister call of methods `update_settings`, `install_code`, `install_chunked_code`, `uninstall_code`, `start_canister`, and `stop_canister` on that canister, code uninstallation due to that canister running out of cycles, canister's running status transitioning from Stopping to Stopped, and successful execution of update methods, response callbacks, heartbeats, global timers and replicated query methods. The system can arbitrarily increment the canister version also if the canister's code, settings, running status, and memory do not change.
```
mraszyk, line 1621:
```suggestion
For each canister, the system maintains a *canister version*. Upon canister creation, it is set to 0, and it is **guaranteed** to be incremented upon every change of the canister's code, settings, running status (Running, Stopping, Stopped), cycles balance, and memory (WASM and stable memory), i.e., upon every successful management canister call of methods `update_settings`, `install_code`, `install_chunked_code`, `uninstall_code`, `start_canister`, and `stop_canister` on that canister, code uninstallation due to that canister running out of cycles, canister's running status transitioning from Stopping to Stopped, and successful execution of update methods, replicated query methods, response callbacks, heartbeats, and global timers. The system can arbitrarily increment the canister version also if the canister's code, settings, running status, and memory do not change.
```

mraszyk: @dsarlis I've made the following changes here:
- Reordered the list of messages guaranteed to bump the canister version (bringing update and query methods next to each other).
- Mentioned "cycles balance" as an additional parameter whose change implies a bump in the canister version.

mraszyk:
Superseded by https://github.com/dfinity/portal/pull/3760

PR #316, opened 2024-06-14
[not FINAL] feat: new System API call ic0.call_cycles_add128_up_to

mraszyk:
This PR adds a new System API call `ic0.call_cycles_add128_up_to` moving the maximum amount of cycles from the canister cycles balance onto an inter-canister call up to a specified upper bound.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1378,6 +1378,8 @@ The following sections describe various System API functions, also referred to a
     ic0.call_data_append : (src : i32, size : i32) -> ();                       // U CQ Ry Rt CRy CRt T
     ic0.call_cycles_add : (amount : i64) -> ();                                 // U Ry Rt T
     ic0.call_cycles_add128 : (amount_high : i64, amount_low: i64) -> ();        // U Ry Rt T
+    ic0.call_cycles_add128_up_to : (max_amount_high : i64, max_amount_low: i64, dst: i32)
+      -> ();                                                                    // U Ry Rt T
     ic0.call_perform : () -> ( err_code : i32 );                                // U CQ Ry Rt CRy CRt T
 
     ic0.stable_size : () -> (page_count : i32);                                 // * s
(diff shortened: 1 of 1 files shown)


Dfinity-Bjoern (commented):
Looks good to me for marking it final, would like to get the opinion from @dsarlis as well.

dsarlis (approved):

spec/index.md
```diff
@@ -1698,6 +1700,24 @@ There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` a
...
+
+    - The cycle balance of the canister after transferring cycles does not decrease below the canister's freezing limit.
+
+    The cycles are deducted from the balance as shown by `ic0.canister_cycles_balance128` immediately, and moved back if the call cannot be performed (e.g. if `ic0.call_perform` signals an error, if the canister invokes `ic0.call_new`, or returns without calling `ic0.call_perform`).
+
+    This may be called multiple times between `ic0.call_new` and `ic0.call_perform`, each time possibly moving more cycles onto the call.
+
+    This system call traps if there is no call under construction, i.e., if not called between `ic0.call_new` and `ic0.call_perform`.
+
+    This system call also copies the actual amount of cycles that were moved onto the call represented by a 128-bit value starting at the location `dst` in the canister memory.
```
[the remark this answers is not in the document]
mraszyk: How about adding:
```
This amount might be lower than the 128-bit value obtained by combining the `max_amount_high` and `max_amount_low` parameters.
```
Currently, it sounds like the amount might be lower than _two_ distinct values.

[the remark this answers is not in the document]
dsarlis: Sure, just wanted to make sure that it's clear the value returned when calling this API might be lower than the amount requested.

spec/index.md
```diff
@@ -1652,6 +1654,25 @@ There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` a
...
 
+-   `ic0.call_cycles_add128_up_to : (max_amount_high : i64, max_amount_low : i64, dst: i32) -> ()`
+
+    This system call moves cycles from the canister balance onto the call under construction, to be transferred with that call.
+
+    This moves the maximum possible amount of cycles onto the call, up to these constraints:
+
+    - It moves no more cycles than represented by a 128-bit value which can be obtained by combining the `max_amount_high` and `max_amount_low` parameters.
+
+    - The cycle balance of the canister after transferring cycles does not decrease below the canister's freezing limit.
```
sesi200, line 1665: Found an issue while implementing this:
- `call_cycles_add128_up_to(u128::MAX)` puts the canister to just above the freezing limit
- The following `call_perform` tries to charge cycles for sending the message and processing the response. This fails because it would put the canister below the freezing threshold.

We could say that `call_cycles_add128_up_to` moves cycles such that a subsequent `call_perform` does not trap if no extra bytes were added to the message, or we could say that it moves cycles such that a `call_perform` with _any_ amount of bytes would succeed.

mraszyk: @dsarlis Could we avoid the issue described above by withdrawing the base fee at ic0.call_new and the argument free at each ic0.call_data_append? In my opinion, this would be cleaner than trying to figure out the amount that must be kept around for ic0.call_perform to succeed.

mraszyk: A potential difficulty is what to do if the cycles can't be withdrawn as those other system APIs (ic0.call_new and ic0.call_data_append) don't return any values and thus we could only trap which would be a major breaking change.

mraszyk: So saying that "ic0.call_cycles_add128_up_to moves cycles such that a subsequent call_perform does not trap if no extra bytes were added to the message" seems to the best way out after all.

dsarlis: > we could say that it moves cycles such that a call_perform with any amount of bytes would succeed.

This might be a good option too. It removes the need to avoid adding more bytes to the call and might be more useful for end users. The extra loss in cycles should be minimal in the grand scheme of things.

sesi200: Wording suggestion:

> `ic0.call_cycles_add128_up_to` moves cycles to the call under construction such that a subsequent `call_perform` does not fail because of an insufficient cycles balance (if no extra bytes were added to the message)

mraszyk: Done: https://github.com/dfinity/interface-spec/pull/316/commits/d692008cc315bd8576e2c4522d8afabf9a55af89

Could you please take a look if this commit makes sense to you?

dsarlis: Lgtm.

spec/index.md
```diff
@@ -6812,24 +6834,44 @@ ic0.call_cycles_add<es>(amount : i64) =
...
+  if es.pending_call = NoPendingCall then Trap {cycles_used = es.cycles_used;}
+  let max_amount = max_amount_high * 2^64 + max_amount_low
+  let amount = min(max_amount, liquid_balance(
+    es.balance - MAX_CYCLES_PER_RESPONSE,
+    es.params.sysenv.reserved_balance,
+    freezing_limit(
+      es.params.sysenv.compute_allocation,
+      es.params.sysenv.memory_allocation,
+      es.params.sysenv.freezing_threshold,
+      memory_usage_wasm_state(es.wasm_state) + es.params.sysenv.memory_usage_raw_module + es.params.sysenv.memory_usage_canister_history,
```
mraszyk, line 6867:
```suggestion
      memory_usage_wasm_state(es.wasm_state) + es.params.sysenv.memory_usage_raw_module + es.params.sysenv.memory_usage_canister_history + max(|es.pending_call.arg|, 2 * 1024 * 1024),
```

mraszyk: Maybe not needed though as we don't model message memory anyway.

spec/index.md
```diff
@@ -1652,6 +1654,26 @@ There must be at most one call to `ic0.call_on_cleanup` between `ic0.call_new` a
...
+-   `ic0.call_cycles_add128_up_to : (max_amount_high : i64, max_amount_low : i64, dst: i32) -> ()`
+
+    This system call moves cycles from the canister balance onto the call under construction, to be transferred with that call.
+
+    This moves the maximum possible amount of cycles onto the call, up to these constraints:
+
+    - It moves no more cycles than represented by a 128-bit value which can be obtained by combining the `max_amount_high` and `max_amount_low` parameters.
+
+    - A subsequent `ic0.call_perform` does not fail because of insufficient cycles balance
+      (if no extra bytes were added to the message via `ic0.call_data_append` between `ic0.call_cycles_add128_up_to` and `ic0.call_perform`)
```
mraszyk, line 1666:
```suggestion
      (if the canister makes no memory allocations and no extra bytes were added to the message via `ic0.call_data_append` between `ic0.call_cycles_add128_up_to` and `ic0.call_perform`)
```

Dfinity-Bjoern: @dsarlis this seems a bit unpredictable to use. In higher-level languages, it seems difficult to guarantee that no extra memory is allocated. Should we maybe extend the implementation so that we can drop the additional conditions (by doing the computation of actual cycles to be sent to `call_perform`)?

Dfinity-Bjoern: @mraszyk just pointed out that this would violate
> This system call also copies the actual amount of cycles that were moved onto the call represented [...]
which seems to make it necessary to eagerly compute the amount of cycles attached. Not sure which is the best path forward here.

mraszyk: It is possible to compute the actual amount of cycles attached by taking the difference of the canister's cycles balance before and after calling the system API `ic0.call_cycles_add128_up_to` (confirmed empirically). So dropping the destination pointer should not make the API substantially harder to use. The "trick" with computing the cycles balance difference does not work across `ic0.call_perform` though (as xnet fees are charged, too). Moreover, it is not clear what the semantics of multiple `ic0.call_cycles_add128_up_to` should be if we only attach the cycles during `ic0.call_perform` (the sum vs the maximum?). Hence, delaying the actual attachment to `ic0.call_perform` might not be the best path forward either.

I'd suggest to simply drop the destination pointer if the CDK teams could confirm that allocations are unlikely in between `ic0.call_cycles_add128_up_to` and `ic0.call_perform`. (Allocations happening after `ic0.call_perform` are fine as long as they are covered by the unused execution cycles that are refunded first.)

PR #317, opened 2024-06-18
chore: make it explicit if external users may call management canister methods

mraszyk:
This PR makes it explicit if external users may call management canister methods.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1974,6 +1974,8 @@ The binary encoding of arguments and results are as per Candid specification.
 
 ### IC method `create_canister` {#ic-create_canister}
 
+This method can only be called by canisters, i.e., it must not be called by external users via ingress messages.
+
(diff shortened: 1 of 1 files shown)


spec/index.md
```diff
@@ -2416,6 +2464,8 @@ The [standard nearest-rank estimation method](https://en.wikipedia.org/wiki/Perc
 
 ### IC method `fetch_canister_logs` {#ic-fetch_canister_logs}
 
+This method can only be called by external users via non-replicated query calls, i.e., it cannot be called by canisters, as replicated update calls, and as nested composite query calls.
```
Dfinity-Bjoern, line 2467:
```suggestion
This method can only be called by external users via non-replicated query calls, i.e., it cannot be called by canisters, as replicated calls, and from composite query calls.
```

PR #324, opened 2024-07-03
chore: fix broken links in headings

dfx-json:
- Update all headings to use `{#heading}` syntax
- Fix broken links where the heading name was changed
- Update docusaurus version to latest


mraszyk (approved):
Thanks you! @dfx-json

PR #326, opened 2024-07-27
[FINAL] feat: [EXC-1676] add allowed viewers variant to canister's log visibility

maksymar:
This PR adds `allowed_viewers` variant to canister's `log_visibility` which allows to fetch logs by specified number of principals.

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -3,8 +3,8 @@ type wasm_module = blob;
 
 type log_visibility = variant {
     controllers;
-    allowed_viewers : vec principal;
     public;
+    allowed_viewers : vec principal;
 };
 
 type canister_settings = record {
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -3,6 +3,7 @@ type wasm_module = blob;
 
 type log_visibility = variant {
     controllers;
+    allowlist : opt vec principal;
     public;
 };
 
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -3,7 +3,7 @@ type wasm_module = blob;
 
 type log_visibility = variant {
     controllers;
-    allowlist : opt vec principal;
+    allowlist : vec principal;
     public;
 };
 
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -3,7 +3,7 @@ type wasm_module = blob;
 
 type log_visibility = variant {
     controllers;
-    allowlist : vec principal;
+    allowed_viewers : vec principal;
     public;
 };
 
--- a/spec/index.md
+++ b/spec/index.md
@@ -2607,7 +2607,7 @@ Logs persist across canister upgrades and they are deleted if the canister is re
 The log visibility is defined in the `log_visibility` field of `canister_settings` and can be one of the following variants:
 
 - `controllers`: only canister's controllers can fetch logs (by default)
-- `allowed_viewers` (`vec principal`): only provided list of principals can fetch logs, the maximum length of the list is 10
+- `allowed_viewers` (`vec principal`): only the provided list of principals can fetch logs, the maximum length of the list is 10
 - `public`: everyone can fetch logs
 
 A single log is a record with the following fields:
--- a/spec/index.md
+++ b/spec/index.md
@@ -2607,7 +2607,7 @@ Logs persist across canister upgrades and they are deleted if the canister is re
 The log visibility is defined in the `log_visibility` field of `canister_settings` and it can be:
 
 - `controllers`: only visible to the canister's controllers (by default)
-- `allowlist`: visible to the provided list of principals
+- `allowlist`: visible to the provided list of principals, max number is limited by 10
 - `public`: visible to everyone
 
 A single log is a record with the following fields:
--- a/spec/index.md
+++ b/spec/index.md
@@ -2506,9 +2506,15 @@ Logs persist across canister upgrades and they are deleted if the canister is re
 
 The log visibility is defined in the `log_visibility` field of `canister_settings` and can be one of the following variants:
 
-- `controllers`: only canister's controllers can fetch logs (by default)
-- `public`: everyone can fetch logs
-- `allowed_viewers` (`vec principal`): only principals in the provided list and the canister's controllers can fetch logs, the maximum length of the list is 10
+- `controllers`: only the canister's controllers can fetch logs (default);
+- `public`: everyone can fetch logs;
+- `allowed_viewers` (`vec principal`): only principals in the provided list and the canister's controllers can fetch logs, the maximum length of the list is 10.
+
+A single log is a record with the following fields:
+
+- `idx` (`nat64`): the unique sequence number of the log for this particular canister;
+- `timestamp_nanos` (`nat64`): the timestamp as nanoseconds since 1970-01-01 at which the log was recorded;
+- `content` (`blob`): the actual content of the log;
 
 :::warning
 
@@ -5903,9 +5909,9 @@ Q.method_name = 'fetch_canister_logs'
 Q.arg = candid(A)
 A.canister_id = effective_canister_id
 (S[A.canister_id].canister_log_visibility = Public)
-  or 
+  or
   (S[A.canister_id].canister_log_visibility = Controllers and Q.sender in S[A.canister_id].controllers)
-  or 
+  or
   (S[A.canister_id].canister_log_visibility = AllowedViewers Principals and (Q.sender in S[A.canister_id].controllers or Q.sender in Principals))
 
 ```
--- a/spec/index.md
+++ b/spec/index.md
@@ -2603,7 +2603,12 @@ The canister logs are *not* collected in canister methods running in non-replica
 The total size of all returned logs does not exceed 4KiB.
 If new logs are added resulting in exceeding the maximum total log size of 4KiB, the oldest logs will be removed.
 Logs persist across canister upgrades and they are deleted if the canister is reinstalled or uninstalled.
-The log visibility is defined in the `log_visibility` field of `canister_settings`: logs can be either public (visible to everyone) or only visible to the canister's controllers (by default).
+
+The log visibility is defined in the `log_visibility` field of `canister_settings` and it can be:
+
+- `controllers`: only visible to the canister's controllers (by default)
+- `allowlist`: visible to the provided list of principals
+- `public`: visible to everyone
 
 A single log is a record with the following fields:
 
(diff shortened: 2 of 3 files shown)


spec/index.md
```diff
@@ -4203,6 +4211,8 @@ New_canister_history = {
 
 if A.settings.log_visibility is not null:
   New_canister_log_visibility = A.settings.log_visibility
+  if New_canister_log_visibility is AllowList:
+    New_canister_log_allowlist = A.settings.log_visibility ?? how to get allowlist?
```
maksymar, line 4215: @mraszyk could you please suggest on how to write down updating allowlist in this notation?

spec/index.md
```diff
@@ -2603,7 +2603,12 @@ The canister logs are *not* collected in canister methods running in non-replica
 The total size of all returned logs does not exceed 4KiB.
 If new logs are added resulting in exceeding the maximum total log size of 4KiB, the oldest logs will be removed.
 Logs persist across canister upgrades and they are deleted if the canister is reinstalled or uninstalled.
-The log visibility is defined in the `log_visibility` field of `canister_settings`: logs can be either public (visible to everyone) or only visible to the canister's controllers (by default).
+
+The log visibility is defined in the `log_visibility` field of `canister_settings` and it can be:
+
+- `controllers`: only visible to the canister's controllers (by default)
+- `allowlist`: visible to the provided list of principals, max number is limited by 10
```
mraszyk, line 2610:
```suggestion
- `allowlist`: visible to the provided list of principals, the maximum length of the list is 10
```

spec/index.md
```diff
@@ -3277,6 +3283,7 @@ S = {
   certified_data: CanisterId ↦ Blob;
   canister_history: CanisterId ↦ CanisterHistory;
   canister_log_visibility: CanisterId ↦ CanisterLogVisibility;
+  canister_log_allowlist: CanisterId ↦ Set Principal;
```
mraszyk, line 3286: I wouldn't introduce a new mapping `canister_log_allowlist` here since `canister_log_visibility` contains all the necessary information. This suggestion then also makes the question how to update this mapping obsolete. Please see my suggestion below on how to formulate the condition if a request sender is authorized to fetch the canister logs.

spec/index.md
```diff
@@ -5878,7 +5893,7 @@ Q.canister_id = ic_principal
 Q.method_name = 'fetch_canister_logs'
 Q.arg = candid(A)
 A.canister_id = effective_canister_id
-S[A.canister_id].canister_log_visibility = Public or Q.sender in S[A.canister_id].controllers
+S[A.canister_id].canister_log_visibility = Public or Q.sender in S[A.canister_id].controllers or Q.sender in S[A.canister_id].canister_log_allowlist
```
mraszyk, line 5896:
```suggestion
S[A.canister_id].canister_log_visibility = Public or (S[A.canister_id].canister_log_visibility = allowlist Principals and Q.sender in Principals) or Q.sender in S[A.canister_id].controllers
```

spec/_attachments/ic.did
```diff
@@ -3,6 +3,7 @@ type wasm_module = blob;
 
 type log_visibility = variant {
     controllers;
+    allowlist : opt vec principal;
```
mraszyk, line 6: Why is the vector optional? What is the semantics if no list of principals is provided?

spec/index.md
maksymar: done.

spec/_attachments/ic.did
maksymar: that was added by mistake. thanks for pointing this out. removed.

spec/index.md
maksymar: Removed `New_canister_log_allowlist`.

spec/index.md
maksymar: Ok, removed `canister_log_allowlist`.

spec/index.md
maksymar: Done.

spec/index.md
```diff
@@ -3247,7 +3247,7 @@ CanisterHistory = {
 }
 CanisterLogVisibility
   = Controllers
-  | AllowList
+  | AllowedViewers
```
mraszyk, line 3250:
```suggestion
  | AllowedViewers [Principal]
```

spec/index.md
```diff
@@ -5893,7 +5884,7 @@ Q.canister_id = ic_principal
 Q.method_name = 'fetch_canister_logs'
 Q.arg = candid(A)
 A.canister_id = effective_canister_id
-S[A.canister_id].canister_log_visibility = Public or Q.sender in S[A.canister_id].controllers or Q.sender in S[A.canister_id].canister_log_allowlist
+S[A.canister_id].canister_log_visibility = Public or (S[A.canister_id].canister_log_visibility = AllowedViewers Principals and Q.sender in Principals) or Q.sender in S[A.canister_id].controllers
```
mraszyk, line 5887:
```suggestion
S[A.canister_id].canister_log_visibility = Public or (S[A.canister_id].canister_log_visibility = AllowedViewers Principals and Q.sender in Principals) or (S[A.canister_id].canister_log_visibility = Controllers and Q.sender in S[A.canister_id].controllers)
```
or can controllers always fetch the logs even if the log visibility is AllowedViewers and the controllers are not on the list?

spec/index.md
```diff
@@ -2604,11 +2604,11 @@ The total size of all returned logs does not exceed 4KiB.
...
 Logs persist across canister upgrades and they are deleted if the canister is reinstalled or uninstalled.
 
-The log visibility is defined in the `log_visibility` field of `canister_settings` and it can be:
+The log visibility is defined in the `log_visibility` field of `canister_settings`:
 
-- `controllers`: only visible to the canister's controllers (by default)
-- `allowlist`: visible to the provided list of principals, the maximum length of the list is 10
-- `public`: visible to everyone
+- `controllers`: only canister's controllers can fetch logs (by default)
+- `allowed_viewers`: only provided list of principals can fetch logs, the maximum length of the list is 10
```
mraszyk:
```suggestion
- `allowed_viewers`: only the provided list of principals can fetch logs, the maximum length of the list is 10
```

spec/index.md
maksymar: shall we allow developers to forbid fetching the logs? if yes, then allowed_viewers can be empty and should not include controllers.

spec/index.md
maksymar: Done.

spec/index.md
maksymar: done.

spec/index.md
mraszyk: > shall we allow developers to forbid fetching the logs?

that sounds like a good motivation to apply my suggestion then; hopefully there aren't many cases when a canister has many controllers and only needs many additional log viewers (as there can be at most 10 allowed viewers, but in such a case we can bump the limit of 10 I suppose)

spec/index.md
```diff
@@ -5878,7 +5884,12 @@ Q.canister_id = ic_principal
 Q.method_name = 'fetch_canister_logs'
 Q.arg = candid(A)
 A.canister_id = effective_canister_id
-S[A.canister_id].canister_log_visibility = Public or Q.sender in S[A.canister_id].controllers
+S[A.canister_id].canister_log_visibility = 
+  Public
```
mraszyk, line 5888:
```suggestion
(S[A.canister_id].canister_log_visibility = Public)
```
it's a top-level disjuction that we're splitting across lines

maksymar: done.

spec/index.md
mraszyk: > shall we allow developers to forbid fetching the logs?

We discussed in the spec meeting and in fact, one can't really prevent canister controllers from fetching the logs as they can just override the settings, fetch the logs, and set the allowed viewers back to the original list of principals. Moreover, when the set of controllers changes, the change needs to be synced in both the list of controllers and the list of allowed log viewers. So maybe it's worthwhile reconsidering the semantics of the allowed viewers' list w.r.t. canister controllers.

mraszyk:
One more point from the Interface spec meeting: should the canister itself be allowed to fetch the logs implicitly, just like it can fetch the canister status even if the canister does not control itself?

dsarlis:
> One more point from the Interface spec meeting: should the canister itself be allowed to fetch the logs implicitly, just like it can fetch the canister status even if the canister does not control itself?

As long as we don't allow fetch_canister_logs through composite queries then the canister cannot really fetch the logs. Or you mean that the canister would make an HTTPS outcall to fetch the logs? I doubt that would be particularly useful to anyone.

spec/index.md
dsarlis: @mraszyk Are you then suggesting that the semantics should be something like:

1. private -> only controllers can view
2. allowed_viewers -> controllers + principals in this list can view
3. public -> any principal is allowed to view

mraszyk:
> Or you mean that the canister would make an HTTPS outcall to fetch the logs?

That won't work implicitly anyway as the HTTPS outcall would have a self-authenticating caller that would need to be whitelisted explicitly.

spec/index.md
```diff
@@ -5878,7 +5884,11 @@ Q.canister_id = ic_principal
 Q.method_name = 'fetch_canister_logs'
 Q.arg = candid(A)
 A.canister_id = effective_canister_id
-S[A.canister_id].canister_log_visibility = Public or Q.sender in S[A.canister_id].controllers
+(S[A.canister_id].canister_log_visibility = Public)
+  or 
+  (S[A.canister_id].canister_log_visibility = Controllers and Q.sender in S[A.canister_id].controllers)
+  or 
+  (S[A.canister_id].canister_log_visibility = AllowedViewers Principals and Q.sender in Principals)
```
mraszyk, line 5891:
```suggestion
  (S[A.canister_id].canister_log_visibility = AllowedViewers Principals and (Q.sender in S[A.canister_id].controllers or Q.sender in Principals))
```
to always allow controllers to fetch the logs.

spec/index.md
```diff
@@ -2603,7 +2603,12 @@ The canister logs are *not* collected in canister methods running in non-replica
 The total size of all returned logs does not exceed 4KiB.
 If new logs are added resulting in exceeding the maximum total log size of 4KiB, the oldest logs will be removed.
 Logs persist across canister upgrades and they are deleted if the canister is reinstalled or uninstalled.
-The log visibility is defined in the `log_visibility` field of `canister_settings`: logs can be either public (visible to everyone) or only visible to the canister's controllers (by default).
+
+The log visibility is defined in the `log_visibility` field of `canister_settings` and can be one of the following variants:
+
+- `controllers`: only canister's controllers can fetch logs (by default)
+- `public`: everyone can fetch logs
+- `allowed_viewers` (`vec principal`): only the provided list of principals can fetch logs, the maximum length of the list is 10
```
Dfinity-Bjoern, line 2611:
```suggestion
- `allowed_viewers` (`vec principal`): only the provided list of principals and the canister's controllers can fetch logs, the maximum length of the list is 10
```

mraszyk:
@dfx-json The PR looks good to me, but the feature is still [disabled](https://github.com/dfinity/ic/blob/666ddbe8ad21d525f7bafcc01216c7b300824cbd/rs/config/src/execution_environment.rs#L363) in production so I haven't approved yet.

Dfinity-Bjoern (approved):

PR #327, opened 2024-07-29
fix: simplify definition of Delegation type

mraszyk:
This PR simplifies the definition of the `Delegation` type which is a simple record type and thus there's no need to formally define it as an enumeration type with a single variant `Delegation`.


oggy-dfin (approved):
LGTM

PR #328, opened 2024-08-06
chore: Move fetch_canister_logs to main management canister subsection

Dfinity-Bjoern:
Moving the method description to the suitable subsection (it's not part of the Bitcoin API).

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -2488,30 +2488,6 @@ A single metric entry is a record with the following fields:
 
 - `num_block_failures_total` (`nat64`): the number of failed block proposals by this node.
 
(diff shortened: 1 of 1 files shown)


mraszyk:
How about moving this further up before the provisional API and adding a section for the provisional API?

Dfinity-Bjoern:
Yes makes sense. As just discussed in the interface spec:
- General API
- Bitcoin
- Provisional
seems the right thing to do.

mraszyk:
> Yes makes sense. As just discussed in the interface spec:
> 
> * General API
> * Bitcoin
> * Provisional
> 
> seems the right thing to do.

Done.

mraszyk (approved):

PR #330, opened 2024-08-07
Fix the anchor of canister_info

lwshang:
Change from `ic-canister-info` to `ic-canister_info` to align with other management canister API anchors.


Dfinity-Bjoern (approved):
Thanks for the quick fix!

PR #333, opened 2024-09-01
fix: chunk store of a deleted canister is dropped

mraszyk:
This PR fixes the specification to drop the chunk store of a deleted canister. The implementation removes the chunk store by removing the entire canister from the replicated state including the chunk store stored in the canister's system state.


Dfinity-Bjoern (approved):
Thanks for fixing!

PR #337, opened 2024-09-03
fix: remove bitcoin query API from candid file

mraszyk:


Dfinity-Bjoern (approved):
Thanks for cleaning up!

PR #338, opened 2024-09-09
fix: canister memory usage - simplify and model chunk store

mraszyk:
This PR fixes the canister memory usage by modelling chunk store memory usage and simplifying notation.


Dfinity-Bjoern (approved):
Great, thank you Martin for cleaning that up!

PR #343, opened 2024-09-28
feat: allow anonymous query and read_state requests with invalid ingress_expiry

mraszyk:
This PR allows anonymous `query` and `read_state` requests with invalid `ingress_expiry`. The motivation is to allow users retrieving the certified subnet time via an anonymous read state request without having to provide a valid `ingress_expiry` (which they might not be able to do if they don't know the subnet time).

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -959,7 +959,7 @@ All requests coming in via the HTTPS interface need to be either *anonymous* or 
 
 -   `nonce` (`blob`, optional): Arbitrary user-provided data of length at most 32 bytes, typically randomly generated. This can be used to create distinct requests with otherwise identical fields.
 
--   `ingress_expiry` (`nat`, required): An upper limit on the validity of the request, expressed in nanoseconds since 1970-01-01 (like [ic0.time()](#system-api-time)). This avoids replay attacks: The IC will not accept requests, or transition requests from status `received` to status `processing`, if their expiry date is in the past. The IC may refuse to accept requests with an ingress expiry date too far in the future. These rules for ingress expiry apply not only to update calls but all requests alike (and could have been called `request_expiry`), except for anonymous `query` and `read_state` requests for which the IC may accept any provided expiry timestamp.
+-   `ingress_expiry` (`nat`, required): An upper limit on the validity of the request, expressed in nanoseconds since 1970-01-01 (like [ic0.time()](#system-api-time)). This avoids replay attacks: The IC will not accept requests, or transition requests from status `received` to status `processing`, if their expiry date is in the past. The IC may refuse to accept requests with an ingress expiry date too far in the future. These rules for ingress expiry apply not only to update calls but all requests alike (and could have been called `request_expiry`), except for anonymous `query` and anonymous `read_state` requests for which the IC may accept any provided expiry timestamp.
 
 -   `sender` (`Principal`, required): The user who issued the request.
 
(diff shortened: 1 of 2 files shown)


spec/index.md
```diff
@@ -959,7 +959,7 @@ All requests coming in via the HTTPS interface need to be either *anonymous* or
...
 
--   `ingress_expiry` (`nat`, required): An upper limit on the validity of the request, expressed in nanoseconds since 1970-01-01 (like [ic0.time()](#system-api-time)). This avoids replay attacks: The IC will not accept requests, or transition requests from status `received` to status `processing`, if their expiry date is in the past. The IC may refuse to accept requests with an ingress expiry date too far in the future. This applies not only to update calls, but all requests alike (and could have been called `request_expiry`).
+-   `ingress_expiry` (`nat`, required): An upper limit on the validity of the request, expressed in nanoseconds since 1970-01-01 (like [ic0.time()](#system-api-time)). This avoids replay attacks: The IC will not accept requests, or transition requests from status `received` to status `processing`, if their expiry date is in the past. The IC may refuse to accept requests with an ingress expiry date too far in the future. This applies not only to update calls, but all requests alike except for anonymous `query` and `read_state` requests (and could have been called `request_expiry`).
```
THLO, line 962:
```suggestion
-   `ingress_expiry` (`nat`, required): An upper limit on the validity of the request, expressed in nanoseconds since 1970-01-01 (like [ic0.time()](#system-api-time)). This avoids replay attacks: The IC will not accept requests, or transition requests from status `received` to status `processing`, if their expiry date is in the past. The IC may refuse to accept requests with an ingress expiry date too far in the future. These rules for ingress expiry apply not only to update calls but all requests alike (and could have been called `request_expiry`), except for anonymous `query` and `read_state` requests for which the IC may accept any provided expiry timestamp.
```

Dfinity-Bjoern (approved):
Looks good, as agreed.

spec/index.md
```diff
@@ -959,7 +959,7 @@ All requests coming in via the HTTPS interface need to be either *anonymous* or
...
+-   `ingress_expiry` (`nat`, required): An upper limit on the validity of the request, expressed in nanoseconds since 1970-01-01 (like [ic0.time()](#system-api-time)). This avoids replay attacks: The IC will not accept requests, or transition requests from status `received` to status `processing`, if their expiry date is in the past. The IC may refuse to accept requests with an ingress expiry date too far in the future. These rules for ingress expiry apply not only to update calls but all requests alike (and could have been called `request_expiry`), except for anonymous `query` and `read_state` requests for which the IC may accept any provided expiry timestamp.
```
fspreiss, line 962: My impression is that this formulation _could_ be interpreted such that _all_ read_state requests are exempt, and not just anonymous ones.

So, to avoid any confusion, we could be more precise/explicit by saying "except for anonymous `query` and anonymous `read_state` requests [...]".

spec/index.md
```diff
@@ -6282,7 +6282,7 @@ E.content = CanisterQuery Q
 Q.canister_id ∈ verify_envelope(E, Q.sender, S.system_time)
 |Q.nonce| <= 32
 is_effective_canister_id(E.content, ECID)
-S.system_time <= Q.ingress_expiry
+S.system_time <= Q.ingress_expiry or Q.sender = anonymous_id
```
rumenov, line 6285: Obviously this is a complication if we have yet another condition. 
I seriously don't know why we need to change the spec instead of relaxing the checks inside the replica without changing the spec.

PR #344, opened 2024-10-02
feat(crypto): Add support for Schnorr auxiliary inputs

randombit:
The PR proposes extending the `sign_with_schnorr` function to support "key-tweaking" as specified in Taproot ([BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)). An optional auxiliary argument is added to the function inputs, allowing the caller to specify the root of the Merkle tree that encodes alternative spending paths, which is then used by the replicas to tweak the signing key. This enhancement enables canisters to create Bitcoin Taproot addresses that fully comply with the BIP341 specifications.

# Background
The [Deuterium Milestone](https://internetcomputer.org/roadmap#Chain%20Fusion-Deuterium) introduced support for threshold Schnorr signatures, including Schnorr-BIP340 signatures used in Bitcoin. This allowed canisters to generate Pay-to-Taproot (P2TR) addresses and execute Taproot transactions. P2TR addresses, as defined in BIP341, support two possible spending paths:
* Key path: Spending by signing a transaction with a Schnorr key.
* Script path:Spending by revealing an alternative script stored in the leaf of a Merkle tree.

A P2TR address consists of a public key that is derived through an additive key derivation scheme. This process involves hashing the internal public key and the Merkle tree root (representing alternative spending scripts) to derive an "additive tweak." The tweak is then added to both the secret and public keys. To spend using the key path, a transaction must include a signature generated with the tweaked secret key, which corresponds to the tweaked public key embedded in the address.

# Motivation
Currently, on the Internet Computer, two types of restricted P2TR addresses can be created:
* **P2TR with an untweaked public key:** This type of address does not include a Merkle tree of alternative scripts, meaning it can only be spent via the key path. The creation of such addresses is supported since, for key-path spends, the transaction only requires a valid signature, and the Merkle root is intentionally omitted to conceal whether alternative spending scripts exist. 
* **P2TR address with a public key in a Merkle tree leaf:** This type of address can only be spent via the script path because canisters currently cannot request the Internet Computer (IC) to sign transactions with a tweaked public key.

This limitation restricts canisters from creating P2TR addresses that support spending using both the key path and script paths. As a result, canisters must decide at the time of address creation how they intend to spend the funds in the future. While this does not seem to constitute an issue to integrate with most Bitcoin meta-protocols, it does reduce the overall flexibility for dapps that require dynamic spending options. Additionally, most libraries do not natively support creating Taproot addresses that can be spent using untweaked keys via the key path. This forces developers to have a deeper understanding of Taproot to integrate the Schnorr signing API in their dapps.

By adding support for key-tweaking in the Schnorr signing API, canisters will be able to create standard Taproot addresses that support both script path and key path spending with a tweaked internal key. This enhancement would make the IC more flexible and aligned with standard Taproot functionality, reducing complexity for developers and dapps.


spec/index.md
```diff
@@ -2447,6 +2447,8 @@ The encoding of the signature depends on the key ID's `algorithm`:
 
 This call requires that a Schnorr key with ID `key_id` was generated by the IC and the signing functionality for that key was enabled. Otherwise, the call is is rejected.
 
+This call accepts an optional auxiliary parameter `aux`. The auxiliary parameter type `schnorr_aux` is a variant. The only currently supported variant is `bip341` which allows passing a Merkle tree root hash, which is required to implement Taproot signatures as defined in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki). The `bip341` variant is anly allowed for `bip340secp256k1` signatures, and if provided the `merkle_root_hash` blob must be exactly 32 bytes long, and should be generated in accordance with BIP341's specification for `taproot_output_script`. If no auxiliary parameter is provided, then `bip340secp256k1` signatures are generated in accordance with BIP340.
```
altkdf, line 2450:
```suggestion
This call accepts an optional auxiliary parameter `aux`. The auxiliary parameter type `schnorr_aux` is a variant. The only currently supported variant is `bip341` which allows passing a Merkle tree root hash, which is required to implement Taproot signatures as defined in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki). The `bip341` variant is only allowed for `bip340secp256k1` signatures, and if provided the `merkle_root_hash` blob must be exactly 32 bytes long, and should be generated in accordance with BIP341's specification for `taproot_output_script`. If no auxiliary parameter is provided, then `bip340secp256k1` signatures are generated in accordance with BIP340.
```

altkdf: It also may be better to reference `taproot_sign_key` instead of `taproot_output_script` because we are tweaking the secret key, although the procedure is essentially the same. It may also be reasonable to say that `merkle_root_hash` corresponds to `h` in `taproot_sign_key` of BIP341.

altkdf:
LGTM! Left some small improvement suggestions.

spec/index.md
andreacerulli: I think in that sentence `taproot_output_script` is the correct reference, as it is where the Merkle root is computed. However, I agree with the general intention of the comment to be a bit more explicit about the tweaking. In particular I would suggest that we clarify which key is used to sign if we pass or not the auxiliary output. 

> It may also be reasonable to say that merkle_root_hash corresponds to h in taproot_sign_key of BIP341.

+ 1

spec/index.md
```diff
@@ -2445,7 +2445,9 @@ The encoding of the signature depends on the key ID's `algorithm`:
...
 
-This call requires that a Schnorr key with ID `key_id` was generated by the IC and the signing functionality for that key was enabled. Otherwise, the call is is rejected.
+This call requires that a Schnorr key with ID `key_id` was generated by the IC and the signing functionality for that key was enabled. Otherwise, the call is rejected.
+
+This call accepts an optional auxiliary parameter `aux`. The auxiliary parameter type `schnorr_aux` is a variant. The only currently supported variant is `bip341` which allows passing a Merkle tree root hash, which is required to implement Taproot signatures as defined in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki). The `bip341` variant is only allowed for `bip340secp256k1` signatures, and if provided the `merkle_root_hash` blob must be exactly 32 bytes long, and should be generated in accordance with BIP341's specification for `taproot_output_script`. If no auxiliary parameter is provided, then `bip340secp256k1` signatures are generated in accordance with BIP340.
```
randombit, line 2450: Need to describe handling for empty root hash

Dfinity-Bjoern:
Looks good to me, we will discuss in the interface spec meeting before marking it final.

spec/index.md
```diff
@@ -2445,7 +2445,13 @@ The encoding of the signature depends on the key ID's `algorithm`:
...
 
-This call requires that a Schnorr key with ID `key_id` was generated by the IC and the signing functionality for that key was enabled. Otherwise, the call is is rejected.
+This call requires that a Schnorr key with ID `key_id` was generated by the IC and the signing functionality for that key was enabled. Otherwise, the call is rejected.
+
+This call accepts an optional auxiliary parameter `aux`. The auxiliary parameter type `schnorr_aux` is a variant. The only currently supported variant is `bip341` which allows passing a Merkle tree root hash, which is required to implement Taproot signatures as defined in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki). The `bip341` variant is only allowed for `bip340secp256k1` signatures, and if provided the `merkle_root_hash` must be generated in accordance with BIP341's specification for `taproot_output_script`. Specifically it should be either an empty bytestring (for the `script == None` case) or else 32 bytes generated using the procedure documented as `taproot_tree_helper`. If no auxiliary parameter is provided, then `bip340secp256k1` signatures are generated in accordance with BIP340.
```
mraszyk, line 2450:
```suggestion
This call accepts an optional auxiliary parameter `aux`. The auxiliary parameter type `schnorr_aux` is an enumeration. The only currently supported variant is `bip341` which allows passing a Merkle tree root hash, which is required to implement Taproot signatures as defined in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki). The `bip341` variant is only allowed for `bip340secp256k1` signatures, and if provided the `merkle_root_hash` must be generated in accordance with BIP341's specification for `taproot_output_script`. Specifically it should be either an empty bytestring (for the `script == None` case) or else 32 bytes generated using the procedure documented as `taproot_tree_helper`. If no auxiliary parameter is provided, then `bip340secp256k1` signatures are generated in accordance with BIP340.
```

spec/index.md
```diff
@@ -2445,7 +2445,13 @@ The encoding of the signature depends on the key ID's `algorithm`:
...
+
+On the Internet Computer, the tuple of the requested master key, the calling canister, and derivation path determines which private key that is used to generate the signature, and which public key is returned by `schnorr_public_key`.
```
mraszyk, line 2452:
```suggestion
On the Internet Computer, the tuple of the requested master key, the calling canister, and derivation path determines which private key is used to generate the signature, and which public key is returned by `schnorr_public_key`.
```

spec/_attachments/ic.did
```diff
@@ -322,10 +322,17 @@ type schnorr_public_key_result = record {
...
+    bip341: record {
+      merkle_root_hash: blob;
+   }
+};
+
 type sign_with_schnorr_args = record {
     message : blob;
     derivation_path : vec blob;
     key_id : record { algorithm : schnorr_algorithm; name : text };
+    aux: opt schnorr_aux;
```
mraszyk, line 335: Is the name `aux` clear/descriptive enough?

randombit: I'm open to alternatives

Dfinity-Bjoern: From interface spec meeting: If we don't see any such field being used for EdDSA, and the typing anyway does not guarantee consistency, we could also just have the `merkle_root_hash: opt blob` here.

randombit: Re `aux` the consideration here is in the future we wanted to support additional fields such as the context field of `Ed25519ctx` (see https://www.rfc-editor.org/rfc/rfc8032) this approach would make it easier to support since we could add a new member to the variant.

Dfinity-Bjoern: Maybe write it out as `auxiliary_parameters` so that it's a bit clearer (for people that do not spend their free time reading CS theory papers)?

mraszyk:
Superseded by https://github.com/dfinity/portal/pull/3758

PR #348, opened 2024-10-17
fix: ic0.call_perform spec

mraszyk:
This PR fixes issues in `ic0.call_perform` spec:
- adds one more condition under which a call successfully finalized by `ic0.call_perform` might still fail;
- specifies that discarding a pending call does not refund the prepayment for execution since this prepayment is not charged for a pending call.


Dfinity-Bjoern (approved):
Thanks for fixing!

PR #351, opened 2024-10-21
[FINAL] feat: Add sections for new subnet_info endpoint

michael-weigelt:
In certain circumstances, canisters wish to learn about subnet-specific data or metadata. This endpoint enables these use cases, starting with exposing the replica version a subnet is currently running. 

Open questions: 
- [x] Should this method be restricted to canisters or can any principal call it? **Update:** Yes, to pass on the cost to the caller
- [x] Naming: Is 'stats' fine if we anticipate that data like number of canisters, memory occupied etc might be added to the response struct? **Update:** Renamed to subnet_metrics **Update** Renaming to subnet_info!

### DIFF
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -342,6 +342,14 @@ type node_metrics_history_result = vec record {
     node_metrics : vec node_metrics;
 };
 
+type subnet_stats_args = record {
+    subnet_id : principal;
+};
+
+type subnet_stats_result = record {
+    replica_version : text;
+};
+
 type provisional_create_canister_with_cycles_args = record {
     amount : opt nat;
     settings : opt canister_settings;
@@ -443,6 +451,9 @@ service ic : {
 
     // metrics interface
     node_metrics_history : (node_metrics_history_args) -> (node_metrics_history_result);
+
+    // subnet stats
+    subnet_stats : (subnet_stats_args) -> (subnet_stats_result);
 
     // provisional interfaces for the pre-ledger world
     provisional_create_canister_with_cycles : (provisional_create_canister_with_cycles_args) -> (provisional_create_canister_with_cycles_result);
--- a/spec/_attachments/ic.did
+++ b/spec/_attachments/ic.did
@@ -342,11 +342,11 @@ type node_metrics_history_result = vec record {
     node_metrics : vec node_metrics;
 };
 
-type subnet_stats_args = record {
+type subnet_metrics_args = record {
     subnet_id : principal;
 };
 
-type subnet_stats_result = record {
+type subnet_metrics_result = record {
     replica_version : text;
 };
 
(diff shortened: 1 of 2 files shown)


mraszyk (commented):
LGTM

spec/_attachments/ic.did
```diff
@@ -342,6 +342,14 @@ type node_metrics_history_result = vec record {
     node_metrics : vec node_metrics;
 };
 
+type subnet_stats_args = record {
```
Dfinity-Bjoern, line 345: For consistency with the node metrics and the subnet metrics in the state tree – which may eventually converge to contain the same information – should we call this `subnet_metrics_...`?

michael-weigelt: Sounds good, I'll change it. Unless you are waiting for an opinion from somebody else?

adambratschikaye:
If a canister wants to get the information for it's own subnet, how is it supposed to figure out which subnet it is on?

mraszyk:
> If a canister wants to get the information for it's own subnet, how is it supposed to figure out which subnet it is on?

The registry has an endpoint that serves this information.

michael-weigelt:
Perhaps we should make the argument optional? If it's None, the canister's own subnet is used.

dsarlis:
Is there a reason why we went with `subnet_metrics` instead of `subnet_metadata` or `subnet_info`? The `replica_version` is not exactly a metric. Do we envision that we would eventually have things like number of canisters or state of the subnet served through this endpoint? Even then, I think `subnet_info` would still be ok (`subnet_metrics` would still suffer from including the `replica_version` which is not really a metric).

michael-weigelt:
>Do we envision that we would eventually have things like number of canisters or state of the subnet served through this endpoint

Yes, I believe. But I see the point. It's a misnomer now, and will be less of a misnomer later, but still not quite right. 

>why

Just in case you missed [the collapsed item](https://github.com/dfinity/interface-spec/pull/351#discussion_r1810717117).

mraszyk:
> Is there a reason why we went with `subnet_metrics` instead of `subnet_metadata` or `subnet_info`? The `replica_version` is not exactly a metric. Do we envision that we would eventually have things like number of canisters or state of the subnet served through this endpoint? Even then, I think `subnet_info` would still be ok (`subnet_metrics` would still suffer from including the `replica_version` which is not really a metric).

I agree that `subnet_info` makes more sense and is consistent with the existing `canister_info` endpoint. We already have a path `/subnet/<subnet_id>/metrics` in the state tree though and if we eventually converge into having the same data in the state tree and in `subnet_info`, then the name would be inconsistent (unless we could rename the state tree path).

dsarlis:
> > Is there a reason why we went with `subnet_metrics` instead of `subnet_metadata` or `subnet_info`? The `replica_version` is not exactly a metric. Do we envision that we would eventually have things like number of canisters or state of the subnet served through this endpoint? Even then, I think `subnet_info` would still be ok (`subnet_metrics` would still suffer from including the `replica_version` which is not really a metric).
> 
> I agree that `subnet_info` makes more sense and is consistent with the existing `canister_info` endpoint. We already have a path `/subnet/<subnet_id>/metrics` in the state tree though and if we eventually converge into having the same data in the state tree and in `subnet_info`, then the name would be inconsistent (unless we could rename the state tree path).

The state tree path contains only metrics for now so that is consistent imo. If we want to unify at some point (the two return a different set of information for now), then we can switch to `subnet_info` for the state tree path too. (Likely by introducing a new path that supersedes the old for backward compatibility).

Dfinity-Bjoern:
I am also fine with `subnet_info`, if that's the preferred term overall.

mraszyk:
Superseded by https://github.com/dfinity/portal/pull/3755

PR #352, opened 2024-10-24
fix: specify effective canister id for install_chunked_code

mraszyk:
This PR specifies the effective canister ID for the `install_chunked_code` method of the management canister whose candid argument does not have a `canister_id` field.


mraszyk:
Thanks, @chenyan-dfinity , for reporting!

Dfinity-Bjoern (approved):
Thanks for the fix, @chenyan-dfinity and @mraszyk

PR #354, opened 2024-10-30
Deprecating the Bitcoin API of the management canister

THLO:
In this PR, all Bitcoin API endpoints on the management canister are declared DEPRECATED.
Developers are advised to call the corresponding endpoints on the Bitcoin canister directly.


Dfinity-Bjoern:
Shouldn't we create an API spec for the Bitcoin canister?

PR #357, opened 2024-11-11
Canister limit

luc-blaeser:
Add function to query current resource limit for the canister.

### DIFF
--- a/spec/index.md
+++ b/spec/index.md
@@ -1469,6 +1469,7 @@ The following sections describe various System API functions, also referred to a
 
     ic0.time : () -> (timestamp : i64);                                         // *
     ic0.global_timer_set : (timestamp : i64) -> i64;                            // I G U Ry Rt C T
+    ic0.canister_limit : (limit_type : i32) -> (limit : i64);                   // * s
     ic0.performance_counter : (counter_type : i32) -> (counter : i64);          // * s
     ic0.is_controller: (src: i32, size: i32) -> ( result: i32);                 // * s
     ic0.in_replicated_execution: () -> (result: i32);                           // * s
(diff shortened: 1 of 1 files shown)


Dfinity-Bjoern:
This seems a great start, Luc! I assume we want something similar for cycles cost, although, as @mraszyk pointed out in the interface spec meeting, it should probably be a separate call since it's going to return `u128` not `i64`.

Are we planning to move implementation forward on the replica side soon? Is that driven by the execution team @dsarlis?

dsarlis:
> Are we planning to move implementation forward on the replica side soon? Is that driven by the execution team @dsarlis?

This would be a discussion point. I cannot promise soon but we can discuss potential timelines.

dsarlis:
> This seems a great start, Luc! I assume we want something similar for cycles cost, although, as @mraszyk pointed out in the interface spec meeting, it should probably be a separate call since it's going to return `u128` not `i64`.

About this point, yeah, the cycles cost would technically need to return `u128` but it might end up similar. One idea we had internally was to expose lower level system api which returns various fees and then CDKs can build higher level abstractions where they combine them in a sensible way. The alternative being to return directly relevant costs, like how much to make this https outcall or ecdsa signature.

The decision would be part of the ticket that we have filed on our side and @michael-weigelt will most likely take on at some point. I personally have a slight preference for the higher level approach of having apis return directly the costs (we can then do some more clever stuff on the replica that is harder to perform on the user level, whether it's CDK or canister code) and it seems more user friendly. But we would have to see what makes the most sense at an API level (the downside of direct costs, is that they depend on size of requests in many cases, making it harder to see a unified API for all costs -- as opposed to a simpler "give me this fee and that fee").

mraszyk:
Superseded by https://github.com/dfinity/portal/pull/3754

